PR #1, opened 2015-11-03
Comment away...

jonasfj:
@gregarndt, @petemoore, pete asked me to file this so people can add comments...

Please feel free to submit PRs for `master`...
Or just flat out commit code, if you feel like pseudo-coding some missing aspect :)

We probably need to come up with the `runTask` driver... and the host abstractions... However, I see host abstraction as pretty far from the `engines` and `plugins` so less important...

Also configuration loader + logic on conditionally loaded plugins, and some solution for each plugin and engine specifying JSON schema for its input, and the worker still being able to do both schema validation, schema export for docs, and JSON parsing... So plugins and engines don't have to deal with those boring things, but just declare what input they want.

-----
To the extend it's feasible without writing way too much docs... I would like to have 2-3 lines of example (often in the context of docker) explaining why we chose to do this abstraction in a given manner.

**Note,** it's not really my intend to drive this process too much... this was just what came out of me playing with golang and having dreamed about better abstractions in docker-worker. I think getting these interfaces down is key to a stable and robust worker. Without some decent abstraction the code easily becomes inefficient and buggy spaghetti.. As keeping track of all the clean ups and corner cases is very hard.


engine/engine.go
```diff
@@ -0,0 +1,132 @@
...
+type FeatureSet struct {
+	IsSingletonEngine bool
+}
+
+// NewEngine returns an engine implementation, if the requested engine is
+// available under the current build contraints, GOOS, GOARCH.
+//
+// This function is intended to be called once immediately after startup, with
+// engine of choice as given by configuration.
+func NewEngine(engineName string, runtime *runtime.EngineContext) (Engine, error) {
```
petemoore: Should we call this simply `New` instead of `NewEngine`? See _Simplify function names_ in the [go blog](https://blog.golang.org/package-names).

engine/errors.go
```diff
@@ -0,0 +1,51 @@
...
+	ErrFeatureNotSupported   = errors.New("Feature not support by current engine")
+	ErrResourceNotFound      = errors.New("The referenced resource wasn't found")
+	ErrSandboxTerminated     = errors.new("The Sandbox has terminated")
+	ErrSandboxAborted        = errors.new("Exection of sandbox was aborted")
+	ErrNonFatalInternalError = errors.new("Engine encountered a non-fatal internal error")
+	ErrContractViolation     = errors.new("Engine has detected a contract violation")
+	ErrEngineIsSingleton     = errors.New("Engine cannot run multiple sandboxes in parallel")
+	ErrEngineNotSupported    = errors.New("Engine is not available in the current configuration")
+	ErrEngineUnknown         = errors.New("Engine with the given doesn't exist")
+)
```
petemoore: Errors will need instance data with them for them to be helpful. For example, an error that says `Feature not supported by current engine` is much harder to troubleshoot than an error `Feature "livelog" not supported by engine "OSX-jail-worker-v1"`. Therefore I think we can create different go types for each class of error, but the errors themselves should keep references to the instance data, so when the `.String()` method is called to generate the error output, we see all the relevant information. This way, the _type_ would map to one of these classes of error, but the _instance_ would contain references to the actual error that occurred.

Let's therefore define types here, and not instantiate any variables.

utils/atomics.go
```diff
@@ -0,0 +1,46 @@
+package utils
```
petemoore: Let's not have a utils package. See "Bad package names" in the [go blog](https://blog.golang.org/package-names).

engine/engine.go
jonasfj: `r+` please land that change :)

engine/errors.go
jonasfj: I think this makes sense for some errors, like the `MalformedPayloadError` which is defined below...

But if the current engine returns `ErrFeatureNotSupported` from `AttachProxy(...)` then it's clearly evident from the context that `AttachProxy` feature isn't supported on the current engine.

The plugin that gets this error, must then either: (A) decide that the task can proceed without the proxy (perhaps injecting the data some other way), or (B) return an ErrMalformedPayload message saying that `engine.Name()` doesn't support `AttachProxy`, hence, `<my-feature>` isn't supported here... And you must formulate your task payload in a manner than doesn't active this feature.

Adding `engine.Name()` might be nice... But I also fear someone could use it to write a plugin that is engine specific.

utils/atomics.go
jonasfj: how about `misc` then? :) he he...

I agree needs refactored at some point! For now I've just been dumping things here...

engine/errors.go
petemoore: These are valid points. One problem I've seen is this pattern:

```go
func XYZ(...) (<result type>, error) {
    ...
    ..., err := <some function call>
    if err != nil {
        return nil, err
    }
    ...
    return result, nil
}
```

Here we are essentially throwing an error up the stack, much like in java you'd just add a throws clause to a method, so that a particular error type just gets passed back up (note this is one example where java would do this more succinctly).

```java
public <return type> XYZ(...) throws <error type> {
    ... = <some function call>;
    return result;
}
```

In the java version, this is absolutely ok, because when you get the error back, you have a stack trace to where the error occurred within `<some function call>` so debugging is simpler. However, in the go version, you don't have a stack trace, just an error message, and you don't know where within the `XYZ` function or any function it called, however deeply nested, the error occurred. This is really nasty, and often makes debugging really difficult, without the stack trace. For this reason I tend to lean on putting more information into error messages, in case they are handled in this way. Of course, if you are in control of the code that calls the methods, you can avoid this pattern above. If others might be calling your functions, it is nice to add this information.

Note using `panic` and `recover` might help, but normally it is recommended in go that `panic` and `recover` can be used internally, but exposed APIs should not panic, rather return errors.

Maybe something like [go-errors](https://bugsnag.com/blog/go-errors) might help us here.

jonasfj: So the case here is a `plugin` that calls a method on `engine` and gets `ErrFeatureNotSupported`.

If the plugin forwards this error to it's parent, task execution will abort and the worker will have an internal error. As `plugin` isn't allowed to return the  `ErrFeatureNotSupported`.

For each method in an interface I specify which non-fatal errors the implementor may return. If the implementor returns an error that isn't declared as non-fatal we won't attempt to handle it in any graceful manner. Most likely we'll just report the task as exception, writing some logs, and possibly crash the machine to prevent trashing of tasks (maybe there is a threshold of tasks you may thrash before crashing).

But essentially, if an error isn't listed as non-fatal the error won't be handled gracefully. So a plugin that just forwards `ErrFeatureNotSupported` has an internal error. It may convert `ErrFeatureNotSupported` into a `MalformedPayloadError` saying that the proxy or cache attach option is invalid...

jonasfj:
@pmoore, closing this.. feel free to open a new PR if you have additional comments...

PR #3, opened 2016-01-22
First attempt at doing something with schema

jonasfj:
Some clean up and moving forward...

Also initial work with go-extpoints..

Notably:
 * I've decided to move some contexts from `runtime` the contexts that we only use as wrappers for a set of arguments (for forward compatibility) will now live in the package where they are used. Typically they are only used once. Also I'm renaming these contexts into options, like SandboxOptions and EngineOptions, they are collections of arguments, rather than contexts.
 * first iteration of a composeable schema thingy... @petemoore  with some luck, I think you go:generate trick to go from JSON schema document to go-type can be used to everything needed to call `NewCompositeSchema`.... Notice the result of NewCompositeSchema.Parse is interface{} but it'll always be the type create by `makeTarget`... So when SandboxOptions has `Payload interface{}` you can always cast that to the result type of what your `makeTarget` returns.
 * CompositeSchema can be used for more than one property, but then you have to deal with positional things... I think that's okay, most plugins will only used one property. But if they really want to they can do fancier things.
 * You need `go install github.com/progrium/go-extpoints` and `go install github.com/jonasfj/go-import-subtree` to do `go generate`...


jonasfj:
@petemoore you're welcome to change things I've done here... just jump in and go crazy..
Just make sure you base on this... This PR is more to open for comments that blocking merging...

jonasfj:
@petemoore, @gregarndt you're still welcome to comment...

PR #5, opened 2016-01-23
Just playing around

gregarndt:
Wanted to start experimenting with some stuff so I took Jonas' rename branch and started messing around.  Probably screwed around with things too much, but hey, it builds! :)


jonasfj:
I like the rename to `engines/`..

engines/extpoints/interfaces.go
```diff
@@ -0,0 +1,10 @@
+// Package extpoints provides methods that engine plugins can register their
+// implements with as an import side-effect.
+package extpoints
+
+import (
+	"github.com/taskcluster/taskcluster-worker/engines"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+)
+
+type EngineProvider func(environment *runtime.Environment, options *engines.EngineOptions) (engines.Engine, error)
```
jonasfj: Does this work? Ie. I think the go-extpoints generator thing strictly looks for interfaces, but I could be wrong?

Ahh, seems I was wrong... This should work:
https://github.com/progrium/go-extpoints/blob/master/main.go#L35

jonasfj: ie. it works for Function types and interface types...

jonasfj: Nice!!! :)

jonasfj: I like simplification...

engines/engine.go
```diff
@@ -1,6 +1,20 @@
...
+
+// An EngineOptions is a wrapper for the set of options/arguments given when
+// an Engine is created.
+//
+// We pass all options as a single argument, so that we can add additional
+// properties without breaking source compatibility.
+type EngineOptions struct {
+	GarbageCollector *gc.GarbageCollector
+	//TODO: Add some sort of interface to the system logger
+	//TODO: Add some interface to submit statistics for influxdb/signalfx
```
jonasfj: I think `runtime.Environment` can be a property on EngineOptions.. .And since this is strictly for wrapping arguments to `EngineProvider` I suggest we move it into the extpoints package which is the only place it is used...

engines/mock/doc.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/taskcluster/taskcluster-worker/engines/extpoints"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+)
+
+func init() {
+	// Register the mock engine as an import side-effect
+	extpoints.EngineProviders.Register(NewEngine, "mock")
+}
+
+func NewEngine(environment *runtime.Environment, options *engines.EngineOptions) (engines.Engine, error) {
```
jonasfj: No need to have this in upper case... why would we want to expose this method from the package?

engines/mock/doc.go
```diff
@@ -0,0 +1,20 @@
...
+	"errors"
+
+	"github.com/taskcluster/taskcluster-worker/engines"
+	"github.com/taskcluster/taskcluster-worker/engines/extpoints"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+)
+
+func init() {
+	// Register the mock engine as an import side-effect
+	extpoints.EngineProviders.Register(NewEngine, "mock")
```
jonasfj: Actually, if we're sneaky I think we could do:
```go
func init() {
    extpoints.EngineProviders.Register(func(options *engines.EngineOptions) {
      return nil, errors.New("Engine not implemented!")
    }, "mock")
}
```

After all most engine constructors are just going to set `engines.EngineOptions` as a property on the object... and then maybe call some functions to validate that the environment it setup correclty.. ie. docker socket exists, etc...

main.go
```diff
@@ -1 +1,56 @@
...
+	}
+
+	e := args["--engine"]
+	if e == nil {
+		panic("Must supply engine type")
+	}
+
+	engine := e.(string)
+
+	if validateEngine(engine, extpoints.EngineProviders.All()) == false {
```
jonasfj: There is also `extpoints.EngineProviders.Lookup`

jonasfj: Just saying...

engines/mock/doc.go
gregarndt: ah yea, I was messing around with something and uppercased it without realizing that wasn't the solution and never reverted it.

main.go
gregarndt: ah yea, I wrote that method before actually using the lookup method on line 48.  I could just get rid of that completely now that I know Lookup() exists.  Part of this is me learning what go-extpoints can do.

plugins/extpoints/interfaces.go
```diff
@@ -0,0 +1,41 @@
...
+	"sync"
+
+	"github.com/taskcluster/taskcluster-worker/engines"
+	p "github.com/taskcluster/taskcluster-worker/plugins"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+)
+
+// The Plugin Factory is responsible for maintaining a list of plugin factories
+// that are used to create new instances of a Plugin for a task execution.
+type PluginFactory func(options *p.PluginOptions) p.Plugin
```
jonasfj: Nope... This is bad... It's also why I renamed `PluginFactory` to `PluginEnvironment` (which still isn't a good name)...

If this is a cache folder plugin, then `PluginFactory` will hold a list of all idle cache folders... If there is just one `PluginFactory` for the cache folder plugin in the entire process then it uses global variables... I'll never need two, but it's nice to have the option for testing...

Ie. `PluginFactory` is stateful therefore we should have the option of having more than one, for the same kind of plugin... Only useful in testing, but still useful.

jonasfj:
I'll take a look playing with this some more... Most of it looks really good.

gregarndt:
thanks, most of it was to mess around with go-extpoints and getting the thing to build and run (which it does now)

plugins/plugin.go
```diff
@@ -1,119 +1,143 @@
...
 //
-// If a required feature is unsupport the methods may return a
-// MalformedPayloadError. All other errors are fatal.
+// The methods are called in the order listed here with the exception of
+// Exception() which may be called following any method, and Dispose() which
+// will always be called as a final step allowing you to clean up.
 type Plugin interface {
-	// Prepare will be called in parallel with NewSandboxBuilder(), this is a good
-	// place to start downloading and extracting resources required.
+	PayloadSchema() runtime.CompositeSchema
```
jonasfj: `PayloadSchema` returns the `CompositeSchema` that is used to construct the `Payload` property on `PluginOptions`... Hence, it needs to exist at a higher level...

PR #6, opened 2016-01-24
Rename 3

jonasfj:
Please take a look... I addressed the comments I had to your playing around... (I think I got them all)..

There is some todos in there... 

Notably; `PluginFactory` is now `Plugin` and `Plugin` is now `TaskPlugin`...


petemoore:
Merging to avoid lag at this early development stage. Let's address issues in other PRs or discuss online. We can always revert commits if we don't like them...

PR #7, opened 2016-01-25
More enginetests

jonasfj:
More tests... 

We need some for proxies too... and some for stuff being written to log...

This is still too naive in that the runtime environemnt has nil as garbagecollector...
Also the taskcontext is nil, we should fake those so we can test more :)

And it doesn't always try to dispose volumes...


jonasfj:
but at least it should be pretty simple to get 4 solid test cases for any engine we come up with... :)

PR #10, opened 2016-01-27
Added proxy tests

jonasfj:


jonasfj:
Note ideally we should deduplicate this code but we can do that any day...

PR #11, opened 2016-01-27
Updating composite schema tests

gregarndt:


.gitignore
```diff
@@ -1 +1,2 @@
 /taskcluster-worker
+coverage.out
```
gregarndt: Added so that a common filename could be given to the output of the coverage report and not be included in checkin.

jonasfj:
Nice, test cases... I haven't really played with reflect before but that is definitely awesome for testing!!!

PR #12, opened 2016-01-28
Update schema tests

gregarndt:


runtime/compositeschema_test.go
```diff
@@ -91,5 +88,112 @@ func TestInvalidSchemaReference(t *testing.T) {
...
+
+	composites := make([]CompositeSchema, 4)
+	for i := range composites {
+		// Create unique property names to avoid collision resulting in error
+		propertyName := fmt.Sprintf("prop%d", i)
+		composite, err := NewCompositeSchema(propertyName, schema, true, func() interface{} { return nil })
+		if err != nil {
+			t.Fatal(err)
+		}
+		composites[i] = composite
```
jonasfj: This is fancy...

Note: typically I would do `composites := []CompositeSchema{}` and then do `composites = append(composites, composite)` as it's safer... Well, not really when you do range over composites... nevermind this is smart :)

runtime/compositeschema_test.go
```diff
@@ -91,5 +88,112 @@ func TestInvalidSchemaReference(t *testing.T) {
...
+	}
+
+	merged, err := MergeCompositeSchemas(composed1, composed2)
+	if err != nil {
+		t.Fatalf("Error creating a merged composite scheme. %v", err)
+	}
 
+	m := merged.(*composedSchema)
+	if len(m.entries) != 2 {
+		t.Fatalf("Merged schema should have 4 entries but had %d", len(m.entries))
```
jonasfj: nope... does this test pass? You are merging in a tree... So you should only have two... as both of those each has two...

jonasfj: oh, wait it's only the comment that is out of date :)

gregarndt: ah yea, sorry, the comment was out of date after I realize it's just two entries

PR #15, opened 2016-01-29
Enginetest

jonasfj:
A few tests...
But mostly I expand abstractions with what is needed to support vnc... and write tests for this...

I totally think I can write tests for VNC, and make the mock engine fake enough of the VNC protocol to make these test pass :)

Note: We certainly don't have to add vnc initially, it's just super powerful if the infrastructure is ready and it's just a matter of patching up the engines to support it...


Makefile
```diff
@@ -20,9 +20,9 @@ generate:
...
 check: test
-test:
-	go test -race ./...
-	go vet ./...
 	# tests should fail if go generate or go fmt results in uncommitted code
 	git status --porcelain
 	/bin/bash -c 'test $$(git status --porcelain | wc -l) == 0'
+test:
+	go test -race ./...
+	go vet ./...
```
jonasfj: Moving things around here... is mainly because I like to run `make test` locally while I have changes in local folder... so the output from `git status --porcelain` is annoying (and makes test fail, because I haven't committed yet).

PR #16, opened 2016-02-02
Bug 1245063 - Add logrus logging package to be used by runtime/plugins/engines

gregarndt:
The scope of this has changed a bit but I preserved commits just in case we want to take something that ended up getting deleted.

This implementation makes use of logrus, which is used by popular projects such as Docker.  It has an easy to use interface with multiple formatters depending on if you're running this in a terminal, writing somewhere else, etc.


gregarndt:
Fixing up some stuff, adding a mutex, docs, etc

log/log.go
```diff
@@ -0,0 +1,114 @@
...
+	mu  sync.Mutex
+	out io.Writer
+	// List of default fields that will be serialized
+	// along with the message and written to `out`
+	defaultFields map[string]interface{}
+	level         loggingLevel
+}
+
+// Log a debug message as long as the logger's debug level is set to at least Debug
+func (l *Logger) Debug(message string, fields map[string]interface{}) {
```
jonasfj: I'm curious have you considered doing something like `logrus`.... ie. declare a `Tags` type like they have `Fields`... and accept `a ...interface{}` as message (var args are nice)...
Using something like `WithFields` is also nice, as we might not always have fields...

Honestly, I'm not a fan of doing our own logger, just using an existing library to have loggers available in various context where they are useful.

-----
Note: I'm not opposed to using `logrus`, I haven't really looked at it... but it seems to support child loggers...

Ie. it would be possible to do something like:
```go
log := logrus.New();

log.WithFields(Fields{"tag": "value"}).Debug("Hello", "world", 5)
log.Debug("Hello", "world", 5)

// Create task log for TaskContext (this is for system messages, not the livelog that is uploaded to queue)
taskLog := log.WithFields(Fields{"taskId": "...", "runId": 1})
taskLog.WithFields(Fields{"pulling": "img", "time": 100}).Debug("Pulling an image: ", img)
taskLog.Debug("Pulled an image: ", img)
```

If we think that `logrus` exposes too many methods.. And we want to do some hiding... 
We can just define an interface, containing only the methods we want plugins and engines to use.
For example I see no need to expose `log.Reader()`...

jonasfj: Proposed interface to hide logrus behind:

https://public.etherpad-mozilla.org/p/logrus-simplified

gregarndt:
Had a discussion with Jonas on IRC and he finally came around to my idea of using logrus it seems :) It looks like we reallyw ould be reinventing the wheel with some niceties we want and what logrus already offers.  It's used by the docker projects and I suspect some other large ones out ther as well so it won't be going away anytime soon.   Also as shown in this PR, it's trivial to add the basic functionality we're looking for.

gregarndt:
Logging has been moved to logrus and a logger interface creates so engines/plugins can just reference the interface even if the underlying logging package changes (as long as the new logging package maintains that interface).

engines/extpoints/interfaces.go
```diff
@@ -32,4 +32,4 @@ type EngineOptions struct {
 // Any error here will be fatal and likely cause the worker to stop working.
 // If an implementor can determine that the platform isn't supported at
 // compile-time it is recommended to not register the implementation.
-type EngineProvider func(options EngineOptions) (engines.Engine, error)
+type EngineProvider func(options *EngineOptions) (engines.Engine, error)
```
jonasfj: I propose we make this a non-pointer type. See Note in the definition of `EngineOptions`.

Basically, I think when we have `<...>Options` where the struct is only used as input parameters for a method, then it makes sense to pass it by value.
It when we do struct grouping just because grouping the input parameters into a struct allows us to add new parameters without breaking old code.

I see this `<...>Options` pattern as passing key-word arguments in python. Ie. you'll always call `EngineProvider` as:
```go
engine, err := engineProvider(EngineOptions{
  Environment: ...
  ....
})
```

So it's just a pattern used to pass arguments, not an actual struct that we recommend people pass around as a value. Not a struct that will have have methods or anything like that.

----
I think doing so keeps the pattern and usage of these `<...>Options` structs very simple.

jonasfj: @petemoore, I haven't coded much go, maybe I'm wrong to suggest this pattern... It just seems like a  good way to solve the problem of adding additional input paramters... Without introducing something that is inherently complicated.

Note: This should all be documented in comments inside `EngineOptions` and similar options structs.

engines/mock/mockengine.go
```diff
@@ -7,19 +7,22 @@ import (
...
 }
 
 func init() {
 	// Register the mock engine as an import side-effect
 	extpoints.EngineProviders.Register(func(
-		options extpoints.EngineOptions,
+		options *extpoints.EngineOptions,
 	) (engines.Engine, error) {
-		return engine{}, nil
+		logger := options.Environment.Log.WithField("engine", "mock")
```
jonasfj: I for the record: `Log` wasn't added to `EngineOptions` :)


Also I think `.WithField("engine", engineName)` is something we should do when we construct the `EngineOptions`. Because there is one place in the main function where we initialize the engine, that is it... So we won't have to implement this in every engine and we are sure we always write the tag the same way.

Basically, create child loggers, when passing a logger into a new component. Rather than doing it inside the constructor of the component. So if mock engine wanted to pass a logger to some sub-component it uses, it could add tags there.

jonasfj: > I for the record: Log wasn't added to EngineOptions :)

Forget that comment I was wrong... it's part of the runtime environment... Hmm... it probably should definitely be there...
But how about we add a child logger to `EngineOptions` when we create it... It'll be a child logger of the logger on the runtime environment.

jonasfj: hmm... maybe it doesn't have to be exposed from runtime environment... because we always want to force engine and plugins to use a childlogger that was passed in using a `<...>Options` structure...

log/log.go
```diff
@@ -0,0 +1,63 @@
+package log
```
jonasfj: Package needs a description...

I suspect we could put this into `runtime/log/` because it's sort of part of the runtime that is necessary to run these things...

log/log.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/Sirupsen/logrus"
+)
+
+// Logger interface allows a logging object to be passed around as long as they
+// implement these methods without requiring a specific package.  Allows migrating
+// between logging packages along as this contract is maintained.
+type Logger interface {
+	WithField(key string, value interface{}) *logrus.Entry
+	WithFields(fields logrus.Fields) *logrus.Entry
+	WithError(err error) *logrus.Entry
```
jonasfj: Return value should be `Logger` as it returns an instance of the Logger interface.

engines/mock/mockengine.go
jonasfj: Note the child loggers should probably be the `log.Logger` interface...

log/log.go
jonasfj: since this is just the definition of what a logger is... and a utility function to create one... We could have it in `runtime/log.go` it's not super important...

engines/extpoints/interfaces.go
gregarndt: Ok, I can change it back to not passing around a pointer.  Changing between a pointer and a value didn't change the behavior we're trying to accomplish though....passing around a struct that can have fields that change without needing to change the method signature of the receivers of it.

log/log.go
gregarndt: Yea, I think moving it to runtime is a good idea.  There's just nothing really here to make it it's own package.

jonasfj: note: for slightly more complex things subpackages of runtime also makes sense...

I'm just trying to avoid blowing up the root folder with small things... Honestly, I'm tempted to move atomics too... either into runtime, or making some utilities package... for all the small stuff..

runtime/log.go
```diff
@@ -0,0 +1,26 @@
+package runtime
+
+import (
+	"fmt"
+
+	"github.com/Sirupsen/logrus"
+)
+
+// Create a logger that can be passed around through the environment.
```
jonasfj: Start docs comments like this with `CreateLogger ....` I think go fmt or one of its friends keeps telling me so...

plugins/extpoints/interfaces.go
```diff
@@ -11,13 +12,14 @@ import (
...
+// adding any large fields please consider adding them as pointers.
+// Note: This is intended to be a simple argument wrapper, do not add methods
+// to this struct.
 type PluginOptions struct {
 	environment *runtime.Environment
 	engine      *engines.Engine
-	// Note: This is passed by-value for efficiency (and to prohibit nil), if
-	// adding any large fields please consider adding them as pointers.
-	// Note: This is intended to be a simple argument wrapper, do not add methods
-	// to this struct.
```
jonasfj: nit, but this wasn't really intended as a doc comment... more as a source comment for people to read when they were editing the struct...

main.go
```diff
@@ -6,45 +6,68 @@ package main
...
-	if e == nil {
-		panic("Must supply engine type")
+	var level string
+	if l := args["--logging-level"]; l != nil {
+		level = l.(string)
+	}
+	logger, err := runtime.CreateLogger(level)
+	if err != nil {
+		os.Stderr.WriteString(err.Error())
+		os.Exit(1)
```
jonasfj: nit, but `panic(err)` is just more elegant...

runtime/log.go
```diff
@@ -0,0 +1,26 @@
...
+// WithField or WithFields and specifying additional fields that the package
+// would like.
+func CreateLogger(level string) (*logrus.Logger, error) {
+	if level == "" {
+		level = "warn"
+	}
+
+	lvl, err := logrus.ParseLevel(level)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to parse logging level: %s\n", level)
```
jonasfj: Nice, didn't know about `fmt.Errorf` I like that...

Note: not sure if it's always wise to wrap errors... I guess it can't really hurt here...

plugins/extpoints/interfaces.go
```diff
@@ -11,13 +12,14 @@ import (
...
+	log         *logrus.Entry
```
jonasfj: So if we did the interface thing... We could to just hide the `WithFields` method as it requires `logrus.Fields`... Looking at this code I think `WithField(string, ínterface{})` is sufficient... In rare cases where we have two fields we can just do `log.WithField(...).WithField(...)`..

jonasfj:
I like this `r+` for sure (we should have merged this last week)... Comments are minor can be ignored... We should just move forward. Feel free to fix any minor comments you agree with...

runtime/log.go
petemoore: `go vet` does that (which is currently in the `make test` target)

main.go
petemoore: I think `panic` is appropriate for crashes, but not appropriate for handling bad input - i.e. a user should never see a stack trace because he/she passed in an unrecognised logging level. In any case, I think we should process the underlying cause and crash if really an internal error, or report the bad value, if it is just a bad value. We could also consider creating and documenting different exit codes for different failures, although I believe exit code 1 is appropriate for invalid arguments.

runtime/log.go
petemoore: For all errors that we wrap, we should include the underlying error in the message, so we don't lose the root cause - i.e. include an extra `%s` for `err`.

main.go
gregarndt: We do not need a stack trace, nor do we need to run deferred functions, so panic does not seem like the best suited option here.  This is an error we can't recover from, and should exit immediately.  It also allows us to define exit codes for things we know went wrong.

Panic seems more appropriate when it happens lower down in the execution of the worker (like if we had a plugin hit some unrecoverable state and we need to unwind the stack all the way up) and complete any deferred functions for cleanup.

For my understanding of what is idiomatic in Go, and in the case of using panic here, what is elegant about using panic vs this method of exiting?

runtime/log.go
gregarndt: is there something I should enable for it to warn me about things like this?  'make test' nor running 'go vet' directly has indicating any problems.

runtime/log.go
gregarndt: Currrently the only error that method returns is "not a valid logrus Level: <level>".  My attempt here was to remove the use of the 'logrus' word in the error message as the user doesn't care about what that package is or what it means.

runtime/log.go
gregarndt: Good call on changing this documentation by the way, still getting used to how things should be documented.

petemoore: My bad, I think it is https://github.com/golang/lint

main.go
jonasfj: Agree... sorry I'm wrong... panic is bad here ! :)

runtime/log.go
jonasfj: > For all errors that we wrap, we should include the underlying error in the message,

Unless we're explicitly handling the error... because we know what it means

but yeah, generally I agree...

main.go
gregarndt: Thanks for raising the question about it though, it really gets me thinking about why something is better than the alternative rather than just copying from other places :)

PR #17, opened 2016-02-05
Integrate query service

gregarndt:


config/config.go
```diff
@@ -0,0 +1,16 @@
+package config
```
gregarndt: This is just to support other things, just boilerplace that will get removed with the config loading

main.go
```diff
@@ -57,13 +59,38 @@ func main() {
...
 		Environment: &runtimeEnvironment,
-		Log:         logger.WithField("engine", engine),
+		Log:         logger.WithField("engine", engineName),
 	})
 	if err != nil {
 		logger.Fatal(err.Error())
 	}
 
+	// TODO (garndt): Need to load up a real config in the future
+	config := &config.Config{
```
gregarndt: more boiler plate that will be removed

config/config.go
```diff
@@ -0,0 +1,16 @@
...
+		Certificate string
+	}
+	ProvisionerId string
+	WorkerGroup   string
+	WorkerId      string
+	Capacity      int
+	QueueService  struct {
+		ExpirationOffset int
+	}
+}
```
jonasfj: I thought we were considering to use `CompositeSchema` for config loading..... So plugins and engines could specify config to be loaded... Obviously, we wouldn't want to validate the config against their subschemas until we've loaded the config and decided which engines and plugins to load... 

Hmm... or maybe that is just a sub key of the config specified here... actually that could work too :)

taskmgr/task_manager.go
```diff
@@ -0,0 +1,60 @@
...
+	Tasks []*runtime.TaskContextController
+	// Maxmimum capacity that the worker is configured for.
+	MaxCapacity   int
+	Engine        *engines.Engine
+	Log           *logrus.Entry
+	Queue         *queueService
+	ProvisionerId string
+	WorkerGroup   string
+	WorkerId      string
+}
```
jonasfj: `workerType` and `PluginManager` (or at least a todo to add the plugin manager).

jonasfj:
So generally I think `config` and `taskmgr` should merged to a single `worker` package.

All `main.go` does is to load the config, ie. fill out the structure... and call `worker.New(config).Start()`

taskmgr/queueservice.go
```diff
@@ -0,0 +1,387 @@
...
+		WorkerId         string
+		WorkerGroup      string
+		Log              *logrus.Entry
+	}
+)
+
+// Given a number of tasks needed, the Azure task queues will be polled in order
+// of priority until either there are no more tasks to claim, or the given number of
+// tasks has been fulfilled.
+func (q queueService) ClaimWork(ntasks int) []*TaskRun {
```
jonasfj: Not saying we have to refactor this right now... But I think the smart interface for this is:
```
taskpoller.New(config, func() time.Duration { return pollDelay }, func(run *TaskRun) { /* start new task */ });
taskpoller.Start() // start polling when tasks are needed
taskpoller.RequestTasks(n int) // Request n tasks to be claimed and /* start new task */ be called in new go routine
// After finishing a task we shall call taskpoller.RequestTasks(1) 
// Initially we call taskpoller.RequestTasks(capacity)
taskpoller.Stop() // Stop polling
```

This way we can start a task as soon as we get it... And we can increasing the maximum number of tasks we're willing to claim between two requests, say after trying the highest queue and finding that it had nothing...

That said, I think this interface isn't too bad either :)

The first func argument to New is all about letting something else decide how long to sleep between polling everything again... I guess in the current interface that isn't controlled here... Which might be okay too.. I need to read me.. :)

taskmgr/queueservice.go
```diff
@@ -0,0 +1,387 @@
...
+
+type (
+	// Used for modelling the xml we get back from Azure
+	QueueMessagesList struct {
+		XMLName       xml.Name       `xml:"QueueMessagesList"`
+		QueueMessages []QueueMessage `xml:"QueueMessage"`
+	}
+
+	// Used for modelling the xml we get back from Azure
+	QueueMessage struct {
```
jonasfj: this should not be exported

taskmgr/queueservice.go
```diff
@@ -0,0 +1,387 @@
...
+		// and run what tasks we do have.
+		if len(tasks) >= ntasks {
+			return &tasks, nil
+		}
+		taskRuns, err := q.pollTaskUrl(&queue, ntasks-len(tasks), q.Log)
+		if err != nil {
+			q.Log.Warnf("Could not retrieve tasks from the Azure queue. %s", err)
+			continue
+		}
+		tasks = append(tasks, taskRuns...)
```
jonasfj: This is not good... It'll work but it's sub-optimal... And increases the risk of timeouts... 
That's why I suggest the alternative interface...

Btw, I propose that the information we get from azure queue is called a `TaskTip` because it's not necessarily a task.. it's not a claim yet. But a _tip_ that there may be a task to claim, so we should attempt to claim it before deleting the tip. But if we get a 409 when trying to claim it we should just delete the obviously invalid tip. 

I see a system as follows: (just pseudo code)
```
//polling thread:
for !self.stopped {
  for self.tasksRequested > 0 && !self.stopped {
    ensureQueueUrls() // Calls queue.PollTaskUrls() if necessary
    for q := range queueUrls {
       // Keep polling from highest level queue until it doesn't return any task
       for !self.stopped {
         taskTips := pollAzureQueue(q, tasksRequested) // returns list of {taskId, runId, signedDeleteUrl}
         if len(taskTips) == 0 {
           break;
         }
         for task := range taskTips {
           go self.claimAndRun();
         }
         self.tasksRequested -= len(taskTips);
       }
       if self.tasksRequested == 0 || self.stopped {
         break;
       }
    }
  }
  // the sleepBetweenPolls is a parameter in the constructor, so the sleep interval can adjust depending on
  // billing cycle (or cloud that we're running on).
  time.Sleep(self.sleepBetweenPolls());
}
```

In the `claimAndRun` something like:
```
// Claim from queue:
resp, err := queue.claimTask(taskId, runId, ...)
if err != nil {
  if err.statusCode == 409 {
    // Call signedDeleteUrl for the taskTip.
  } else {
    log.Error(err);
  }
  self.tasksRequested += 1;
  return;

  // Call signedDeleteUrl for the taskTip as the claimTask was successful
 
  taskClaim := {... resp... taskTip, or combination there of}

  // Again the function runTask is given by the constructor to QueuePoller
  // The worker here supplies a function that takes care of processing the task.
  // and when the worker is done processing the task.. .The worker calls
  // QueuePoller.incTaskRequest(1); Effectively QueuePoller have an internal
  // counter of how many tasks the worker is currently willing to accept.
  // The worker gets these task via. a callback.
  // We could also use a channel, but a callback just seems easier.
  go self.runTask(taskClaim);
}
```

Note: all this psuedo code doesn't account for locking, some places self is missing too... etc...
For some of these things like the `tasksRequested` it might make sense to use a atomic integer, we can easily make one just like I made the AtomicBool (both should probably live in runtime/atomics or something like that).

For `stopped` we should use an atomic bool too...


----
What I'm trying to accomplish here is two things:
 1) When a taskTip is received from azure queue storage, we start a new go routine and start with claimTask, deleteSignedUrl, as well as sending it off to the worker for processing.
 2) If we have two azure queues p1 and p2 and have just polled p1 for 1 task, but didn't get any (we're about to poll p2). But the worker finishes a task and calls `QueuePoller.incTaskRequest(1)` to get a new task for the now empty capacity, p2 will be polled for 2 tasks. And if p1 had returned 1 tasks, but during this request the worker had increased the taskRequest count by one, p1 would be polled again. 
 3) Ensure that the sleeping interval between polls if the polling was empty, is controlled by the worker. (this is also the case in your design as the worker must be doing the sleeping).
 4) If we have two azure queues p1, p2, we request 5 tasks from p1 and get 1, we should poll p1 again, until we get zero tasks, before proceeding to p2. This is a bug in the current implementation. 

(1) is the only one here that is really important. It reduces unnecessary delay and risk of lengthy requests causing azure queue claims to timeout. Note, in the future there will likely be a variable number of priority queues, as datacenter workerTypes might actually build up quite a pending queue. So if there is bad internet weather so retries and polling takes time, the first azure message could risk timing out before we get to call deleteSignedUrl for it. Let's avoid this, and do maximum concurrency when we do I/O, especially when it's time sensitive (even if the sensitivity is on the order of 5 minutes).

jonasfj: Just outlining what I think is a better interface... Perhaps we can:
  A) implement this interface using the current queue service... it won't have the benefits of (1) - (4), but we'll be able to improve it later (without having to refactor the rest of the worker package)
  B) Not do anything and refactor this later...

jonasfj: Also this is about how far I got with the review... :)

taskmgr/task_status_updater.go
```diff
@@ -0,0 +1,174 @@
...
+				e <- err
+			}
+		} else {
+			// current status is such that we shouldn't update to new
+			// status, so just report that no error occurred...
+			e <- nil
+		}
+	}()
+	return e
+}
```
petemoore: I think we can make the dynamic internal functions (e.g. `reportCompleted := func(...) { ... }`) exported and static, and such that instead of passing something into the update channel, you just call the method, and internally it pushes a struct down a channel. Then you would just call taskMgr.ReportCompleted(...) etc.

Also, I think the updater instance should be specific to the `TaskRun` - i.e. we could have something like:

```
	TaskRun struct {
		TaskId              string                         `json:"taskId"`
		RunId               uint                           `json:"runId"`
		SignedDeleteUrl     string                         `json:"-"`
		TaskClaimResponse   tcqueue.TaskClaimResponse      `json:"-"`
		TaskReclaimResponse tcqueue.TaskReclaimResponse    `json:"-"`
		Definition          tcqueue.TaskDefinitionResponse `json:"-"`
		reclaimTimer        *time.Timer
		taskUpdater         *taskmgr.Manager
	}
```
Then the updating of two different task runs can be done in parallel without any need for blocking, as there would be one `Manager` per task run.

config/config.go
gregarndt: I added a comment at the base of this file (in the PR) indicating that this was just mocking out some config stuff until we had it officially in place.  Pete has been working on his jsonschemago library so we can specify the schemas (in yaml) for the configuration pieces that the worker, engines, and plugins care about and generate a complete schema along with the types for it.

PR #19, opened 2016-02-08
Run `go lint` inside `make test`

petemoore:
Note, it will return a zero exit code, even if there are lint suggestions, so won't break builds. This way it brings it to your attention, but doesn't force you to fix it.


Makefile
```diff
@@ -9,7 +9,7 @@ export PATH := $(GOPATH)/bin:$(PATH)
 all: build
 build:
 	go fmt ./...
-	go install ./...
+	go get ./...
```
gregarndt: Hrm, why the change from install to get?

petemoore: In case dependencies get added or changed. Also if someone downloads the git repo and builds, rather than using `go get`. Note, `go get -u ./...` would also update dependencies, but that comes at quite a cost in performance, so we might want to do that in a different target (`rebuild`? or maybe create a new `clean` target?) .....

Makefile
```diff
@@ -24,5 +24,8 @@ check: test
 	git status --porcelain
 	/bin/bash -c 'test $$(git status --porcelain | wc -l) == 0'
 test:
+	go get -t ./...
+	go get github.com/golang/lint/golint
 	go test -race ./...
 	go vet ./...
+	golint ./...
```
jonasfj: I would love to have a target that just runs:
```
go test -race ./...
go vet ./...
golint ./...
```
Actually, all I care about is `go test -race ./...` I need this target because it is what I run from terminal during debugging... I can't remember to spell `go test -race ./...` every time.. :)

We could totally call it `make runtest` or something... or `make go-test`, I moved the git checks into `make check` for this reason...

jonasfj:
This seems sane to me... land it :)

PR #21, opened 2016-02-10
Auto-generate source code from json schema defintions and go generate instructions

petemoore:
@jonasfj @gregarndt 

Work in progress to auto-generate code based on json schema for engine config, plugin config, and plugin payload schema config...

Raising PR before work is done so interested parties can see what I am playing with. I'll add a comment when it is ready for formal review and is complete - it is just my work in progress at the moment.


codegen/plugin-gen/main.go
```diff
@@ -0,0 +1,108 @@
...
+	version = "plugin-gen 1.0.0"
+	usage   = `
+plugin-gen
+plugin-gen is a tool for generating source code for a plugin, based on two json schema files
+(stored as either json or yaml) for its config settings and its portion of the task payload.
+As well as creating go types that the payload and config can be unmarshalled into, it also
+generates the PayloadSchema() and ConfigSchema() method implementations for the plugin.
+
+  Usage:
+    plugin-gen -c CONFIG-SCHEMA-FILE -p PAYLOAD-SCHEMA-FILE -o GO-OUTPUT-FILE
```
jonasfj: I would propose the following:
 1) this tool be called `go-composite-schema`
 2) Commandline be on the form: `go-composite-schema [--required] <property> <input.yml> <output.go>`
 3) `<input.yml>` must be a JSON schema
 4) `<output.go>` will be a file as follows:
```go
type <schema-title> struct {
  .... // Whatever is generated from the schema
};

var <schema-title>Schema = func() runtime.CompositeSchema {
  schema, err := NewCompositeSchema('<property>', `
    <contents of input.yml as json>
  `, <true if --required>, func() interface{} { return &<schema-title>{}; });
  if err != nil {
    panic(err)
  }
  return schema;
}()
```

Maybe instead of using `<schema-title>` we could define a property called `typeName` on the schema, which jsonschema2go could use if present. That way we can name the type something sane, without compromising our documentation.

@petemoore, I would even be open to allowing you to add `typeName` as a top-level property to all our schemas in all TC components... Granted that would obviously be a lot of work to do :)

With this tool, the PluginProvider implementation becomes very simple, you just return `<schema-title>Schema` from the `ConfigSchema()` method...
I don't advice that we auto generate that method, because it's very easy to write... And this way we have a generation tool that ensures all of our `CompositeSchema`s are sane... And we can use it where ever we have CompositeSchemas. Whather for engine/plugin payload, or engine/plugins config, or config for some completely different sub-system we make in the future.

-----

On-topic, I suggest that we focus our energy on something other than this... but if you want to finish it up, this is a direction that would keep it simple, relevant and ensure that it can be widely used throughout our codebase. By focusing on CompositeSchema rather than plugins and engines.

Note: main value here is obviously that we can write in YAML, we can validate the schema again the meta-schema upfront, and we probably should do that :)

petemoore:
@jonasfj I think this is in a state you can review it in now. Note, I added some dummy config files for illustrative purposes - we can remove them if you like.

codegen/global-config.yml
```diff
@@ -0,0 +1,16 @@
...
+  workerId:
+    title:          WorkerID
+    description: |-
+      blah blah
+    type:           string
+additionalProperties: false
+required:
+  - engines
+  - plugins
+  - workerId
```
petemoore: Note, I wanted to create a yml file for the global config (everything except plugins and engines) but haven't yet written anything to use it. Probably we could use `go-composite-schema` for this too.

plugins/livelog/config-schema.yml
```diff
@@ -0,0 +1,14 @@
...
+type:               object
+properties:
+  liveLogExe:
+    title:          LiveLogExecutable
+    description: |-
+      The executable (.exe file) to run the livelog service
+    type:           string
+additionalProperties: false
+required:
+  - liveLogExe
```
petemoore: @jonasfj Note - this file is for illustration - I *DON'T* propose we use an .exe file for the livelog service - just stole the config from generic worker for illustration.

engines/docker/generated_configschema.go
```diff
@@ -0,0 +1,47 @@
...
+type (
+	// Config applicable to docker engine
+	Config struct {
+
+		// Root Volume blah blah
+		RootVolume string `json:"rootVolume"`
+	}
+)
+
+func ConfigSchema() runtime.CompositeSchema {
```
jonasfj: I was tempted to do:
`var configSchema = func() runtime.CompositeSchema { .... return ... ; }()`

So it's just a variable...


Also quick question, is it possible to make these types private... I don't want to export them from the package.
Obviously `RootVolume` need to be exported for json to work... but `Config struct {` does not.
Similar for `ConfigSchema()` no need to export it... Also I suspect it's easier and safer to do it as a variable..
That way if there is a panic() inside the function it'll always happen when the program starts, so tests would never work, if the is a panic()...

engines/docker/generated_configschema.go
```diff
@@ -0,0 +1,47 @@
...
+		      "description": "Root Volume blah blah",
+		      "title": "Root Volume",
+		      "type": "string"
+		    }
+		  },
+		  "required": [
+		    "rootVolume"
+		  ],
+		  "title": "Config",
+		  "type": "object"
```
jonasfj: Not sure what go does here... But a stable JSON result wouldn't very useful... Maybe avoid whitespace..

CompositeSchema does a simple string comparison if two entries specifies the same schema... Maybe one day we can improve that... but stable json string wouldn't hurt either...

petemoore: No worries - I already have written something for that: https://github.com/taskcluster/taskcluster-base-go/blob/fa319fce2375b4a38488af1521317f4a1b15d3fa/jsontest/jsontest.go#L5-L18

engines/docker/generated_configschema.go
petemoore: Are you sure we don't want to export them - I think it wouldn't hurt if someone using a package could see the json schema, as this helps when we write automated tooling to e.g. publish schemas etc. Of course, if we really don't want it, we can - it feels though correct that we should be able to generically infer stuff about plugins/engines etc. That said, the generated code is not for the engine/plugin itself, but just package identifiers, so yeah, I guess we don't want to export that. Fair enough. Yes, I can add support for that.

petemoore: We can stick the function as a variable, and admittedly that means it would be easier to mock (although I'm not sure why we'd mock it - but at least we'd be able to). However, I don't think it means we'll have startup-panics, as the function would be defined, but not executed, wouldn't it? Or would something call that function at startup?

jonasfj: >  Or would something call that function at startup?

The pattern I wrote here:
`var configSchema = func() runtime.CompositeSchema { .... return ... ; }()`
Notice the subtle `}()` that the end... I'm actually setting `configSchema` to be the result of an anonymous function that is immediately invoked :)

So it'll be executed before `main()` is called.

jonasfj: > Are you sure we don't want to export them - I think it wouldn't hurt if someone using a package could see the json schema, as this helps when we write automated tooling to e.g. publish schemas etc. Of course, if we really don't want it, we can - it feels though correct that we should be able to generically infer stuff about plugins/engines etc. That said, the generated code is not for the engine/plugin itself, but just package identifiers, so yeah, I guess we don't want to export that. Fair enough. Yes, I can add support for that.

This reads like one of my comments where I start out disagreeing and end up agreeing as I write more arguments :) he he...
(honestly I do those kind of comments too often).

In short, yes, if someone is using the package, they should use it by importing it and then use `extpoints` to find the EngineProvider/PluginProvider of it.. which will return the schema in question...

engines/docker/generated_configschema.go
jonasfj: awesome... we should copy that into the compositeSchema stuff...

engines/engine.go
```diff
@@ -44,6 +40,10 @@ type Engine interface {
 	// from this method.
 	PayloadSchema() runtime.CompositeSchema
 
+	// ConfigSchema returns the CompositeSchema that represents the engine
+	// configuration
+	ConfigSchema() runtime.CompositeSchema
```
jonasfj: Wrong place, this should be added to the `EngineProvider` which should be changed from function to interface.

plugins/plugin.go
```diff
@@ -34,6 +34,10 @@ type Plugin interface {
 	// from this method.
 	PayloadSchema() (runtime.CompositeSchema, error)
 
+	// ConfigSchema returns the CompositeSchema that represents the plugin
+	// config.
+	ConfigSchema() runtime.CompositeSchema
```
jonasfj: Same as the other one...

jonasfj:
Generally I really like this...

engines/docker/generated_configschema.go
petemoore: > Notice the subtle }() that the end...

D'oh! Yeah, totally makes sense. This also has the nice side effect that `init()` functions inside plugin/engine packages that register plugin and engine providers can even safely refer to the generated variables (since variables are initialised before `init()` is called). :+1:

Makefile
```diff
@@ -14,6 +14,7 @@ build:
 generate:
 	go get github.com/progrium/go-extpoints
 	go get github.com/jonasfj/go-import-subtree
```
petemoore: By the way Jonas, can we also move this into taskcluster org?

engines/docker/generated_configschema.go
petemoore: Thinking about this some more, we don't really have a place to put general purpose go packages. Should we keep taskcluster-base-go for general purpose packages? For example I'm using jsontest in several projects (generic-worker, taskcluster-client-go, taskcluster-worker, ...). I could stick it in the client, but it is not client-specific.

We could rename taskcluster-base-go to just 'go' or 'go-libs' - then we'd have e.g.:

* `github.com/taskcluster/go/jsonutil`
* `github.com/taskcluster/go/text`
* `github.com/taskcluster/go/atomics`
* `github.com/taskcluster/go/atomics`

I think it would be overkill to place each package in its own repository, so a single repository for general utility packages seems like a reasonable compromise.

jonasfj: > We could rename taskcluster-base-go to just 'go'
I don't oppose that... But... it breaks the github fork/PR model..
If we really want this I propose calling it tc-go-libs...

------

But honestly, I don't see the need for us to have a collection of go libraries...
For small things I propose that we:
 1) put them in `taskcluster-client-go/tcutil/`
 2) put them in `taskcluster-worker/runtime/`
 3) Make then well documented and put them in `<yourname>/<packagename>` or `taskcluster/<packagename>` depending on how generic they are...

I think (1) and (2) will cover all the small stuff. (3) makes sense for things that are reusable outside the taskcluster project...

Makefile
jonasfj: sure... but we'll need to update references too... can we do it separately...

Note: added you can garndt as collaborators for now if you have any hacks for it...
I'm very much trying to keep `go-import-subtree` super generic and not taskcluster specific, which I why I wasn't exactly sure where to put it...

jonasfj:
just merge this and we can work some more on the small things... I would rather have it in than out...

even if we change the names a bit, it's easy to refactor...

petemoore:
Sorry I got caught up with other issues yesterday. I'm happy to make the fixes before merging - I agree with them all! :)

I actually created the review request for this PR in bugzilla only because I was cleaning up bug-state at the time, not because I was finished - I should have clarified in the bug. I'll ping you when it is done, and then hopefully we can review again then, and all being well, merge.

gregarndt:
This is something that has been on my mind for awhile regarding how we treat pull requests and patches.  I think for taskcluster-worker we should consider a PR as the initial starting point that will be mutable and not the final solution.  The goal is that once it's merged it's the final agreed upon solution but not before hand.

This way everyone is under the understanding that when a PR is open, it's not that it's a final review, but rather just a request for feedback in the early stages of the PR changes.

petemoore:
@jonasfj good enough to go now?

petemoore:
jonasfj
19:18 pmoore ship it

PR #22, opened 2016-02-11
Update Queue Service Tests

gregarndt:
These tests could be consolidated probably using fixtures and some generative testing, but they do work and caught some issues with the logic in the queue service.

### DIFF
--- a/taskmgr/mock_test.go
+++ b/taskmgr/mock_test.go
@@ -1,0 +1,42 @@
+package taskmgr
+
+import (
+	"github.com/stretchr/testify/mock"
+	"github.com/taskcluster/taskcluster-client-go/queue"
+	"github.com/taskcluster/taskcluster-client-go/tcclient"
+)
+
+type MockQueue struct {
+	mock.Mock
+}
+
+func (m *MockQueue) ReportCompleted(taskId, runId string) (*queue.TaskStatusResponse, *tcclient.CallSummary, error) {
+	args := m.Called(taskId, runId)
+	return args.Get(0).(*queue.TaskStatusResponse), args.Get(1).(*tcclient.CallSummary), args.Error(2)
+}
+
+func (m *MockQueue) ReclaimTask(taskId, runId string) (*queue.TaskReclaimResponse, *tcclient.CallSummary, error) {
+	args := m.Called(taskId, runId)
+	return args.Get(0).(*queue.TaskReclaimResponse), args.Get(1).(*tcclient.CallSummary), args.Error(2)
+}
+
+func (m *MockQueue) PollTaskUrls(provisionerId, workerType string) (*queue.PollTaskUrlsResponse, *tcclient.CallSummary, error) {
+	args := m.Called(provisionerId, workerType)
+	return args.Get(0).(*queue.PollTaskUrlsResponse), args.Get(1).(*tcclient.CallSummary), args.Error(2)
+}
+func (m *MockQueue) CancelTask(taskId string) (*queue.TaskStatusResponse, *tcclient.CallSummary, error) {
+	args := m.Called(taskId)
+	return args.Get(0).(*queue.TaskStatusResponse), args.Get(1).(*tcclient.CallSummary), args.Error(2)
+}
+func (m *MockQueue) ClaimTask(taskId, runId string, payload *queue.TaskClaimRequest) (*queue.TaskClaimResponse, *tcclient.CallSummary, error) {
+	args := m.Called(taskId, runId, payload)
+	return args.Get(0).(*queue.TaskClaimResponse), args.Get(1).(*tcclient.CallSummary), args.Error(2)
+}
+func (m *MockQueue) ReportFailed(taskId, runId string) (*queue.TaskStatusResponse, *tcclient.CallSummary, error) {
+	args := m.Called(taskId, runId)
+	return args.Get(0).(*queue.TaskStatusResponse), args.Get(1).(*tcclient.CallSummary), args.Error(2)
+}
+func (m *MockQueue) ReportException(taskId, runId string, payload *queue.TaskExceptionRequest) (*queue.TaskStatusResponse, *tcclient.CallSummary, error) {
+	args := m.Called(taskId, runId, payload)
+	return args.Get(0).(*queue.TaskStatusResponse), args.Get(1).(*tcclient.CallSummary), args.Error(2)
+}
--- a/taskmgr/queueservice.go
+++ b/taskmgr/queueservice.go
@@ -132,9 +132,13 @@ func (q queueService) claimTask(task *TaskRun) bool {
 
 	err := <-UpdateTaskStatus(update, q.client, q.Log)
 	if err != nil {
-		if err.StatusCode != 401 || err.StatusCode != 403 || err.StatusCode < 500 {
-			_ = q.deleteFromAzure(task.TaskId, task.SignedDeleteUrl)
+		if err.StatusCode == 401 || err.StatusCode == 403 || err.StatusCode >= 500 {
+			// Do not delete the message if task could not be claimed because of server
+			// or authorization failures
+			return false
 		}
+
+		_ = q.deleteFromAzure(task.TaskId, task.SignedDeleteUrl)
 		return false
 	}
 	_ = q.deleteFromAzure(task.TaskId, task.SignedDeleteUrl)
--- a/taskmgr/queueservice.go
+++ b/taskmgr/queueservice.go
@@ -63,12 +63,25 @@ type (
 
 	taskRuns []*TaskRun
 
+	// TaskclusterQueue is an interface to the Queue client provided by the
+	// taskcluster-client-go package.  Passing around an interface allows the
+	// queue client to be mocked
+	TaskclusterQueue interface {
+		ReportCompleted(string, string) (*tcqueue.TaskStatusResponse, *tcclient.CallSummary, error)
+		ReportException(string, string, *tcqueue.TaskExceptionRequest) (*tcqueue.TaskStatusResponse, *tcclient.CallSummary, error)
+		ReportFailed(string, string) (*tcqueue.TaskStatusResponse, *tcclient.CallSummary, error)
+		ClaimTask(string, string, *tcqueue.TaskClaimRequest) (*tcqueue.TaskClaimResponse, *tcclient.CallSummary, error)
+		ReclaimTask(string, string) (*tcqueue.TaskReclaimResponse, *tcclient.CallSummary, error)
+		PollTaskUrls(string, string) (*tcqueue.PollTaskUrlsResponse, *tcclient.CallSummary, error)
+		CancelTask(string) (*tcqueue.TaskStatusResponse, *tcclient.CallSummary, error)
+	}
+
 	queueService struct {
 		mu               sync.Mutex
 		queues           []taskQueue
 		Expires          tcclient.Time
 		ExpirationOffset int
-		client           *tcqueue.Queue
+		client           TaskclusterQueue
 		ProvisionerId    string
 		WorkerType       string
 		WorkerId         string
--- a/taskmgr/queueservice.go
+++ b/taskmgr/queueservice.go
@@ -44,7 +44,6 @@ type (
 		TaskClaimResponse   tcqueue.TaskClaimResponse      `json:"-"`
 		TaskReclaimResponse tcqueue.TaskReclaimResponse    `json:"-"`
 		Definition          tcqueue.TaskDefinitionResponse `json:"-"`
-		reclaimTimer        *time.Timer
 	}
 
 	taskQueue struct {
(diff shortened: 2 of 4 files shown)


taskmgr/mock_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/stretchr/testify/mock"
+	"github.com/taskcluster/taskcluster-client-go/queue"
+	"github.com/taskcluster/taskcluster-client-go/tcclient"
+)
+
+type MockQueue struct {
+	mock.Mock
+}
+
+func (m *MockQueue) ReportCompleted(taskId, runId string) (*queue.TaskStatusResponse, *tcclient.CallSummary, error) {
```
petemoore: Rather than passing `taskId` and `runId` strings, I wonder if it might be nicer for the task run to be encapsulated in a type, and just to pass that.

taskmgr/queueservice_test.go
```diff
@@ -15,49 +19,736 @@ import (
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, 1, len(tasks))
+	assert.True(t, deleteCalled, "Delete should have been called when dequeue count above threshold")
+
+}
+
+func TestClaimTask(t *testing.T) {
```
petemoore: Some of these functions are quite long, so it might be helpful to add a sentence or two to the function comments to give a high level about what the test is checking.

taskmgr/queueservice_test.go
```diff
@@ -15,49 +19,736 @@ import (
...
+	var handler = func(w http.ResponseWriter, r *http.Request) {
+		var message string
+		if r.URL.Path == "/tasks/1234" {
+			message = messages[1]
+		} else {
+			message = messages[0]
+		}
+		w.Header().Set("Content-Type", "application/xml")
+		w.Write([]byte(message))
+	}
```
petemoore: It looks like this handler appears in more than one place, I wonder if we could pull it out to avoid the code duplication.

taskmgr/queueservice_test.go
```diff
@@ -15,49 +19,736 @@ import (
...
+	messages := `<?xml version="1.0" encoding="utf-8"?>
+	<QueueMessagesList>
+	  <QueueMessage>
+		<MessageId>5974b586-0df3-4e2d-ad0c-18e3892bfca3</MessageId>
+		<InsertionTime>Fri, 09 Oct 2009 21:04:30 GMT</InsertionTime>
+		<ExpirationTime>Fri, 16 Oct 2009 21:04:30 GMT</ExpirationTime>
+		<PopReceipt>YzQ4Yzg1MDItYTc0Ny00OWNjLTkxYTUtZGM0MDFiZDAwYzEw</PopReceipt>
+		<TimeNextVisible>Fri, 09 Oct 2009 23:29:20 GMT</TimeNextVisible>
+		<DequeueCount>1</DequeueCount>
+		<MessageText>eyJhYmMiLDB9xxxxxxx</MessageText>
```
petemoore: Maybe to make it more obvious, we could put non-base64 characters in this string such as `*` or `$` etc. I think it only fails to be base64 since it is not a multiple of 4 bytes, but one extra `x` and it would be.

taskmgr/mock_test.go
```diff
@@ -0,0 +1,42 @@
+package taskmgr
+
+import (
+	"github.com/stretchr/testify/mock"
+	"github.com/taskcluster/taskcluster-client-go/queue"
+	"github.com/taskcluster/taskcluster-client-go/tcclient"
+)
+
+type MockQueue struct {
+	mock.Mock
```
petemoore: Great that you found the mock client - that is awesome! :+1:

taskmgr/queueservice.go
```diff
@@ -53,22 +44,34 @@ type (
...
+	// queue client to be mocked
+	TaskclusterQueue interface {
+		ReportCompleted(string, string) (*tcqueue.TaskStatusResponse, *tcclient.CallSummary, error)
+		ReportException(string, string, *tcqueue.TaskExceptionRequest) (*tcqueue.TaskStatusResponse, *tcclient.CallSummary, error)
+		ReportFailed(string, string) (*tcqueue.TaskStatusResponse, *tcclient.CallSummary, error)
+		ClaimTask(string, string, *tcqueue.TaskClaimRequest) (*tcqueue.TaskClaimResponse, *tcclient.CallSummary, error)
+		ReclaimTask(string, string) (*tcqueue.TaskReclaimResponse, *tcclient.CallSummary, error)
+		PollTaskUrls(string, string) (*tcqueue.PollTaskUrlsResponse, *tcclient.CallSummary, error)
+		CancelTask(string) (*tcqueue.TaskStatusResponse, *tcclient.CallSummary, error)
+	}
```
petemoore: Here it would be nice to declare names for the parameters (e.g. `CencelTask(taskId string)`) to make it clearer what each parameter is for.

taskmgr/queueservice.go
```diff
@@ -53,22 +44,34 @@ type (
...
 		SignedPollUrl   string `json:"signedPollUrl"`
 	}
 
-	TaskRuns []*TaskRun
+	taskRuns []*TaskRun
+
+	// TaskclusterQueue is an interface to the Queue client provided by the
+	// taskcluster-client-go package.  Passing around an interface allows the
+	// queue client to be mocked
+	TaskclusterQueue interface {
```
petemoore: Maybe we could call this simply `Queue` as it is in a different package to `github.com/taskcluster/taskcluster-client-id/queue` (i.e. from client side, it would be always be referred to as `taskmgr.Queue` rather than `queue.Queue` so the names would not collide). This might be nicer than `taskmgr.TaskclusterQueue`. Note, [this blog post](http://blog.golang.org/package-names) on package/type naming totally inspired me.

taskmgr/queueservice.go
petemoore: As above, it might be nicer in these interface methods to receive the `*TaskRun` object which was provided when the work was claimed. This should also help ensure (although not guarantee) that the `taskId` and `runId` are coherent.

taskmgr/queueservice.go
```diff
@@ -186,16 +189,16 @@ func (q queueService) deleteFromAzure(taskId string, deleteUrl string) error {
...
-	queueMessagesList := &QueueMessagesList{}
+	var r queueMessagesList
 	// To poll an Azure Queue the worker must do a `GET` request to the
 	// `signedPollUrl` from the object, representing the Azure queue. To
 	// receive multiple messages at once the parameter `&numofmessages=N`
 	// may be appended to `signedPollUrl`. The parameter `N` is the
 	// maximum number of messages desired, `N` can be up to 32.
 	// Since we can only process one task at a time, grab only one.
-	resp, _, err := httpbackoff.Get(fmt.Sprintf("%s%s%d", taskQueue.SignedPollUrl, "&numofmessages=", ntasks))
+	resp, _, err := httpbackoff.Get(fmt.Sprintf("%s%s%d", taskQueue.SignedPollUrl, "&numofmessages=", int(math.Min(32, float64(ntasks)))))
```
petemoore: The comment and the implementation seem to be inconsistent - i.e. the comment says up to a max of 32, but I think the implementation gives a minimum of 32, with no upper bound. Should this instead be `int(math.Max(32, float64(ntasks)))`?
Also I think we need to remove the last line of the comment `// Since we can only process one task at a time, grab only one.`.

petemoore:
Thanks for writing all the tests too, awesome! :)

taskmgr/mock_test.go
gregarndt: This is the exact call of the taskcluster-client-go queue client.  This is mocking those methods.

taskmgr/queueservice.go
gregarndt: I didn't realize you could name parameters in an interface definition, thanks.

gregarndt: See my other comment about what these methods are.  They are so that the taskcluster-client-go queue client can be mocked out and replaced with another queue-like object.  If we wanted to encapsulate this, we would need  a wrapper around the queue client, which might be unnecessary I think.

taskmgr/queueservice.go
gregarndt: Ah sorry, I copied and pasted this from the generic worker, then reworked it a little without updating the comment, good catch!

We want the minimum in the case ntasks is for some reason > 32.  For instance maybe it's a worker that can run 50 tasks in parallel, Azure limits us to retrieving 32 tasks at a time.  What this actually should be I think is (it should be ntasks - taskruns)
```
int(math.Min(32, float64(ntasks-len(taskRuns))))
```

taskmgr/queueservice_test.go
gregarndt: I just added enough characters to make it invalid.  If we want to remove characters to make it invalid and just make it a single character, that's fine by me too.

taskmgr/queueservice_test.go
gregarndt: yea I left a PR comment when I opened this realizing that there is a bit of duplication in here that could be extracted out.

gregarndt: I think that there are a couple of handlers just returning a single xml string, and a couple that are verifying the delete endpoint was called (I guess we would need to make a package variable that stores that state rather than a variable in the test method then), and a few handlers that return some invalid things or a bad status code.

taskmgr/queueservice_test.go
gregarndt: Ok, I'm up for adding comments for these if it makes things clearer.  Thanks for the tip!

Is it common to add docs to the tests in Go to describe what they're doing?  I tried adding comments in some of the tests to explain some of the data being used and why.  The test names I also tried to make clear, but I can definitely work on making them better if they're not.  In this case it's called "TestClaimTask" and looking at lines 610-616 that's basically what it does.  There is just a lot of boiler plate needed to set that up.

taskmgr/queueservice.go
gregarndt: soon this package will become "worker" , so it will either be worker.Queue or worker.TaskclusterQueue.

I'm thinking perhaps naming it "queueClient" (unexpected because I don't think we need to since it's only used in thi spackage)

taskmgr/mock_test.go
petemoore: d'oh!! Yes, sorry, ignore my comment! :D

taskmgr/queueservice.go
petemoore: OK!

taskmgr/queueservice.go
petemoore: Agreed - my mistake, it is correct as you have it. :smile_cat:

taskmgr/queueservice.go
petemoore: I'm so stupid - of course Min is correct - my mistake! I should only review code in the afternoons!!

Isn't `len(taskRuns)` always zero, because it is instantiated as an empty slice of `*TaskRun` just before this is evaluated?

taskmgr/mock_test.go
gregarndt: No problem, thanks for checking up on it! :)

taskmgr/queueservice_test.go
petemoore: It wasn't really a biggy, I just thought if it had non-base64 characters in it, it would be more immediately obvious that it isn't a valid base64 string, but it doesn't really matter. E.g. `<MessageText>invalid MessageText, not base64 encoded!</MessageText>` might be clearer, but the comment already explains it is not valid base64, so it really doesn't matter.

taskmgr/queueservice.go
gregarndt: Ah yea, I was thinking this method was doing something else.  Really what should happen is the caller of pollTaskUrl should keep calling that method until either A) it has as many tasks as it needs B) the queue returns 0

taskmgr/queueservice_test.go
gregarndt: Fixed this up I hope in 61dbeff0cfb697567ed34d69c26ce70ea85a0cf6

taskmgr/queueservice.go
gregarndt: I think this is fixed up in 19bd452

taskmgr/queueservice.go
gregarndt: 9a25eb0 should fix this up

taskmgr/task_status_updater.go
```diff
@@ -36,7 +36,7 @@ const (
 )
 
 type updateError struct {
-	StatusCode int
+	statusCode int
```
petemoore: Would it be useful to make these exported fields, so that if a caller of `UpdateTaskStatus` receives an error, they can inspect the http status code that caused the error? I believe at the moment they can get the message, and parse it to determine the status code, but maybe it is useful to expose it?

petemoore:
r+ :)

PR #23, opened 2016-02-12
destruction!!

petemoore:
Minor simplification. No need to use structs for single member types.

### DIFF
--- a/runtime/compositeschema.go
+++ b/runtime/compositeschema.go
@@ -26,9 +26,7 @@ type schemaEntry struct {
 	validator  *gojsonschema.Schema
 }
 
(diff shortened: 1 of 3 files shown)


runtime/compositeschema.go
```diff
@@ -26,9 +26,7 @@ type schemaEntry struct {
 	validator  *gojsonschema.Schema
 }
 
-type composedSchema struct {
-	entries []CompositeSchema
-}
+type composedSchema []CompositeSchema
```
jonasfj: Nice trick!!!

jonasfj:
R+

PR #25, opened 2016-02-18
Enginetest env vars

jonasfj:
I hope the new `run` struct in engine test will make it easier to write these tests...

I'll start looking at refactoring existing tests to use it... to make them simpler... Well maybe later...

Also have tests to write for the interactive features... and artifacts...


engines/mock/mockengine_test.go
```diff
@@ -96,3 +96,22 @@ func TestPing404IsUnsuccessful(t *t.T) { proxyTestCase.TestPing404IsUnsuccessful
...
+      "argument": "HELLO_WORLD"
+    }
+  }`,
+}
+
+func TestPrintVariable(t *t.T) { envVarTestCase.TestPrintVariable() }
+
+func TestVariableNameConflict(t *t.T) { envVarTestCase.TestVariableNameConflict() }
+func TestInvalidVariableNames(t *t.T) { envVarTestCase.TestInvalidVariableNames() }
+func TestEnvVarTestCase(t *t.T)       { envVarTestCase.Test() }
```
jonasfj: @petemoore again the idea with the `EnvVarTestCase` structure...
Is that all you have to do to write tests for an engine is these 15 lines :)

engines/enginetest/testutils.go
```diff
@@ -27,6 +30,12 @@ func evalNilOrPanic(f func() error, a ...interface{}) {
 	nilOrpanic(f(), a...)
 }
 
+func assert(condition bool, a ...interface{}) {
+	if !condition {
+		fmtPanic(a...)
+	}
+}
+
```
petemoore: I think @gregarndt has been using an assert library for this - we should probably align on either using our own or using this. Since assertions are not useful only for engine tests, this should probably go somewhere outside the engine test package.

engines/enginetest/testutils.go
```diff
@@ -114,3 +123,101 @@ func buildRunSandbox(b engines.SandboxBuilder) bool {
...
+	defer reader.Close()
+	nilOrpanic(err, "Failed to open log reader")
+	data, err := ioutil.ReadAll(reader)
+	nilOrpanic(err, "Failed to read log")
+	return string(data)
+}
+
+func (r *run) Dispose() {
+	if r.sandboxBuilder != nil {
+		nilOrpanic(r.sandboxBuilder.Discard(), "")
```
petemoore: Should probably be `nilOrPanic`

engines/errors.go
```diff
@@ -26,6 +26,9 @@ var ErrImmutableMountNotSupported = errors.New("The engine doesn't support immut
 // doesn't exist.
 var ErrResourceNotFound = errors.New("The referenced resource wasn't found")
 
+// ErrHandlerInterrupt is returned when a handler that was given returns and error
+var ErrHandlerInterrupt = errors.New("Handler returned an error and interrupted iteration")
```
petemoore: This error (like ErrResourceNotFound) seems not to contain relevant context. However, if we log the reference used in the log message (and store it outside the error) then that would be ok. It just feels wrong to have an error type that doesn't store the context it relates to, as it is not inspectable programmatically. However, if the context is logged at least, that is better than nothing.

engines/mock/mocksandbox.go
```diff
@@ -70,13 +74,29 @@ func (s *sandbox) AttachProxy(name string, handler http.Handler) error {
 	// Lock before we access proxies as this method may be called concurrently
 	s.Lock()
 	defer s.Unlock()
+	if strings.ContainsAny(name, " ") {
+		return engines.NewMalformedPayloadError("MockEngine proxy names cannot contain space")
+	}
```
petemoore: Would be better to include the proxy name that was used in the error.

`return engines.NewMalformedPayloadError(fmt.Sprintf("MockEngine proxy name %q should not contain space", name))`

engines/mock/mocksandbox.go
```diff
@@ -70,13 +74,29 @@ func (s *sandbox) AttachProxy(name string, handler http.Handler) error {
...
 	}
 	s.proxies[name] = handler
 	return nil
 }
 
+func (s *sandbox) SetEnvironmentVariable(name string, value string) error {
+	s.Lock()
+	defer s.Unlock()
+	if strings.Contains(name, " ") {
+		return engines.NewMalformedPayloadError("MockEngine env var names cannot contain space")
```
petemoore: Same here.

engines/resultset.go
```diff
@@ -55,8 +55,8 @@ type ResultSet interface {
 	// through as return value from the ExtractFolder call.
 	//
 	// If an error occurs during iteration iteration is halted, and when all
```
petemoore: iteration iteration

engines/resultset.go
```diff
@@ -55,8 +55,8 @@ type ResultSet interface {
...
-	// calls to handler(path, stream) has returned ExtractFolder may with return
-	// with the error.
+	// calls to handler(path, stream) has returned ExtractFolder should with
```
petemoore: as returned => have returned,
should with => should

engines/mock/mocksandbox.go
jonasfj: Yes... That would be better... This is just a mock engine... But yes lets do it anyway

engines/errors.go
jonasfj: This an an error that **must** be handled... It's non-fatal... So the caller has the context..

They are easier to handle when done like this... Hence, why I'm willing to not include any context

engines/enginetest/testutils.go
gregarndt: Whatever we consider best practices for contributing code to the worker should be documented in a conventions file somewhere for new contributors.

jonasfj: Yes and no, let's not make too much useless docs... :)

I'll take a look at the assert lib... I just wanted to keep this short... I'm less concerned when it's not exported...

engines/mock/mockengine_test.go
petemoore: Very neat!

engines/enginetest/testutils.go
petemoore: @jonasfj I've addressed this one separately in #27 as it wasn't introduced in this PR.

engines/enginetest/testutils.go
jonasfj: @petemoore, @gregarndt, I don't want to use the `assert` library from:
https://godoc.org/github.com/stretchr/testify/assert

It requires `t` be passed around.. and I'm too lazy for that... Honestly I'm quiet happy just doing a `panic()` it also gives me a stack trace... which is really what I want :)

Note, IMO, it's okay to define these non-exported helpers whenever we feel like it... This assert and nilOrPanic stuff is local to the `enginetest` package and can't be used outside it. I use them to keep the code concise otherwise the generic test cases I implement would simply be huge.

Note: also `panic` works from a go rutine.. `t` can only be called from the go rutine that called the `Test...` method.

engines/resultset.go
jonasfj: Missing comma...

jonasfj:
@petemoore, I don't think it was nice of me to add more tests while doing this... Sorry :)

I suspect we might still revise the shell interface... Transform it into something that support executing arbitrary sub-processes... I'm not quiet sure yet... Or we add an generic sub-process extension and add TTY width/height control to the Shell interface. I suspect it might need that...

Regardless this is a good start...

jonasfj:
There is also an outline for display test cases in there... I'll add them at once point... I can make generic as I know what the first few bytes of a VNC connection looks like..

petemoore:
@jonasfj :+1: r+

PR #28, opened 2016-02-23
Bug 1246643 - auto generate global config struct from json schema yaml file

petemoore:
This should be mostly a no-op change, but generating the config struct from the checked in json schema.
When we are have this in place, we'll be in a position to validate the config against the json schema, and also publish the json schema for documentation (e.g. if the taskcluster worker page on our docs page wants to explain how the worker should be configured, e.g. if RelEng wants to install their own workers on hardware without using the AWS provisioner).


petemoore:
@jonasfj FYI you may be interested in this PR too...

petemoore:
And obviously we can take advantage of the CompositeSchema mechanics to do the loading/parsing/validation etc....

codegen/go-composite-schema/main.go
```diff
@@ -50,11 +50,13 @@ go-composite-schema to correctly determine the correct package name.
...
 
   Usage:
-    go-composite-schema [--required] PROPERTY INPUT-FILE OUTPUT-FILE
+    go-composite-schema [--required] [--unexported] PROPERTY INPUT-FILE OUTPUT-FILE
     go-composite-schema -h|--help
     go-composite-schema --version
 
   Options:
+    --required             Implies the given composite schema defines a required property.
+    --unexported           Generate an unexported type for the given composite schema.
```
jonasfj: We could just make this implied from the `title` property... Note: I still propose introducing a `TypeName` property in the JSON schema, so `title` only is used as a fallback.

jonasfj: Not that we need to block this PR on this!!!

It's just a nice to have some day when we feel like we have time...

config/doc.go
```diff
@@ -0,0 +1,4 @@
+//go:generate go-composite-schema --required config global-config.yml config.go
+//
+// Package config stores the config for the taskcluster-worker
+package config
```
jonasfj: I propose this lives in the `/worker/` folder/package... currently named taskmgr

config/config.go
```diff
@@ -1,16 +1,139 @@
...
+		  "$schema": "http://json-schema.org/draft-04/schema#",
+		  "additionalProperties": false,
+		  "description": "The global configuration settings used by taskcluster-worker, not particular\nto any engine or plugin.",
+		  "properties": {
+		    "capacity": {
+		      "description": "The number of tasks that this worker supports running in parallel.",
+		      "title": "Capacity",
+		      "type": "integer"
+		    },
+		    "credentials": {
```
gregarndt: we can tweak this later, but I wonder if there will be other credentails we will be storing on the worker so it might make sense to put it under it's own key

PR #29, opened 2016-02-23
Update travis to use go 1.5 instead of go 1.4

petemoore:
Let's see if we start getting race errors in travis too now....


jonasfj:
@gregarndt, @petemoore, should we not just move to 1.6, I don't see any reasons not to... eventually we'll probably upgrade locally too :)

PR #31, opened 2016-02-27
Merge schemas

jonasfj:
Based on #30 

This adds a `Schema` function that returns the schema for a composite schema...
As it can operate on the interface definition I made it function instead of a member method...

I think this is the pattern when you implement abstract methods using virtual methods only...

Also added tests... Now we can add command line interface to print config schema and payload schema etc... :)


jonasfj:
@petemoore ping... Not that there is any rush :)

petemoore:
Awesome!

PR #32, opened 2016-03-03
Worker loop

gregarndt:


worker/task_manager.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+
+	task.plugin = taskPlugins
+
+	// each method can return error channel, and the next stage can consume it
+	prepare, errc := task.Prepare(m.engine)
+	build, errc := task.Build(prepare, done, errc)
+	run, errc := task.Run(build, done, errc)
+	stop, errc := task.Stop(run, done, errc)
+	finish, errc := task.Finish(stop, done, errc)
```
jonasfj: interesting.... So you're sticking them up so they all run... no matter what happens...

worker/task.go
```diff
@@ -0,0 +1,201 @@
...
+func (t *TaskRun) Stop(in <-chan engines.ResultSet, done <-chan struct{}, inerrc <-chan error) (<-chan bool, <-chan error) {
+	out := make(chan bool, 1)
+	errc := make(chan error, 1)
+
+	go func() {
+		defer func() {
+			close(errc)
+			close(out)
+		}()
+		r := <-in
```
jonasfj: wouldn't you need to handled the case where `r` is `nil` ? because there was an error and the input channel was closed..

worker/task.go
```diff
@@ -0,0 +1,201 @@
...
+	context         *runtime.TaskContext
+	plugin          plugins.TaskPlugin
+	log             *logrus.Entry
+	payload         interface{}
+}
+
+func (t *TaskRun) Prepare(engine engines.Engine) (<-chan engines.SandboxBuilder, <-chan error) {
+	out := make(chan engines.SandboxBuilder, 1)
+	errc := make(chan error, 1)
+	go func() {
```
jonasfj: So I see how these channels and go routines are all stitched together... Normally, you would never do this in any other language, because threads are expensive.

What is going on here is that we are starting 5 threads, where thread N waits on thread N - 1 before starting. So there is no need for the concurrency, it is all inherently serial.
Given that go routines are super cheap this is not slow or resource intensive in any concerning way.

And I won't dare to claim that channels and go routines can't be used to structure code in a convenient manner. I'm sure there are good design pattern for serial code which relies on channels and go routines. But channels and go routines are generally tools for dealing with concurrency, so maybe we should look for other ways to structure this.

I'll try to outline something here:
https://public.etherpad-mozilla.org/p/jonasfj-taskcluster-worker-task-controller

worker/task_manager.go
gregarndt: Yea so this attempt was to see what a flow of stages looked like, where a task will always end up in the finished stages regardless of what happened in any of the intermediary stages.  It's probably completely crazy, just something I was experimenting with

worker/task.go
gregarndt: Yea let's talk about it.  One of my commits started going down the path of everything being serial, and that worked.  But then when we added error handling, and especially the case of having to abort somewhere in there made things a bit more muddled.  

I'm excited to discuss it though when you have a moment.

worker/task.go
```diff
@@ -0,0 +1,284 @@
...
+	t.controller = contextController
+	t.Unlock()
+
+	defer t.DisposeStage()
+
+	jsonPayload := map[string]json.RawMessage{}
+	if err := json.Unmarshal(t.Definition.Payload, &jsonPayload); err != nil {
+		t.context.LogError(fmt.Sprintf("Invalid task payload. %s", err))
+		t.ExceptionStage(runtime.Errored, err)
+		return
```
jonasfj: Okay, how about `ExceptionStage` also returns the error if it is an internal error... And then we return the error from `Run`.

Specifically, I think we'll want to shutdown (kill the ec2 instance) in the face of internal errors, or have some logic that counts the number of internal errors and shutdown after it goes really bad...

Regardless we probably don't want to deal with that here so let's propagate internal errors up (after reporting the task as an exception and cleaning up, etc...)...

That way this also becomes `return t.ExceptionStage(runtime.Errored, err)` which is nice :)

jonasfj:
From a quick look this seems like the right direction :)

Still some channels, but I think that's when interfacing QueueService, we can look at that later... Maybe there is also stuff there that I don't fully understand yet.

PR #33, opened 2016-03-04
[WIP] Started work on windows engine

petemoore:
Note: **"Started"**, not "Finished" :stuck_out_tongue_winking_eye:


petemoore:
@jonasfj @gregarndt Note, the some stuff in the config should probably be not engine-specific (i.e. sit outside of engine config) and some stuff in the payload should be handled by plugins, not by the engine. To start the ball rolling, I ported over the config and payload used by the generic worker, so we can pull stuff around into the right place as part of this PR...

engines/winnative/payload-schema.yml
```diff
@@ -0,0 +1,57 @@
...
+  command: 
+    title: "Commands to run"
+    type: "array"
+    minItems: 1
+    items: 
+      type: "array"
+      minItems: 1
+      items: 
+        type: "string"
+    description: "One array per command (each command is an array of arguments). Several arrays for several commands."
```
jonasfj: I think we should restrict it to one command in total. I know it sucks, but it provides better consistency across platforms...

Also we don't want people uploading scripts as part of the task-definition. I would rather focus on enabling preloaded caches...

jonasfj:
Seems like a solid start :)

If you are crazy you could setup the `enginetest` cases... just copy paste from the mock engine tests... and change the payload definition to be what you expect..

engines/winnative/payload-schema.yml
petemoore: In linux, it is relatively straightforward to exit-on-error, e.g. in bash with `#!/bin/bash -e` or `set -e`. On Windows, there is no equivalent, so for most tasks, which will be a sequence of commands, it will be non-trivial for tasks to exit if one of the steps fail. Typically you also need at least two steps: one to checkout sources, and the second step to execute a script. I'm not sure if there is an equivalent construct to `&&` for batch scripts.

jonasfj: just install bash on windows... it's part of mozbuild anyways...

note: you were never intended to specify the entire task script in the `task.payload`. The `task.payload` property is only for specifying the arguments for whatever task script you have. The task script should be part of the image, or a script downloading the task script from hg/git should be part of the image.

Tasks are not scripts. That will just make people sad when they hit arbitrary limits. So we shouldn't let people do that... If people really want to using `&&` then it pretty obvious that this is not the intended design.

walac: I have provided the same behavior for Mac OSX, just for the case of consistency, but I rather prefer a single command. One annoying thing is that I have to specify the word `payoad` twice:

```json
"payload" : {
   "payload": {
       "command": [
       ],
      "maxRunTime": 1000
   }
}
```

engines/winnative/generated_payloadschema.go
```diff
@@ -0,0 +1,134 @@
...
+var payloadSchema = func() runtime.CompositeSchema {
+	schema, err := runtime.NewCompositeSchema(
+		"payload",
+		`
+		{
+		  "$schema": "http://json-schema.org/draft-04/schema#",
+		  "description": "This schema defines the structure of the `+"`"+`payload`+"`"+` property referred to in a Task Cluster Task definition.",
+		  "id": "http://schemas.taskcluster.net/generic-worker/v1/payload.json#",
+		  "properties": {
+		    "artifacts": {
```
jonasfj: artifacts aren't part of the engine payload

engines/winnative/payload-schema.yml
```diff
@@ -0,0 +1,57 @@
...
+    description: "Example: ```{ \"PATH\": \"C:\\\\Windows\\\\system32;C:\\\\Windows\", \"GOOS\": \"darwin\" }```"
+    type: "object"
+  maxRunTime: 
+    type: "integer"
+    title: "Maximum run time in seconds"
+    description: "Maximum time the task container can run in seconds"
+    multipleOf: 1
+    minimum: 1
+    maximum: 86400
+  artifacts:
```
jonasfj: sorry this is the place to say artifacts aren't part of the engine payload... they are handled by plugins.

engines/winnative/payload-schema.yml
```diff
@@ -0,0 +1,57 @@
...
+    title: "Commands to run"
+    type: "array"
+    minItems: 1
+    items: 
+      type: "array"
+      minItems: 1
+      items: 
+        type: "string"
+    description: "One array per command (each command is an array of arguments). Several arrays for several commands."
+  env:
```
jonasfj: not part of engine payload

engines/winnative/payload-schema.yml
jonasfj: let's discuss this tomorrow... i know single command is a pain... but really tasks shall not carry data... scripts are data... task should at most carry metadata like arguments for a script

walac:
Closing as this became obsolete.

PR #36, opened 2016-03-08
generate a valid call to NewCompositeSchema when not required.

walac:
If don't specify a schema as required, composite-schema will not
generate the boolean required argument to NewCompositeSchema.

Fix this always reflecting the boolean value of req parameter into the
call NewCompositeSchema.


codegen/go-composite-schema/main.go
```diff
@@ -155,9 +155,7 @@ func generateFunctions(ymlFile, goType, schemaProperty string, req bool) string
 	// the following strings.Replace function call safely escapes backticks (`) in rawJson
 	result += strings.Replace(text.Indent(fmt.Sprintf("%v", string(rawJson)), "\t\t")+"\n", "`", "` + \"`\" + `", -1)
 	result += "\t\t`,\n"
-	if req {
-		result += "\t\ttrue,\n"
-	}
+	result += fmt.Sprintf("\t\t%t,\n", req)
```
petemoore: Nice spot!

petemoore:
Wow, not sure how that got in there in the first place, thanks for fixing.

PR #38, opened 2016-03-08
Added sentry

jonasfj:
I think this will work even when client is `nil`:
https://github.com/getsentry/raven-go/blob/master/client.go#L432-L435

So just adding it like this we can already start reporting exceptions, and deal with setting it up later...


imbstack:
:+1: :question:

PR #39, opened 2016-03-08
Implement env plugin.

walac:
env plugin is responsible to process the section "env" from the task
payload. It setups environment variables inside the sandbox.


plugins/env/payload-schema.yml
```diff
@@ -0,0 +1,6 @@
+$schema:            http://json-schema.org/draft-04/schema#
+title:              "EnvPayload"
+type:               "object"
+patternProperties:
+  .+:
+    type: string
```
jonasfj: Don't use pattern properties...
Just use `additionalProperties: {type: string}`

Also add some "description" properties, ideally with a description that can serve as documentation... It doesn't have to be long, but a line, two or 5 would be good. Usually using markdown in description properties is fine, we do it everywhere...

walac:
Why must `generated_*.go` files be checked in?

jonasfj:
> Why must generated_*.go files be checked in?

Because they are hard to generate... there are pros/cons... but they aren't generated by `go build` so it's nice to do... especially because it prevents anyone from using it as a library..

plugins/env/payload-schema.yml
walac: Fixed.

walac:
Implementation changed based on what @petemoore, @jonasfj and I discussed.

plugins/env/payloadschema.go
```diff
@@ -0,0 +1,40 @@
...
+		    "type": "string"
+		  },
+		  "description": "env payload section contents",
+		  "title": "EnvPayload",
+		  "type": "object"
+		}
+		`,
+		false,
+		func() interface{} {
+			return &envPayload{}
```
jonasfj: Does this build?
I would suggest `make(map[string]string)` and not define a type `envPayload`

plugins/env/env.go
```diff
@@ -0,0 +1,59 @@
...
+type taskPlugin struct {
+	plugins.TaskPluginBase
+	payload envPayload
+}
+
+func (self taskPlugin) BuildSandbox(sandboxBuilder engines.SandboxBuilder) error {
+	for k, v := range self.payload {
+		err := sandboxBuilder.SetEnvironmentVariable(k, v)
+		if err != nil {
+			return err
```
jonasfj: Most errors here should be transformed to `MalformedPayloadError` see: https://github.com/taskcluster/taskcluster-worker/blob/2130156c8e55bdc710af9a7c16e5877844ef7f17/engines/sandboxbuilder.go#L64-L73

Basically, `ErrNamingConflict` is a problem that can be solved by changing the task payload. The same is the case for `ErrFeatureNotSupported`, if env vars aren't support... user just shouldn't specify any...

jonasfj: Errors other than the ones specified as "Non-fatal errors" in https://github.com/taskcluster/taskcluster-worker/blob/2130156c8e55bdc710af9a7c16e5877844ef7f17/engines/sandboxbuilder.go#L64-L73

Can be passed through as you have no possible way to handle those gracefully.

jonasfj:
`r+`

plugins/env/env.go
```diff
@@ -0,0 +1,71 @@
...
+
+		// We can only return MalFormedPayloadError
+		switch err {
+		case engines.ErrNamingConflict:
+			return engines.NewMalformedPayloadError("Environment variable ", k, " has already been set.")
+		case engines.ErrFeatureNotSupported:
+			return engines.NewMalformedPayloadError(
+				"Cannot set environment variable ",
+				k,
+				". Engine does not support this operation")
```
petemoore: won't this introduce a space between the value of `k` and the `.`?

petemoore:
This looks awesome! Tests too! ++

PR #40, opened 2016-03-08
A quick update to the README and adding a bootstrap method to the Mak…

selenamarie:
…efile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,13 @@
 # Ensure go-extpoints and go-import-subtree are available for go generate
 export PATH := $(GOPATH)/bin:$(PATH)
 
-bootstrap:
+dev:
+	go get github.com/axw/gocov/gocov
+	go get github.com/golang/lint/golint
+	go get github.com/mattn/goveralls
+	go get github.com/pierrre/gotestcover
 	go get github.com/stretchr/testify/assert
+	go get golang.org/x/tools/cmd/cover
 	go get golang.org/x/tools/cmd/vet
 
 all: build
@@ -30,7 +35,6 @@ check: test
 	/bin/bash -c 'test $$(git status --porcelain | wc -l) == 0'
 test:
 	go get -t ./...
-	go get github.com/golang/lint/golint
 	go test -race ./...
 	go vet ./...
 	golint ./...
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,10 @@
 # Ensure go-extpoints and go-import-subtree are available for go generate
 export PATH := $(GOPATH)/bin:$(PATH)
 
+bootstrap:
+	go get github.com/stretchr/testify/assert
+	go get golang.org/x/tools/cmd/vet
+
 all: build
 build:
 	go fmt ./...
--- a/README.md
+++ b/README.md
@@ -10,6 +10,27 @@ TaskCluster Worker
 A generic worker for task cluster, written in go.
 
 
+Hacking
+=======
+
+How to get started
+------------------
+
+1. Install a version of golang > 1.4: [Download Go](https://golang.org/dl/) and set up your environment. 
+2. Get the taskcluster-worker source.  (Optionally: fork the code first, and then substitute your user name in for taskcluster below)
+
+```
+go get github.com/taskcluster/taskcluster-worker
+
+```
+
+cd into the taskcluster-worker directory.
+
+3. Run: `make bootstrap ; make test`
+
+4. Start hacking!
+
+
 ** Warning very EARLY notes**
 
 Place we outline interfaces and patterns for next generation worker.
@@ -17,14 +38,10 @@ Place we outline interfaces and patterns for next generation worker.
 ```
 Goal: have all interfaces and central classes defined and documented.
 
-
-
-
 Things we sort of need:
   - garbage collector...
   - state machines... (http://www.timschmelmer.com/2013/08/state-machines-with-state-functions-in.html)
     - basically a state is a function that takes an action and returns a new function (state)
-
 
 
 state machine:


Makefile
```diff
@@ -6,6 +6,10 @@
 # Ensure go-extpoints and go-import-subtree are available for go generate
 export PATH := $(GOPATH)/bin:$(PATH)
 
+bootstrap:
+	go get github.com/stretchr/testify/assert
```
gregarndt: Is testify needed here for bootstrapping?  when running 'make test' it should get all the packages required by the taskcluster-worker

Makefile
```diff
@@ -6,6 +6,10 @@
 # Ensure go-extpoints and go-import-subtree are available for go generate
 export PATH := $(GOPATH)/bin:$(PATH)
 
+bootstrap:
```
gregarndt: I wonder if something could be conslidated here with bootstrapping: 
https://github.com/taskcluster/taskcluster-worker/blob/master/.travis.yml#L21

gregarndt: Sorry, I was a lot less than clear with this comment.  What I'm referring to is the "bootstrapping" that is done in the travis environment and I wonder i the same needs to be done here in this make target.

Makefile
```diff
@@ -6,6 +6,15 @@
 # Ensure go-extpoints and go-import-subtree are available for go generate
 export PATH := $(GOPATH)/bin:$(PATH)
 
+dev:
+	go get github.com/axw/gocov/gocov
+	go get github.com/golang/lint/golint
+	go get github.com/mattn/goveralls
+	go get github.com/pierrre/gotestcover
+	go get github.com/stretchr/testify/assert
```
petemoore: I think line 14 is automatically handled by line 37.

Makefile
```diff
@@ -6,6 +6,15 @@
 # Ensure go-extpoints and go-import-subtree are available for go generate
 export PATH := $(GOPATH)/bin:$(PATH)
 
+dev:
+	go get github.com/axw/gocov/gocov
+	go get github.com/golang/lint/golint
+	go get github.com/mattn/goveralls
```
petemoore: I think line 12 is only needed by travis - this is for publishing coverage reports to coveralls.

README.md
```diff
@@ -10,15 +10,33 @@ TaskCluster Worker
...
-** Warning very EARLY notes**
+Hacking
+=======
 
-Place we outline interfaces and patterns for next generation worker.
+How to get started
+------------------
+
+1. Install a version of golang > 1.4: [Download Go](https://golang.org/dl/) and set up your environment. 
+2. Get the taskcluster-worker source.  (Optionally: fork the code first, and then substitute your user name in for taskcluster below)
```
petemoore: Due to explicit package naming, the workflow would need to be:
* Fork
* Go get from taskcluster repo (so keep line 23 the same)
* Go to the checkout directory under `taskcluster` directory, add a separate remote for your fork (i.e. don't go get from your local fork)
* Work, and when you push, push to your fork (the new remote you added)

This is unfortunately the best workflow to use, since if you fork and `go get` your fork, the go imports will import the wrong version.

README.md
```diff
@@ -10,15 +10,33 @@ TaskCluster Worker
...
 
 ```
-Goal: have all interfaces and central classes defined and documented.
+go get github.com/taskcluster/taskcluster-worker
 
+```
+
+cd into the taskcluster-worker directory.
 
+3. Run: `make bootstrap ; make test`
```
petemoore: Should this be `make dev test`?

petemoore:
Thanks Selena for helping out with the docs - really important we have decent ones, so thanks for this. :+1:

Makefile
selenamarie: It isn't! I tried running this without it and got an error about assert being missing.

Makefile
selenamarie: Ah ok.

Makefile
gregarndt: @petemoore what trickery is going on???? :)  I thought line 37 should fetch all packages we import.

petemoore: Me too! I can have a look on Monday.

petemoore: @selenamarie: did you run `make dev` first?

petemoore: Ah that also shouldn't matter.. hmmmm I think I'll look on Monday.

petemoore: Works for me. @selenamarie can you try running this script? It builds a fresh version, with lines 12 and 14 deleted.

```bash
#!/bin/bash -xv

cd "$(dirname "${0}")"
rm -rf tmpgopath
mkdir tmpgopath
export GOPATH=$(pwd)/tmpgopath
go get github.com/taskcluster/taskcluster-worker
cd tmpgopath/src/github.com/taskcluster/taskcluster-worker
git remote add selena git@github.com:selenamarie/taskcluster-worker.git
git fetch selena
git checkout readme-update
cat Makefile | sed -e 12d -e 14d > m
mv m Makefile
git merge -m "pulling in latest changes" origin/master # needed for compatibility with latest taskcluster-client-go
make dev test
echo "'make dev test' exit code was: $?"
cd ../../../../..
rm -rf tmpgopath
```

Let me know if you get:
```
'make dev test' exit code was: 0
```

petemoore:
The go cover packages are only needed if you run coverage reports, which we don't do in the make targets - so I'm not sure we need to add them there. We produce coverage reports on travis, so I think people are more likely to refer to those rather than produce locally. If we really want to support coverage reports from a local dev environment, we should update/add make targets to produce the coverage reports, and generate e.g. html files that are ignored in ~/.gitignore, but can be inspected locally.

When `go get` runs against a package that is already present locally, go will skip over it, without doing e.g. a `git pull` (you would need the `-u` option for that). Therefore running `go get` against a package you already have checked out has effectively no overhead, other than a couple of milliseconds needed to check if the folder exists locally - therefore I'd prefer to keep these commands in the `make` targets that depend on them, rather than introducing a separate `dev` target. It keeps it simpler for the developer, as they won't get failures if they haven't manually run the `dev` target, and also ensures if for any reason a dependency gets deleted from their workspace, they will fetch it again if they need it.

petemoore:
I think we can close this now. In the meantime the README has been quite heavily updated.

@gregarndt ok to close?

PR #43, opened 2016-03-08
Adding engine config to enginetest

jonasfj:


gregarndt:
:+1:

gregarndt:
If you're cool with this, click merge.  I think your recent changes for the go fmt issue will fix things up.

PR #45, opened 2016-03-09
Added *ProviderBase

jonasfj:


engines/extpoints/interfaces.go
```diff
@@ -40,3 +40,15 @@ type EngineProvider interface {
 	// configuration
 	ConfigSchema() runtime.CompositeSchema
 }
+
+// EngineProviderBase is a base struct that provides empty implementations of
+// some methods for EngineProvider
+//
+// Implementors of EngineProvider should embed this struct to ensure forward
+// compatibility when we add new optional method to EngineProvider.
+type EngineProviderBase struct{}
```
petemoore: This should be called `ProviderBase`.

engines/mock/mockengine.go
```diff
@@ -18,7 +18,9 @@ type engine struct {
 	Log *logrus.Entry
 }
 
-type engineProvider struct{}
+type engineProvider struct {
+	extpoints.EngineProviderBase
```
gregarndt: you should remove ConfigSchema() from this package then (line 35)

petemoore:
Thanks

PR #49, opened 2016-03-09
Return channel when waiting for shutdown

gregarndt:


runtime/shutdownmgr.go
```diff
@@ -25,22 +29,24 @@ func NewShutdownManager(host string) ShutdownManager {
 
 // AWSShutdownManager is a ShutdownManager that will listen for shutdowns on the notification
 // api provided by AWS.
-type AWSShutdownManager struct{}
+type AWSShutdownManager struct {
```
jonasfj: Why do you export this type?

runtime/shutdownmgr.go
```diff
@@ -25,22 +29,24 @@ func NewShutdownManager(host string) ShutdownManager {
...
-	c := make(chan struct{})
-	return c
+func (a *AWSShutdownManager) WaitForShutdown() <-chan struct{} {
+	return a.sc
 }
 
 // LocalShutdownManager simple ShutdownManager that could listen for shutdown events
 // suitable for a local non-cloud environment (such as SIGTERM).
-type LocalShutdownManager struct{}
+type LocalShutdownManager struct {
```
jonasfj: Same for this one...

gregarndt: Both have been changed

gregarndt:
merge if I addressed your comments

PR #50, opened 2016-03-09
Engine docs fixes.

walac:


engines/sandbox.go
```diff
@@ -87,12 +87,12 @@ type Sandbox interface {
...
 	OpenDisplay(name string) (io.ReadWriteCloser, error)
 
-	// Abort the sandbox, this means killing the task execution as well as all
+	// Abort the sandbox. This means killing the task execution as well as all
 	// associated shells and releasing all resources held.
 	//
 	// If called before the sandbox execution finished, then WaitForResult() must
 	// return ErrSandboxAborted. If sandbox execution has finished when Abort() is
-	// called Abort() should return ErrSandboxTerminated and not release any
+	// called Abort(), should return ErrSandboxTerminated and not release any
```
petemoore: I think the comma should be after `called`.

petemoore:
Thanks for the fixes!

engines/sandbox.go
walac: Oops.

PR #51, opened 2016-03-10
Refector queue service

gregarndt:
Some high level goals are to:

[X] Return a channel which task claims will be written to
[X] Separate out task claims, task messages, and task runs into their own types to make it clearer
[X] Add method for creating a new task run from a task claim

### DIFF
--- a/main.go
+++ b/main.go
@@ -115,9 +115,5 @@ func main() {
 		logger.Fatalf("Could not create worker. %s", err)
 	}
 
-	err = w.Start()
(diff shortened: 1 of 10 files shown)


gregarndt:
@jonasfj mind reviewing this? I renamed some things, tried to clean up what was exported, and created a method for generating a new task run.

jonasfj:
`r+`, sorry it took a while to get around to it... But it looks really solid...

PR #57, opened 2016-03-16
Swap out ReadCloser for ReadSeekCloser to support retries in artifact uploads

imbstack:
This will make implementing certain engines (like the docker engine) more difficult, and likely require tmp files to be created for artifacts, but will make the implementation of artifact uploading much simpler. This PR exists for discussion about whether or not this would be preferable. @jonasfj and @gregarndt are both interested parties.

### DIFF
--- a/engines/resultset.go
+++ b/engines/resultset.go
@@ -1,6 +1,16 @@
 package engines
 
-import "io"
+import (
+	"io"
+)
+
+// A ReadSeekCloser wraps some basic io operations such that we can do
+// retries in artifact upload logic
+type ReadSeekCloser interface {
+	io.Reader
+	io.Seeker
+	io.Closer
+}
 
 // FileHandler is given as callback when iterating through a list of files.
 //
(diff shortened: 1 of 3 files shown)


engines/resultset.go
```diff
@@ -1,6 +1,16 @@
 package engines
 
-import "io"
+import (
+	"io"
+)
+
+// A ReadSeekCloser wraps some basic io operations such that we can do
+// retries in artifact upload logic
+type ReadSeekCloser interface {
```
jonasfj: Maybe:
> ReadSeekCloser implements io.Reader, io.Seeker, and io.Closer and is trivially implemented by os.File.
> This interface is used to expose file objects from containers.

After all the main purpose is to expose a file object of some form.

gregarndt:
This seems ok with me.  Perhaps as we're doing something with docker, we can figure out what works out best but for now this should be cool since files will exist in all other engines rather than within a container that's streamed.

jonasfj:
Looks good to me `r+`.

@gregarndt any comments?

engines/resultset.go
gregarndt: Since this is an exported interface, doesn't the interface already document what it implements?  

I'm confused by "This interface is used to expose file objects from containers." since "containers" is a very linux container thing and this same interface would be used when extracting files from a windows environment, right? Or from a file created on a host by a linux/os x process?

jonasfj: True... My wording isn't perfect either...

Maybe this is better:

> ReadSeekCloser implements io.Reader, io.Seeker, and io.Closer. It is trivially implemented by os.File.
> The interface is used to expose file objects from ResultSet while abstracting with file-system specifics.

PR #58, opened 2016-03-17
CommandProvider as proposed by @petemoore

jonasfj:
What do you think... Will this do?


petemoore:
Perfect! Thanks for implementing! That is exactly as I had imagined it.

PR #59, opened 2016-03-17
Create queue client in task context

gregarndt:
This PR allows:
1. Task Context to have a queue client
2. Queue client can be updated safely by a claimer/reclaimer
3. Anything with the task context reference can get the queue client with the claims credentials
4. Base URL for the Queue can be defined in the config.  Right now this is mostly just for mocking out the queue endpoint for tests, but could be used when the queue does not live in the canonical place any longer.

### DIFF
--- a/runtime/taskcontext.go
+++ b/runtime/taskcontext.go
@@ -101,22 +101,22 @@ func (c *TaskContextController) Dispose() error {
 	return c.logStream.Remove()
 }
 
-// GetQueueClient will return a client for the TaskCluster Queue.  This client
-// is useful for plugins that require interactions with the queue, such as creating
-// artifacts.
-func (c *TaskContext) GetQueueClient() QueueClient {
-	c.mu.RLock()
-	defer c.mu.RUnlock()
-	return c.queue
-}
-
-// CreateQueueClient will create a client for the TaskCluster Queue.  This client
+// SetQueueClient will set a client for the TaskCluster Queue.  This client
 // can then be used by others that have access to the task context and require
 // interaction with the queue.
-func (c *TaskContext) SetQueueClient(client QueueClient) {
+func (c *TaskContextController) SetQueueClient(client QueueClient) {
 	c.mu.Lock()
 	c.queue = client
 	c.mu.Unlock()
+}
+
+// Queue will return a client for the TaskCluster Queue.  This client
+// is useful for plugins that require interactions with the queue, such as creating
+// artifacts.
+func (c *TaskContext) Queue() QueueClient {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.queue
 }
 
 // Abort sets the status to aborted
--- a/runtime/taskcontext.go
+++ b/runtime/taskcontext.go
@@ -7,6 +7,7 @@ import (
 
 	"sync"
 
+	"github.com/taskcluster/taskcluster-client-go/queue"
 	"github.com/taskcluster/taskcluster-worker/runtime/webhookserver"
 
 	"gopkg.in/djherbis/stream.v1"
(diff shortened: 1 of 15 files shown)


runtime/taskcontext.go
```diff
@@ -55,6 +60,7 @@ type TaskContext struct {
 	webHookSet *webhookserver.WebHookSet
 	logStream  *stream.Stream
 	mu         sync.RWMutex
+	queue      *queue.Queue
```
jonasfj: It would be great if we had an interface type for queue.Queue here... :)

jonasfj: So I'm tempted to do: `Queue: queue.QueueInterface`

Sure someone could do `context.Queue = nil`, but that's just pure evil...

gregarndt: I already have an interface used elsewhere so that I can mock it out when needed, I plan to just move it here into the runtime

gregarndt:
Right now Travis is failing because of bandwidth limits on googlesource.com

gregarndt:
@jonasfj  Any thoughts on this?

gregarndt:
So this was updated to make use of the new mocked objects in the library but tests are failing because the generated interfaces are wrong for the Queue client.

jonasfj:
@gregarndt, do we want to merge this without fixing the mocked objects?

gregarndt:
That's entirely up to the original author of the mocked objects PR :)  This particular PR was not adding auto generated mocked objects for all clients in taskcluster-client-go but adding as needed.  

We can still fix the mocked object issue and updating this stuff would take less than a minute really (I did it before when I thought the generated interfaces might work)

Up to you :)

runtime/taskcontext.go
```diff
@@ -89,6 +101,24 @@ func (c *TaskContextController) Dispose() error {
...
+func (c *TaskContext) GetQueueClient() QueueClient {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.queue
+}
+
+// CreateQueueClient will create a client for the TaskCluster Queue.  This client
+// can then be used by others that have access to the task context and require
+// interaction with the queue.
+func (c *TaskContext) SetQueueClient(client QueueClient) {
```
jonasfj: Let's move this method to the `TaskContextController`

runtime/taskcontext.go
```diff
@@ -89,6 +101,24 @@ func (c *TaskContextController) Dispose() error {
 	return c.logStream.Remove()
 }
 
+// GetQueueClient will return a client for the TaskCluster Queue.  This client
+// is useful for plugins that require interactions with the queue, such as creating
+// artifacts.
+func (c *TaskContext) GetQueueClient() QueueClient {
```
jonasfj: I propose calling this `Queue()`

gregarndt: Yea that's probably a good thing, thanks

runtime/taskcontext.go
gregarndt: ah right, just so no one is naughty and updating the client themselves

runtime/taskcontext.go
gregarndt: fixed in bc5da5e

runtime/taskcontext.go
gregarndt: fixed in bc5da5e

imbstack:
Looks good to me!

jonasfj:
> That's entirely up to the original author of the mocked objects PR
Lol, the original author recolonizes that and realizes that this PR isn't blocked on it...

runtime/taskcontext.go
```diff
@@ -68,13 +76,17 @@ type TaskContextController struct {
 }
 
 // NewTaskContext creates a TaskContext and associated TaskContextController
-func NewTaskContext(tempLogFile string) (*TaskContext, *TaskContextController, error) {
+func NewTaskContext(tempLogFile string, claim *queue.TaskClaimResponse) (*TaskContext, *TaskContextController, error) {
```
jonasfj: @imbstack would you not wish to avoid having to mock that?
Isn't easier if we just pass in `TaskInfo` ?

jonasfj: Then it's more obvious when we write tests that needs to fake it... Meaning it's more obvious what they have to provide...

jonasfj:
`r+`

runtime/taskcontext.go
gregarndt: I don't know why it didn't occur to me, but creating the &taskInfo{} object elsewhere and passing that in might be a lot easier/nicer.

imbstack:
r+

jonasfj:
okay, enough talk let's land...

gregarndt:
What if I wasn't ready???? I like pushing green buttons.

PR #60, opened 2016-03-17
Added interfaces and mock objects

jonasfj:
Could be that this is better added to `taskcluster-client-go`

@petemoore, what do you think?


petemoore:
I think it is ok to be added to a client package of taskcluster-client-go that wants to mock. I agree other clients may also want to mock but if we want to share them, maybe we should create a project taskcluster-mocks? I like not forcing client users to have to use the interfaces, but be able to use the methods directly if they wish, as they can always mock with an interface on their side, if they want to.

There is also an argument we don't need to generate mocks for everything, but I see the benefit in auto-generating to a) keep in sync with upstream, and b) avoid manual mistakes. However, a benefit of not autogenerating as that we'd just mock the methods we call, and then we'd have a smaller interface showing just the methods we use. Anyway, r+ - we can land as is if you like. :+1:

gregarndt:
I'm good with generating this stuff even though it really took about 10 seconds by hand to build what I needed to for other things.  It seems like this is hugely overkill for what little we need to mock right now and might become more useful when we introduce proxies maybe?

jonasfj:
@gregarndt,
Yeah, as we start to use more features from queue... it'll be nice to have mock like this.
@imbstack already wants to use this in artifact plugin... livelog plugin will want to upload too..
I can imagine an index plugin that would want to use index mock, and a docker engine would want to use mock for index and queue too (for artifacts as images)... And a secret plugin could definitely want to use a mock for secrets...

jonasfj:
@petemoore,

Maybe we should move the interfaces into `taskcluster-client-go/queue/interface.go` and the mock implementations into `taskcluster-client-go/mock/queue.go`... for queue, etc...

That way you don't have to use the interface for a Queue object, but you can if you wish to. And if you do, then you can get a mock from the mock package...

I like the idea that mocks are easily available in the taskcluster-client-go, rather than a separate library. If you don't import them they don't cost anything.

gregarndt:
It's not been discovered that these interfaces are incorrect.  There are methods defined in these interfaces that are part of the tcclient/creds object and not part of Queue/Auth/Etc.  This breaks anywhere that we create a client and pass it to methods or assigned it to a field in a struct that expected the incorrect interface (such as client.Queue)

PR #63, opened 2016-03-25
Fix malformed global config

petemoore:
Fix [malformed global config](https://travis-ci.org/taskcluster/taskcluster-worker/jobs/118505373#L227).


petemoore:
CC @imbstack @walac

imbstack:
r+ looks good to me

PR #67, opened 2016-04-21
Bug 1265789 and Bug 1265792

walac:
Bug 1265789: Define the config and payload Mac OSX engine schemas.
The configs are a subset of the windows definitions.

Bug 1265792: Make OSX engine able to run a simple commmad.
It simply blindly run the given command. Sample run: https://tools.taskcluster.net/task-inspector/#PTi3sU8VSYWQwRR3GdB7QQ/

There is also a small commit to report the missing field in a schema parsing error.

r? @gregarndt @petemoore


engines/osxnative/config-schema.yml
```diff
@@ -0,0 +1,37 @@
+$schema:            http://json-schema.org/draft-04/schema#
+title:              Config
+description: |-
+  Config applicable to Mac OSX native engine
+type:               object
+properties:
+  workerGroup:
+    title:          Worker Group
```
gregarndt: All of these should be in a global worker config regardless of the engine that is running.

engines/osxnative/engine.go
```diff
@@ -0,0 +1,52 @@
+//go:generate go-composite-schema --unexported --required config config-schema.yml generated_configschema.go
+//go:generate go-composite-schema --unexported --required payload payload-schema.yml generated_payloadschema.go
+
+// Package macengine implements the Mac OSX engine
+package macengine
```
gregarndt: package name should match the directory name usually.  I would opt to call this "osxnative"

engines/osxnative/sandbox.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+}
+
+func (s *sandbox) WaitForResult() (engines.ResultSet, error) {
+	if s.aborted {
+		return resultset{success: false}, engines.ErrSandboxAborted
+	}
+
+	// Use the host environment plus custom environment variables
+	env := os.Environ()
```
gregarndt: I think that we should not copy any host environment variables if we can.  If we do need some of the host variables, we can white list those instead.  Secrets could be in the host variables.

engines/osxnative/sandbox.go
```diff
@@ -0,0 +1,68 @@
...
+	env = append(env, *s.env...)
+
+	for _, command := range s.taskPayload.Command {
+		cmd := exec.Command(command[0], command[1:]...)
+		cmd.Env = env
+
+		output, err := cmd.CombinedOutput()
```
gregarndt: So I haven't dealt with cmd much, but will this stream to the task log? We want the command streaming so live logging can be enabled rather than writing at the end.

runtime/compositeschema.go
```diff
@@ -107,7 +107,7 @@ func (s *schemaEntry) Parse(data map[string]json.RawMessage) (interface{}, error
 	value := data[s.property]
 	if value == nil {
 		if s.required {
-			return nil, errors.New("Property is missing")
+			return nil, errors.New("Property \"" + s.property + "\" is missing")
```
gregarndt: nice!

engines/osxnative/sandbox.go
walac: `output` is dump to the logs later in the code.

engines/osxnative/sandbox.go
walac: Hrm, there are some that are crucial, like `PATH`, `HOME`, etc. I will create the white list skeleton and we can add more variables as they find out we need them.

engines/osxnative/engine.go
walac: Ok, fine.

engines/osxnative/config-schema.yml
walac: So, we end up with an empty schema for now? cc @petemoore as this also applies for winnative engine.

engines/osxnative/sandbox.go
gregarndt: yea where we would want it to stream the command output to the task log so live logging can display it to the end user.

engines/osxnative/sandbox.go
gregarndt: yea what worried me is if we had a secret like SOME_API_SERVICE_TOKEN that the host is using to communicate somewhere and now that is leaked into the task.  The task environment really needs to be as isolated from anything from the host as possible.

engines/osxnative/config-schema.yml
gregarndt: Another reason why this is in the worker level config is because we will have a configuration  provider reading data from things like aws and such to determine these values, and these values will be used not only with the engine but elsewhere.

walac:
Addressed all comments, plus @jonasfj suggestion to use logging test

engines/osxnative/payload-schema.yml
```diff
@@ -0,0 +1,31 @@
+$schema:            http://json-schema.org/draft-04/schema#
+title:              "Payload"
+type:               "object"
+description: |-
+  Payload for Mac OSX Engine
+properties:
+  link:
```
gregarndt: So really I think that we should include any of this in the command part of the payload for now.  In the future we might want to look into doing something like this:

1. add task.payload.context to act similar to how .image is for docker.
    a. this context will be built from the decision task
    b. context lives in tree
    c. context will contain a single setup.sh script that will be invoked by the engine to setup the environment (hg clone, install packages, whatever)
    d. would look like task.payload.context = "https://queue.taskcluster.net/v1/1234/artifacts/public/contexts/osxnative/context.tar.gz"

engines/osxnative/payload-schema.yml
```diff
@@ -0,0 +1,31 @@
...
+      explicitly be referenced by the command field.
+  command:
+    title: "Command to run"
+    type: "array"
+    minItems: 1
+    items:
+      type: "string"
+    description: |-
+      The first item is the executable to run, followed by command line parameters
+  maxRunTime:
```
gregarndt: I do not think this is part of the engine payload, I think we will be writing a plugin that will define this part of the payload can abort the task when the run time is reached.

engines/osxnative/sandbox.go
```diff
@@ -0,0 +1,147 @@
...
+func newSandbox(context *runtime.TaskContext, taskPayload *payload, env *[]string) *sandbox {
+	return &sandbox{
+		context,
+		taskPayload,
+		env,
+		false,
+	}
+}
+
+func downloadLink(link string) (string, error) {
```
gregarndt: if we agree to get rid of task.payload.link in favor of doing something with preloaded volumes later, then we can get rid of this method entirely.

engines/osxnative/engine.go
```diff
@@ -0,0 +1,51 @@
...
+func (e engineProvider) ConfigSchema() runtime.CompositeSchema {
+	return runtime.NewEmptyCompositeSchema()
+}
+
+func (e engine) PayloadSchema() runtime.CompositeSchema {
+	return payloadSchema
+}
+
+func (e engine) NewSandboxBuilder(options engines.SandboxOptions) (engines.SandboxBuilder, error) {
+	taskPayload := options.Payload.(*payload)
```
gregarndt: if we do a type conversion here, and there is an error, this will cause the worker to panic and quit.

What we could possibly do is:
```
taskPayload, ok := options.Payload.(*payload)
if !ok {
  <do something here to log that there was an error and handle gracefully>
}
<continue on here if everything is ok>
```

Here is an example: https://play.golang.org/p/0EYrxyjhgC

engines/osxnative/engine.go
```diff
@@ -0,0 +1,51 @@
...
+	return sandboxbuilder{
+		engines.SandboxBuilderBase{},
```
gregarndt: I'm not sure how you feel, and this is totally a nit on my part, but how about naming each of these fields?  It makes it explicit what fields are being set to what values and doesn't rely on ordering to assign them.

engines/osxnative/engine.go
```diff
@@ -0,0 +1,51 @@
...
+		map[string]string{},
+		taskPayload,
+		options.TaskContext,
+	}, nil
+}
+
+func (engine) NewCacheFolder() (engines.Volume, error) {
```
gregarndt: I think you get this by default when embedding the enginebase

engines/osxnative/sandboxbuilder.go
```diff
@@ -0,0 +1,33 @@
...
+
+type sandboxbuilder struct {
+	engines.SandboxBuilderBase
+	env         map[string]string
+	taskPayload *payload
+	context     *runtime.TaskContext
+}
+
+func (s sandboxbuilder) SetEnvironmentVariable(name string, value string) error {
+	_, ok := s.env[name]
```
gregarndt: I assume we're doing this because if it's already set, then there is an issue?

nit: could "ok" be named something better, like "exists" which is what we're checking here?

engines/osxnative/sandboxbuilder.go
```diff
@@ -0,0 +1,33 @@
...
+	if ok {
+		return engines.ErrNamingConflict
+	}
+
+	s.context.Log("Setting \"", name, "\" environment variable to \"", value, "\" ")
```
gregarndt: Do we want these in the task logs for every environment variable we're setting?

engines/osxnative/sandboxbuilder.go
```diff
@@ -0,0 +1,33 @@
...
+	s.env[name] = value
```
gregarndt: Because it might be possible for other plugins to be setting environment variables, we probably should add a lock around this. (and the reading of it on line 16)

engines/osxnative/sandboxbuilder.go
```diff
@@ -0,0 +1,33 @@
...
+	return nil
+}
+
+func (s sandboxbuilder) StartSandbox() (engines.Sandbox, error) {
+	env := []string{}
+	for name, value := range s.env {
+		env = append(env, name+"="+value)
+	}
+
+	return newSandbox(s.context, s.taskPayload, &env), nil
```
gregarndt: Do you see a time where we might have other configuration options fo rthe sandbox that we want to pass in? Perhaps there could be a sandbox configuration object we pass in, and env is part of that.  Then newSandbox knows what to do with that config that we pass in.

engines/osxnative/sandbox.go
```diff
@@ -0,0 +1,147 @@
...
+
+type sandbox struct {
+	context     *runtime.TaskContext
+	taskPayload *payload
+	env         *[]string
+	aborted     bool
+}
+
+func newSandbox(context *runtime.TaskContext, taskPayload *payload, env *[]string) *sandbox {
+	return &sandbox{
```
gregarndt: nit: be explicit about the field names when there's more than one field.

engines/osxnative/sandbox.go
```diff
@@ -0,0 +1,147 @@
...
+
+	return resultset{success: true}, nil
+}
+
+func (s *sandbox) Abort() error {
+	s.aborted = true
+	return nil
+}
+
+func (*sandbox) NewShell() (engines.Shell, error) {
```
gregarndt: I think if you embed the engines.SandboxBase you get these not supported method for free

engines/osxnative/sandboxbuilder.go
```diff
@@ -0,0 +1,33 @@
...
+	if ok {
+		return engines.ErrNamingConflict
+	}
+
+	s.context.Log("Setting \"", name, "\" environment variable to \"", value, "\" ")
+	s.env[name] = value
+	return nil
+}
+
+func (s sandboxbuilder) StartSandbox() (engines.Sandbox, error) {
```
gregarndt: It seems that based on this method name of "Start...." that the sandbox (process) should actually be started at this point.

Calling StartSandbox to return a new (unstarted) sandbox seems to go against how this method is named.

engines/osxnative/sandbox.go
```diff
@@ -0,0 +1,147 @@
...
+
+		err = os.Chmod(filename, 0777)
+
+		if err != nil {
+			s.context.LogError(err)
+			return nil, engines.ErrNonFatalInternalError
+		}
+	}
+
+	env := make([]string, len(*s.env), len(*s.env)+len(environmentWhitelist))
```
gregarndt: One of the things that we do currently in docker-worker that is being used in production is setting the folloowing environment variables https://github.com/taskcluster/docker-worker/blob/master/lib/task.js#L324-L329

I think these variables should be added to the sandbox's environment variables and cannot be overwritten by a plugin.

engines/osxnative/sandbox.go
```diff
@@ -0,0 +1,147 @@
...
+	"LANG",
+	"LOGNAME",
+	"TERM",
+	"TERM_PROGRAM",
+}
+
+type sandbox struct {
+	context     *runtime.TaskContext
+	taskPayload *payload
+	env         *[]string
```
gregarndt: This does not need to be a pointer.  slices and maps are already references

engines/osxnative/sandbox.go
```diff
@@ -0,0 +1,147 @@
...
+			return nil, engines.ErrNonFatalInternalError
+		}
+	}
+
+	env := make([]string, len(*s.env), len(*s.env)+len(environmentWhitelist))
+	copy(env, *s.env)
+
+	// Use the host environment plus custom environment variables
+	for _, e := range environmentWhitelist {
+		value := os.Getenv(e)
```
gregarndt: Was LookupEnv considered here?  This way there is a boolean to check to see if the environment variable existed at all, and if it exists, we can just add it to the sandbox's environment variables even if it's empty.

engines/osxnative/sandbox.go
```diff
@@ -0,0 +1,147 @@
...
+		if value != "" {
+			env = append(env, e+"="+value)
+		}
+	}
+
+	cmd := exec.Command(s.taskPayload.Command[0], s.taskPayload.Command[1:]...)
+	cmd.Env = env
+
+	output, err := cmd.CombinedOutput()
```
gregarndt: We should do something where stdout and stderr are streams to taskContext.Log() so that it's written to the task log as thre is data streaming in and the live log can later display it.  This appears to dump it once the command is done running.

engines/osxnative/sandbox.go
walac: Well, I think meanwhile we need to land it. When we have preloaded volumes, we can get rid of it.

engines/osxnative/engine.go
walac: I have mixed feelings about this. If we have a type conversion error here, something is terribly wrong, I am not sure the worker should live in a world where expected type conversions fail.

engines/osxnative/engine.go
walac: That's fine. #lazyboy

engines/osxnative/payload-schema.yml
walac: Removed.

engines/osxnative/engine.go
walac: Fixed.

engines/osxnative/engine.go
gregarndt: Yea, I have a little mixed feelings about it too and no strong opinion either way.

The one decision we need to make is if this should be a "panic" event, which it will be if a type conversion fails or if it's something that we should fail on but continue to keep the worker running.

engines/osxnative/sandbox.go
gregarndt: Couldn't we just use "curl" as one of the commands to get the script if we agree that at some point this is probably going away anyways?

engines/osxnative/sandboxbuilder.go
walac: Fixed.

engines/osxnative/sandboxbuilder.go
walac: We need this for automated tests, and usually there are not more than a half dozen variables.

gregarndt: Hrm, could you point me to where they're needed for automated tests? I wonder if there is a way we could get around that being needed

engines/osxnative/sandboxbuilder.go
walac: Done.

engines/osxnative/sandboxbuilder.go
walac: I am not sure, but I am in favor delay that change for when the time comes.

engines/osxnative/sandbox.go
walac: Done.

engines/osxnative/sandbox.go
walac: `SanboxBase` has a object receiver, and I need a pointer receiver to modify `aborted`.

engines/osxnative/sandboxbuilder.go
gregarndt: Ok :)

engines/osxnative/sandboxbuilder.go
walac: ¯\_(ツ)_/¯

engines/osxnative/sandbox.go
walac: Done.

engines/osxnative/sandbox.go
walac: Good catch.

engines/osxnative/sandbox.go
gregarndt: hrm, I might not be understanding embedding of fields then...

engines/osxnative/sandbox.go
walac: Fixed.

engines/osxnative/sandbox.go
walac: Done.

engines/osxnative/sandbox.go
walac: It took a while for me to understand, but if I use `SandboxBase`, some methods will be implemented with an object receiver, others with a pointer receiver, so neither will completely implement the `Sandbox`interface. Kinda weird.

engines/osxnative/engine.go
walac: Looking at the code and giving a little more thought, I included your suggestion.

engines/osxnative/sandbox.go
walac: Do you mean implement `downloadLink` as a call to `curl`?

engines/osxnative/sandboxbuilder.go
walac: [Here](https://github.com/walac/taskcluster-worker/blob/macosx/engines/osxnative/engine_test.go#L13-L27), it looks at logs to see if we printed the environment variables.

engines/osxnative/sandbox.go
gregarndt: I was thinking that curling the script down would just be one of the first "command" that is run instead of any special handling within the engine.

engines/osxnative/sandbox.go
petemoore: On the Windows workers we make a system call which provides the user env variables for the temp user we created (so all the required variables exist, but are for the temp user, not the TC user).

gregarndt: I believe right now this engine is running under the same user that the worker is running under.  I believe a later iteration is investigating using temp users to run tasks.

PR #69, opened 2016-04-21
Remove vet from travis setup.

walac:
For some reason vet installation is failing on travis, but it is not
required anyway because go 1.5 already ships it.


gregarndt:
Verified in the travis logs that `go vet ./...` was still run successfully.  Looks good to me.

PR #70, opened 2016-04-21
Dump task logs to log entry.

walac:
Task context writes log messages to a temporary file, this makes
difficulty to look at the logs when running the tasks locally.

TaskContext now write logs to logrus debug output besides temporary
files.

### DIFF
--- a/runtime/taskcontext.go
+++ b/runtime/taskcontext.go
@@ -7,6 +7,7 @@ import (
 
 	"sync"
 
+	"github.com/Sirupsen/logrus"
 	"github.com/taskcluster/taskcluster-client-go/tcclient"
 	"github.com/taskcluster/taskcluster-worker/runtime/client"
 	"github.com/taskcluster/taskcluster-worker/runtime/webhookserver"
@@ -63,6 +64,7 @@ type TaskInfo struct {
 // properties, and abortion notifications.
 type TaskContext struct {
 	TaskInfo
+	LogEntry   *logrus.Entry
 	webHookSet *webhookserver.WebHookSet
 	logStream  *stream.Stream
 	mu         sync.RWMutex
@@ -173,6 +175,10 @@ func (c *TaskContext) log(prefix string, a ...interface{}) {
 	_, err := fmt.Fprintln(c.logStream, a...)
 	if err != nil {
 		//TODO: Forward this to the system log, it's not a critical error
+	}
+
+	if c.LogEntry != nil {
+		c.LogEntry.Debug(a...)
 	}
 }
 
--- a/worker/task.go
+++ b/worker/task.go
@@ -55,6 +55,7 @@ func NewTaskRun(
 		Expires:  claim.taskClaim.Task.Expires,
 	}
 	ctxt, ctxtctl, err := runtime.NewTaskContext(tp, info)
+	ctxt.LogEntry = log
 
 	queueClient := queue.New(&tcclient.Credentials{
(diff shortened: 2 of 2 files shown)


runtime/taskcontext.go
```diff
@@ -63,6 +64,7 @@ type TaskInfo struct {
 // properties, and abortion notifications.
 type TaskContext struct {
 	TaskInfo
+	LogEntry   *logrus.Entry
```
jonasfj: Let's call this `systemLog` or `workerLog`

worker/task.go
```diff
@@ -55,6 +55,7 @@ func NewTaskRun(
 		Expires:  claim.taskClaim.Task.Expires,
 	}
 	ctxt, ctxtctl, err := runtime.NewTaskContext(tp, info)
+	ctxt.LogEntry = log
```
jonasfj: This should probably be a sub-logger that has a property with the `taskId:<taskId>`

jonasfj: Oh, maybe I'm wrong here... It probably already has that property...

Maybe a property: `kind:task-log`

jonasfj:
I'm not sure this is a good idea...
Worker logs and task logs are two very different things.
task logs is a product similar to build artifacts and test result. worker logs is a log of the process..

When running locally, and creating a task on queue.taskcluster.net, you should be able to open the taskId on the task-inspector and view the log there...
Granted maybe we aren't uploading yet :)

I guess that this being added as debug log it's not all bad...

runtime/taskcontext.go
```diff
@@ -174,6 +176,10 @@ func (c *TaskContext) log(prefix string, a ...interface{}) {
 	if err != nil {
 		//TODO: Forward this to the system log, it's not a critical error
 	}
+
+	if c.LogEntry != nil {
```
jonasfj: What are the use-cases for this?

If running a process you normally won't write logs using this method, but using the `LogDrain()`...
Ie if you have a subprocess running you set stdout and stderr to `TaskContext.LogDrain()`

worker/task.go
jonasfj: Okay maybe a better label is something like task info or error...

gregarndt:
I get why this might be necessary, especially when we currently do not have logging support.  I think though that we could wire up uploading the task log (maybe not live logging just yet) easily enough (I did it when we were in portland, I would just need to make it not so hacky).  Perhaps that will get us what you're looking for?

I think that intermixing the worker logs with task logs is going to become extremely cumbersome.  Perhaps instead of mixing those logs, and if uploading the task log won't accomplish what you want, we could output in the worker logs what the temporary log file name is for a particular task ID, and then if the worker is running in a "debug" mode, the temporary file is not removed once complete.  I'm not sure how much I like that.. just a crazy thing I just thought of.

jonasfj:
note: Normally when debugging an engine I suggest using the `enginetest` package.. It'll record the log for you... you can have tests that just runs a given payload and greps the log...

So maybe this is less of an issue if we don't debug by running a worker locally.
@walac here is an example of how to test an engine:
https://github.com/taskcluster/taskcluster-worker/blob/51ceb04dc0960f045f7da2ee3b598be07b66835d/engines/mock/mockengine_test.go#L39-L68

If your mac engine implements the `Engine` interface as intended, it should be easy to write a bunch of test cases like this :)

walac:
This is mostly for convenience, while playing with taskcluster-worker. Maybe a command line option to dump the task logs to screen?

PR #71, opened 2016-05-15
Commands

jonasfj:
@petemoore, @walac 
I'm not sure if this breaks anything that is currently working... It could be... I'll admit I didn't test the main code as I'm not exactly sure how do that...

But with a refactor like this we can have more than one top-level command.
I think we have can have command like "run-task-locally", or "build-virtual-machine", etc...
These could be quiet useful when debugging...

The command to run the worker should be:
`taskcluster-worker work <engine> --logging-level <level>`
Everything else should be as before, but please let me know if I broke something.


walac:
Looking at the code, I can't any breakage on my side.

commands/work/work.go
```diff
@@ -0,0 +1,120 @@
...
+
+type cmd struct{}
+
+func (cmd) Summary() string {
+	return "Start the worker."
+}
+
+func (cmd) Usage() string {
+	return `Usage:
+  taskcluster-worker work <engine> --logging-level <level>
```
gregarndt: so here we want it to look like `taskcluster-worker work qemu` ? or should we be explicit witht he engine we're passing in and use `taskcluster-worker work --engine qemo`

commands/run.go
```diff
@@ -46,7 +46,7 @@ func Run(argv []string) {
 
 	// Parse args for command provider
 	subArguments, _ := docopt.Parse(
-		provider.Usage(), arguments["<args>"].([]string),
+		provider.Usage(), append([]string{cmd}, arguments["<args>"].([]string)...),
```
gregarndt: Has taskcluster-worker been compiled and tested to work with this new sub command? I attempted to use it but it keeps printing out the help infromation and exiting with a status code of 1.  If I set the exit argument to false, then it seems to pass this line.

commands/work/work.go
jonasfj: Hmm... My thinking was that it's a required argument... and not an option.
But I see no basis for that in:
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html
or
http://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

commands/run.go
jonasfj: I can get help working... I have testing the `work` command...
Maybe I need to figure out how it works...

commands/work/work.go
gregarndt: Ok, just wondering.  I like the `taskcluster-worker work` piece, but just didn't know why engine was a like this when it's just another argument to "work".  I suppose it matters when we define what we want to be passing in from the command line to run the worker.  Do we see us also passing a list of plugins or load, or some other options? If so, specifying the engine with "--engine" might make sense.

jonasfj: I suspect having everything as config might be preferable... but we'll see...

jonasfj:
Got it working with `assume:project:taskcluster:worker-test-scopes`

gregarndt:
looks good to me now.

PR #72, opened 2016-05-18
Tasks should be reclaimed on an interval

gregarndt:


worker/task.go
```diff
@@ -79,10 +82,41 @@ func NewTaskRun(
...
+	duration := until.Sub(time.Now()).Seconds()
+	// Using a reclaim divisor of 1.3 with the default reclaim deadline (20 minutes),
+	// means that a reclaim event will happen with a few minutes left of the origin claim.
+	nextReclaim := duration / 1.3
+	select {
+	case <-time.After(time.Duration(nextReclaim * 1e+9)):
+		client := t.controller.Queue()
+		claim, err := reclaimTask(client, t.TaskID, t.RunID, t.log)
+		if err != nil {
+			t.log.WithField("error", err.Error()).Warn("Error reclaiming task")
```
walac: Shouldn't this log as an error?

worker/task_test.go
```diff
@@ -53,6 +57,7 @@ var claim = &taskClaim{
 			ClientID:    "abc",
 			Certificate: "",
 		},
+		TakenUntil: tcclient.Time(time.Now().Add(time.Minute * 5)),
```
walac: nit: document `TakenUntil` scale.

walac:
r+

worker/task_test.go
gregarndt: Document why 5 minutes was picked?

worker/task.go
gregarndt: You're very right sir, I'll change that.  And with you pointing that out and me checking out the logrus documentation, I have discovered:
https://github.com/Sirupsen/logrus/blob/master/entry.go#L60

which I pretty much do in a lot of places so that can be simplified.

Thanks!

worker/task_test.go
walac: What's the unit of TakenUntil? Milisecs?

gregarndt: TakenUntil is just a time stamp https://docs.taskcluster.net/reference/platform/queue/api-docs#reclaimTask

gregarndt: And if you jump through it using something like vimgo, it'll take you to the tcclient package in taskcluster-client-go that shows it's a tcclient.Time, which is really time.Time with a few methods added for serializing/deserializing

PR #74, opened 2016-05-26
Call Dispose methods. r=garndt

walac:
Call TaskPlugin and ResultSet Dispose methods after task is finished.


gregarndt:
If travis is happy then I'm happy.

walac:
Travis is not happy, but seems unrelated to the patch. @petemoore ?

gregarndt:
I don't think it was related to this patch or what pmoore changed.  I think it was just travis being wonky.  I retriggered the builds and it looks fine.

walac:
Poor travis. Anyway, merged.

PR #77, opened 2016-06-01
Run tasks with their own user context.

walac:
To run tasks concurrently in the same environment and make sure they don't
mess up with each other, we run each of them with a different user, which
also prevents tasks from doing fancy things to get secret information.

For each new task, we create a new user on the fly, associate it to the
staff group member and create a home directory. The task is triggered
from user's home directory. After task is done, we remove the user as
well as its home directory.

In the case we fail to create a new user, we log the error but don't
fail the task. This behavior is to let the tests pass even when
running without capability to create users.

### DIFF
--- a/engines/osxnative/dscl.go
+++ b/engines/osxnative/dscl.go
@@ -1,0 +1,64 @@
+package osxnative
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+)
+
+type dscl struct {
+}
+
+// run will exectue a dscl command using the provided arguments
+// returning the output of the command as a string.
+func (d dscl) run(command string, a ...string) (string, error) {
+	args := []string{".", command}
+	args = append(args, a...)
+	cmd := exec.Command("dscl", args...)
+	cmd.Env = os.Environ()
+	output, err := cmd.Output()
+	return string(output), err
+}
+
+func (d dscl) list(a ...string) ([][]string, error) {
+	output, err := d.run("list", a...)
+
+	if err != nil {
+		return nil, err
+	}
+
+	ret := [][]string{}
+	for _, s := range strings.Split(output, "\n") {
+		fields := strings.Fields(s)
+		if len(fields) != 0 {
+			ret = append(ret, strings.Fields(s))
+		}
+	}
+
+	return ret, nil
+}
+
+func (d dscl) read(a ...string) (string, error) {
+	s, err := d.run("read", a...)
+
+	if err != nil {
+		return s, err
+	}
+
+	return strings.Fields(s)[1], nil
+}
+
+func (d dscl) create(a ...string) error {
+	_, err := d.run("create", a...)
+	return err
+}
+
+func (d dscl) append(a ...string) error {
+	_, err := d.run("append", a...)
+	return err
+}
+
+func (d dscl) delete(a ...string) error {
+	_, err := d.run("delete", a...)
+	return err
+}
--- a/engines/osxnative/dscl.go
+++ b/engines/osxnative/dscl.go
@@ -1,3 +1,5 @@
+// +build darwin
+
 package osxnative
 
 import (
--- a/engines/osxnative/resultset.go
+++ b/engines/osxnative/resultset.go
@@ -5,40 +5,30 @@ import (
 	"github.com/taskcluster/taskcluster-worker/runtime"
 	"github.com/taskcluster/taskcluster-worker/runtime/ioext"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
-	"strings"
 )
 
 type resultset struct {
 	engines.ResultSetBase
-	context *runtime.TaskContext
-	success bool
+	taskUser user
+	context  *runtime.TaskContext
+	success  bool
+	engine   *engine
 }
 
 var pathMatcher *regexp.Regexp
 
-func init() {
-	home := os.Getenv("HOME")
-	if strings.HasPrefix(home, "/") {
-		// remove trailing "/"
-		home = home[:len(home)]
-	}
-
-	pathMatcher = regexp.MustCompile("^(" + home + "|/tmp)(/.*)?$")
-}
-
-func newResultSet(context *runtime.TaskContext, success bool) resultset {
-	return resultset{
-		ResultSetBase: engines.ResultSetBase{},
-		context:       context,
-		success:       success,
-	}
-}
-
 // Check if the path is valid. The only valid paths are those
 // inside user's home directory and in /tmp/
-func (r resultset) validPath(pathName string) bool {
+func (r resultset) validPath(home string, pathName string) bool {
+	pathMatcher = regexp.MustCompile("^(" + home + "|/tmp)(/.*)?$")
+
+	if !filepath.IsAbs(pathName) {
+		pathName = filepath.Join(home, pathName)
+	}
+
 	absPath, err := filepath.Abs(pathName)
 
 	if err != nil {
@@ -56,8 +46,13 @@ func (r resultset) Success() bool {
 }
 
 func (r resultset) ExtractFile(path string) (ioext.ReadSeekCloser, error) {
-	if !r.validPath(path) {
+	cwd := getWorkingDir(r.taskUser, r.context)
+	if !r.validPath(cwd, path) {
 		return nil, engines.NewMalformedPayloadError(path + " is invalid")
+	}
+
+	if !filepath.IsAbs(path) {
+		path = filepath.Join(cwd, path)
 	}
 
 	file, err := os.Open(path)
@@ -71,8 +66,13 @@ func (r resultset) ExtractFile(path string) (ioext.ReadSeekCloser, error) {
 }
 
 func (r resultset) ExtractFolder(path string, handler engines.FileHandler) error {
-	if !r.validPath(path) {
+	cwd := getWorkingDir(r.taskUser, r.context)
+	if !r.validPath(cwd, path) {
 		return engines.NewMalformedPayloadError(path + " is invalid")
+	}
+
+	if !filepath.IsAbs(path) {
+		path = filepath.Join(cwd, path)
 	}
 
 	return filepath.Walk(path, func(p string, info os.FileInfo, e error) error {
@@ -96,4 +96,19 @@ func (r resultset) ExtractFolder(path string, handler engines.FileHandler) error
 
 		return nil
 	})
+}
+
+func (r resultset) Dispose() error {
+	err := r.taskUser.delete()
+	if err != nil {
+		r.engine.log.WithField("user", r.taskUser.name).WithError(err).Error("Error removing user")
+		exitError, ok := err.(*exec.ExitError)
+		if ok {
+			r.engine.log.Error(string(exitError.Stderr))
+		}
+
+		return engines.ErrNonFatalInternalError
+	}
+
+	return nil
 }
--- a/engines/osxnative/resultset.go
+++ b/engines/osxnative/resultset.go
@@ -1,3 +1,5 @@
+// +build darwin
+
 package osxnative
 
 import (
--- a/engines/osxnative/sandbox.go
+++ b/engines/osxnative/sandbox.go
@@ -8,6 +8,11 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	osuser "os/user"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"syscall"
 )
 
 // white list of environment variables that must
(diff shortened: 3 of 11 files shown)


engines/osxnative/sandbox.go
```diff
@@ -136,21 +125,99 @@ func (s *sandbox) WaitForResult() (engines.ResultSet, error) {
...
+	// depend on either we create the new user successfuly or not
+	processUser := os.Getenv("USER")
+	processHome := os.Getenv("HOME")
+
+	// If we fail to create a new user, the most probable cause is that
+	// we don't have enough permissions. Chances are that we are running
+	// in in a development environment, so do not fail the task to tests
+	// run successfully.
+	u := user{}
+	if err = u.create(); err != nil {
```
gregarndt: So I think that this can be rather dangerous.  If for some reason creating a user fails, this could be run under escalated permissions without knowing it.

I think it should be an either/or thing.  Either the engine is configured to run a process under a temp user and fails if that user can't be used, or the engine is configured to not create a temp user and run under the worker user.

engines/osxnative/user.go
```diff
@@ -0,0 +1,149 @@
...
+		u.delete()
+		return err
+	}
+
+	gid, err := strconv.Atoi(staffGid)
+	if err != nil {
+		return err
+	}
+
+	name := "worker-" + uuid.NewV4().String()
```
gregarndt: for the sake of consistency, we might want to use the uuid library that's used in our slugid package.

engines/osxnative/user.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (u *user) create() error {
+	uid, err := u.getMaxUid()
+	if err != nil {
+		return err
+	}
+
+	staffGid, err := u.d.read("/Groups/staff", "PrimaryGroupID")
+	if err != nil {
+		u.delete()
```
gregarndt: There's no user to delete at this point.

engines/osxnative/user.go
```diff
@@ -0,0 +1,149 @@
...
+		return err
+	}
+
+	gid, err := strconv.Atoi(staffGid)
```
gregarndt: What makes this group significant? Is this a group only on the releng loaner or is it something that's by default on OS X?

engines/osxnative/user.go
```diff
@@ -0,0 +1,149 @@
...
+	name string
+}
+
+func (u *user) create() error {
+	uid, err := u.getMaxUid()
+	if err != nil {
+		return err
+	}
+
+	staffGid, err := u.d.read("/Groups/staff", "PrimaryGroupID")
```
gregarndt: What makes this group significant? Is this a group only on the releng loaner or is it something that's by default on OS X?

gregarndt: So I think this won't return an error if the group doesn't exist, what would we do in that case?

engines/osxnative/user.go
```diff
@@ -0,0 +1,149 @@
...
+func (u *user) getMaxUid() (int, error) {
+	uids, err := u.d.list("/Users", "uid")
+
+	if err != nil {
+		return -1, err
+	}
+
+	var maxUid int = 0
+	for _, entry := range uids {
+		uid, err := strconv.Atoi(entry[1])
```
gregarndt: Have you seen converting entry[1] result in an error? Wouldn't everything have a uid?

engines/osxnative/user.go
```diff
@@ -0,0 +1,149 @@
...
+
+// return the next uid available
+func (u *user) getMaxUid() (int, error) {
+	uids, err := u.d.list("/Users", "uid")
+
+	if err != nil {
+		return -1, err
+	}
+
+	var maxUid int = 0
```
gregarndt: you could do:

maxUid := 0
or even:
var maxUid int

default value for an int is 0, but that's a little magical.

engines/osxnative/user.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return err
+	}
+
+	// We set the uid, then check if it is unique, if
+	// not, increment and try again
+	duplicated := true
+	for duplicated {
+		uid++
+		strUid := strconv.FormatInt(int64(uid), 10)
```
gregarndt: You could save a step by doing:
strconf.Itoa(uid) which really is just calling FormatInt(int64(<int>), 10) under the hood.

engines/osxnative/user.go
```diff
@@ -0,0 +1,149 @@
...
+
+		err = u.d.create(userPath, "uid", strUid)
+		if err != nil {
+			u.delete()
+			return err
+		}
+
+		duplicated, err = u.isDuplicatedId("uid", uid)
```
gregarndt: I have to admit I'm not quite sure what isDuplicatedId does.  Can you create a user with the same ID as another user?

Also some comments around some of these methods as to the intention of them would help.  

I tried running the 'dscl . list /Users <name>' command for a known user on my system and it returned nothing.  Not sure what I should be expecting.

engines/osxnative/user.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	// We set the uid, then check if it is unique, if
+	// not, increment and try again
+	duplicated := true
+	for duplicated {
+		uid++
+		strUid := strconv.FormatInt(int64(uid), 10)
+
+		err = u.d.create(userPath, "uid", strUid)
```
gregarndt: I wonder how racey this would get if a worker was running like 10 parallel tasks.

engines/osxnative/user.go
```diff
@@ -0,0 +1,149 @@
...
+		return err
+	}
+
+	err = os.MkdirAll(userPath, 0700)
+	if err != nil {
+		u.delete()
+		return err
+	}
+
+	err = os.Chown(userPath, uid, gid)
```
gregarndt: Wouldn't this make the path owned by the gid, and so other tasks sharing the same gid could access the userPath of other tasks?

engines/osxnative/user_test.go
```diff
@@ -0,0 +1,58 @@
...
+		t.Fatal(err)
+	}
+
+	if currentUser.Username != "root" {
+		t.Skip("This test requires root privileges, skipping.")
+	}
+
+	u := user{}
+	err = u.create()
+	if err != nil {
```
gregarndt: one thing that's kind of nice is the assert library I've used elsewhere that you could do:
assert.Nil(t, err, "Couldn't create user")

engines/osxnative/user_test.go
```diff
@@ -0,0 +1,58 @@
+package osxnative
+
+import (
+	"os"
+	osuser "os/user"
+	"testing"
+)
+
+func TestUser(t *testing.T) {
```
gregarndt: Tests should probably be added for the engine that tests running two sandboxes at the same time attempting to access eachothers environment.

engines/osxnative/dscl.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"os"
+	"os/exec"
+	"strings"
+)
+
+type dscl struct {
+}
+
+// Run dscl
```
gregarndt: I think it's usually common to document a method like this:
// run will exectue a dscl command using the provided arguments returning the output of
//the command as a string.

engines/osxnative/resultset.go
```diff
@@ -97,3 +96,18 @@ func (r resultset) ExtractFolder(path string, handler engines.FileHandler) error
...
 	})
 }
+
+func (r resultset) Dispose() error {
+	err := r.taskUser.delete()
+	if err != nil {
+		r.context.LogError("Error removing user \""+r.taskUser.name+"\": ", err)
+		exitError, ok := err.(*exec.ExitError)
+		if ok {
+			r.context.LogError(string(exitError.Stderr))
```
gregarndt: the task context log is closed before the Dispose() method is called on a resultset.

engines/osxnative/resultset.go
```diff
@@ -97,3 +96,18 @@ func (r resultset) ExtractFolder(path string, handler engines.FileHandler) error
 		return nil
 	})
 }
+
+func (r resultset) Dispose() error {
+	err := r.taskUser.delete()
+	if err != nil {
```
gregarndt: We should have tests for testing these error scenarios as well.

engines/osxnative/sandbox.go
gregarndt: Thinking about it a little more, i wonder if it would be possible to have the engine configuration option to either use a module for creating the user, or using the user that the engine is running under.  If that config setting says to create a user using some other module, then this .create() stuff can be called and the user returned from that used for the process. Also would allow someone to specify different creation scripts depending on need.  Like if there are some special groups that the user should be part of on one type of worker type but not another.

What do you think?

engines/osxnative/resultset.go
walac: I am skipping user creation tests because they are quite difficult to implement. Going to address this in a follow up patch.

engines/osxnative/dscl.go
walac: Done.

engines/osxnative/resultset.go
walac: Done.

engines/osxnative/sandbox.go
walac: As we talked earlier, I implemented a environment variable based configuration. If `TASKCLUSTER_WORKER_ENV=Production` (non case sensitive), the task fails if it can't create the user.

engines/osxnative/user.go
walac: It is the same user as `cltbld`. `staff` seems to be the default group for normal users.

engines/osxnative/user.go
walac: np, `delete` function is smart enough, there is no harm on calling it.

engines/osxnative/user.go
walac: Done.

engines/osxnative/user.go
walac: That's why there's a loop here :)

engines/osxnative/user.go
gregarndt: so really this user could still have the same permissions as the cltbld user, but just running under a different user?  This is a step in the right direction but I think we need to spend time understand what this user in that group is capable of.  We need to makes sure that it can't see things it's not supposed to, change ACLs, or anything else destructive.

engines/osxnative/user.go
gregarndt: Yea, but will they all eventually win in a short amount of time?

engines/osxnative/user.go
walac: The loop is here because we might have more than one task running at the same time, and they can eventually create duplicated uid's. `isDuplicatedId` check's if there is a duplicated id, and if so, we try another uid.

gregarndt: yea, but isn't it checking after it already created the user on line 47? If so, how did os allow two users to be created using the same UID.  I would have expected the os to block this and only on create() would have succeeded.

engines/osxnative/user.go
walac: Do you think it is better to let the `admin` group owning the path?

engines/osxnative/user.go
walac: Yeah, `golint` just told me that.

engines/osxnative/user.go
walac: No, but if I don't check for errors, I'll never do :p
My default policy is *always* check return values.

gregarndt: Oh, I'm totally cool with checking for errors, I was more so just curious for my own knowledge.

engines/osxnative/user.go
gregarndt: hrm, would that allow the uid to have ownership and anyone within the admin group? If so, that might work.

engines/osxnative/user_test.go
walac: Oh, I looked for such a library a long time ago. Refactored.

walac:
@gregarndt may I merge this?

gregarndt:
do eeet

engines/osxnative/sandbox.go
petemoore: I'd prefer that we used a configuration file exclusively, rather than a combination of env vars and a config file. This also has the advantage that we will have json schema for the config file, so it can be validated reasonably easily, and we'll detect problems on startup, rather than at task execution time.

walac:
I rebased the branch, but for some reason travis doesn't understand that, going to merge it as fails are unrelated to this patch.

PR #79, opened 2016-06-20
Add plugin for managing volumes

gregarndt:
Currently this just supports persistent volumes (aka caches).

There were some things to work around here which leads some oddities in the payload...

rather than having keys in the payload for each type of volume, it's much cleaner to encapsulate them into one volume object in task.payload.  However, each volume cannot be implemented as a plugin because only one top level key "volume" can be defined and plugin can't define a schema for something under the "volume" key.

this leads to:
```
payload: {
  volumes: {
    persistent: []
  }
}
```

Also, the test for reusing a persistent volume is long because it needs to go through all the stages of the task lifecyle twice to ensure a volume is reused.  The plugintest test case cannot be used in this case because it recreates the environment, engine, and task plugins between task runs.

Also, there could be tests for a few of the missing cases where an error is returned form the engine, but because plugintest does nilOrPanic when there is an error, there is no way to validate it while also using the base test case.  Rather than adding a lot more code to check this, it's assuming the engine test cases cover these scenarios.


plugins/volume/persistent.go
```diff
@@ -0,0 +1,145 @@
...
+func (pv *persistentVolumePlugin) newVolume(opts *volumeOptions) (volume, error) {
+	if v := pv.claimAvailableVolume(opts); v != nil {
+		return v, nil
+	}
+
+	// Create a volume since an available one doesn't exist
+	return pv.createVolume(opts)
+}
+
+// claimAvailableVolume will claim and return a volume is one is found that is not
```
walac: "...will claim and return a volume that is found and not..."

plugins/volume/persistent.go
```diff
@@ -0,0 +1,145 @@
...
+// in use currently by a task.
+func (pv *persistentVolumePlugin) claimAvailableVolume(opts *volumeOptions) *persistentVolume {
+	pv.mutex.Lock()
+	defer pv.mutex.Unlock()
+	volumes, _ := pv.volumes[opts.name]
+	for _, v := range volumes {
```
walac: I assume we won't have more than some dozens of volumes, so a linear search won't impact performance.

plugins/volume/persistent.go
```diff
@@ -0,0 +1,145 @@
...
+func (pv *persistentVolumePlugin) createVolume(opts *volumeOptions) (*persistentVolume, error) {
+	v := &persistentVolume{
+		id:   uuid.New(),
+		name: opts.name,
+	}
+
+	var err error
+	v.volume, err = pv.engine.NewCacheFolder()
+	if err != nil {
+		e := fmt.Errorf("Could not create persistent volume. %s", err)
```
walac: Isn't `%v` here? I guess `error` implements `String()`, but I often use `%v`.

plugins/volume/persistent.go
```diff
@@ -0,0 +1,145 @@
...
+// createVolume will request a new cache folder from the engine, claim it for the current
+// task run, and add it to a list of known persistent volumes.
+func (pv *persistentVolumePlugin) createVolume(opts *volumeOptions) (*persistentVolume, error) {
+	v := &persistentVolume{
+		id:   uuid.New(),
+		name: opts.name,
+	}
+
+	var err error
+	v.volume, err = pv.engine.NewCacheFolder()
```
walac:
```go
v.volume, err := pv.engine.NewCacheFolder()
```

works if you have go >= 1.5, iirc.

plugins/volume/persistent_test.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/taskcluster/taskcluster-worker/engines"
+	"github.com/taskcluster/taskcluster-worker/engines/extpoints"
+	_ "github.com/taskcluster/taskcluster-worker/engines/mock"
+	"github.com/taskcluster/taskcluster-worker/plugins"
+	pluginExtpoints "github.com/taskcluster/taskcluster-worker/plugins/extpoints"
+	"github.com/taskcluster/taskcluster-worker/plugins/plugintest"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+)
+
+var logger, _ = runtime.CreateLogger(os.Getenv("LOGGING_LEVEL"))
```
walac:
```go
logger, _ := runtime.CreateLogger(os.Getenv("LOGGING_LEVEL"))
```

plugins/volume/volumeset.go
```diff
@@ -0,0 +1,86 @@
...
+					name:       vol.Name,
+					mountPoint: vol.MountPoint,
+				}
+				v, err := plugin.newVolume(opts)
+				if err != nil {
+					return err
+				}
+				vs.volumes = append(vs.volumes, v)
+			}
+		}
```
gregarndt: The task payload is validated against a schema before it gets to this point and i set additionalProperties to False so it should fail schema validation.

plugins/volume/persistent.go
gregarndt: Are you referring to assigning one of the return values to a field of 'v' ?  If so, that's good to know! Do you have a link to some documentation about that so I can educate myself?

Fortunately this project was born once 1.5 was already out and we are running tests against 1.6 now so we should be able to have >=1.5 as a requirement for taskcluster-worker.

plugins/volume/persistent_test.go
gregarndt: Doesn't seem to work when it's at the package level and not inside the body of a func.

plugins/volume/persistent_test.go:20:1: expected declaration, found 'IDENT' logger

plugins/volume/persistent.go
gregarndt: yea if we want to get clever later with how we search we can, I don't expect any of this to be carved in stone.  I can't imagine we would have more than a couple of volumes per name because most of our workers are running with only 1 task at a time anyways.

plugins/volume/persistent.go
gregarndt: Ah that's much better language, thank you!

plugins/volume/persistent.go
gregarndt: I'm not expert on the string formatting in Go yet, but from a quick skim it seems that if the object implements the Error interface, it's converted to string first, and the default formatting for a String is %s when %v is used so in the end it all boils down to %s.  I'm not sure what's idiomatic in Go though for doing something like this.

plugins/volume/volumeset.go
gregarndt: Do you think we should still handle if the volume type doesn't exist?  I don't think we should panic since that would cause the worker to crash if someone submitted a bad task payload, but we could log a task error.

plugins/volume/persistent.go
walac: I mean, you don't need to declare `err` beforehand because the other receiver of the assignment already exists. New versions of the language are smart enough.

plugins/volume/persistent.go
walac: Neither do I. @petemoore ?

plugins/volume/persistent_test.go
walac: :o that's surprising

plugins/volume/persistent.go
gregarndt: Hrm, using go 1.6, if I change this, I get:
# github.com/taskcluster/taskcluster-worker/plugins/volume
plugins/volume/persistent.go:73: non-name v.volume on left side of :=

plugins/volume/volumeset.go
gregarndt: ok, I think I addressed this in 391f7ab

plugins/volume/persistent.go
petemoore: My code is littered with both!

pmoore@Petes-iMac:~/go/src/github.com/taskcluster $ grep -rF '%s' . | grep '\.go' | wc -l
     232
pmoore@Petes-iMac:~/go/src/github.com/taskcluster $ grep -rF '%v' . | grep '\.go' | wc -l
     330

So I ran a test to see if I could find differences. In short, both seem not to interpret the bytes of the string in any way, and are equivalent:

https://play.golang.org/p/ffEIkRmIJa

So I think either should be fine.

gregarndt: Thanks Pete for the deep look! yea from my quick reading I found that strings are formated using %s if %v is used, and errors that are passed in get treated as a string, which in turn uses %s so I think everything matches up.  It's just a matter of finding a consensus for how we do it in taskcluster-worker.

walac:
r+

PR #80, opened 2016-06-23
Implement live logging

gregarndt:


walac:
@gregarndt could you please rebase this branch?

gregarndt:
I know that @jonasfj is doing a lot of refactoring as part of his qemu and config loading stuff, it's possible that this has a lot of conflicts with that effort.  I would talk with him about what will go on with this and if this is thing he'll want to rip out and replace.

jonasfj:
yeah, let's see what it looks like when I land that stuff...

gregarndt:
Either way, do know that a solution to live logging needs to be in place soon to deploy on releng machines for os x tests.

walac:
I've rebased the code.

plugins/livelog/livelog.go
```diff
@@ -1,4 +1,125 @@
...
+			line, isPrefix, err := reader.ReadLine()
+			if err != nil {
+				return
+			}
+			w.Write(line)
+			if !isPrefix {
+				w.Write([]byte("\r\n"))
+			}
+
+			flusher.Flush()
```
jonasfj: Flushing every line seems expensive... we would have to turn of nagle... and if there is a lot logs this would be bad...

I propose that we flush every 500ms or something like that..

runtime/taskcontext.go
```diff
@@ -80,25 +84,38 @@ type TaskContextController struct {
...
+	server webhookserver.WebHookServer) (*TaskContext, *TaskContextController, error) {
 	logStream, err := stream.New(tempLogFile)
 	if err != nil {
 		return nil, nil, err
 	}
 	ctx := &TaskContext{
-		logStream: logStream,
-		TaskInfo:  task,
+		logStream:   logStream,
+		logLocation: tempLogFile,
```
jonasfj: What do you need `logLocation` for?

runtime/taskcontext.go
```diff
@@ -195,6 +212,15 @@ func (c *TaskContext) NewLogReader() (io.ReadCloser, error) {
 	return c.logStream.NextReader()
 }
 
+func (c *TaskContext) ExtractLog() (ioext.ReadSeekCloser, error) {
+	file, err := os.Open(c.logLocation)
```
jonasfj: There is already an interface for reading the log... I don't think we should expose a seek interface..

jonasfj: This opens for bugs if people use this interface before the log is closed...

runtime/taskcontext.go
```diff
@@ -80,25 +84,38 @@ type TaskContextController struct {
...
 // CloseLog will close the log so no more messages can be written.
 func (c *TaskContextController) CloseLog() error {
+	if c.logClosed {
+		return nil
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.logClosed = true
 	return c.logStream.Close()
```
jonasfj: hmm... I have mixed feeling here..
Ie. should we panic if we try to close twice, or always allow multiple close calls..  I guess this is easier :)

runtime/taskcontext.go
gregarndt: It's necessary for retrying artifact uploading if I recall correctly.  It's been awhile since I worked on this.

runtime/taskcontext.go
gregarndt: It's used later to open up a readseekcloser to the actual temp file that was created for storing the log.

runtime/taskcontext.go
jonasfj: Hmm.... I guess it's totally overkill to suggest that we use the existing
interface... Can we at-least agree to return an ErrLogNotClosed error until
the log is closed... So we avoid races here... Obviously using this method
outside of Finished handler likely won't work...

Den 6. sep. 2016 12.26 PM skrev "Greg Arndt" <notifications@github.com>:

config/config.go
```diff
@@ -133,6 +146,33 @@ var ConfigSchema = func() runtime.CompositeSchema {
...
+		          "description": "Secret used when generating stateless hostname",
+		          "title": "Secret",
+		          "type": "string"
+		        }
+		      },
+		      "required": [
+		        "enabled",
+		        "secret",
+		        "domain"
+		      ],
```
petemoore: we should probably have `"additionalProperties": false` here too, for safety...

config/global-config.yml
```diff
@@ -89,6 +89,31 @@ properties:
...
+        type: string
+      domain:
+        title: Domain
+        description: |-
+          Domain to use when generating stateless hostname
+        type: string
+    required:
+      - enabled
+      - secret
+      - domain
```
petemoore: we should probably have an extra (indented) `additionalProperties: false` here too for `statelessHostname`...

petemoore:
lgtm

runtime/taskcontext.go
gregarndt: @walac Do you have a moment to implement the error handling Jonas suggested above? It's a great idea to prevent races here.

walac:
I think I addressed all @jonasfj comments

runtime/taskcontext.go
```diff
@@ -63,12 +73,14 @@ type TaskInfo struct {
...
 	TaskInfo
-	webHookSet *webhookserver.WebHookSet
-	logStream  *stream.Stream
-	mu         sync.RWMutex
-	queue      client.Queue
-	status     TaskStatus
-	cancelled  bool
+	webHookSet  *webhookserver.WebHookSet
+	logStream   *stream.Stream
+	logLocation string
```
petemoore: Is `logLocation` an absolute path, a relative path to task home directory, does it have engine-specific path separator or a canonical `/` separator, etc? Can you add comments for these struct members?

gregarndt: Hopefully Wander can add some documentation here, but logLocation is a string that's a path to where you can find the task log file, so absolute path.  This is within the task context, which means the separators are whatever are used on the platform the worker is running on.  In this case task context logLocation is the location of the temp log file that's passed into NewTaskContext which then should probably have similar documentation as to what that should be.  The value that's passed into NewTaskContext is supplied by the environment (outside of the engine) and I believe uses the os golang package for creating temporary locations and constructing paths with the appropriate path separator.

jonasfj:
FYI, I'm working to merge this...

walac:
Nice then, I always working on it too...

jonasfj:
There is still a few dirty things. Eventually we'll have to figure out how to setup the environment in a way that it's not just a long method in `worker.New` and that we have the config schema for the individual parts located  outside worker package.

walac:
Something is wrong, I see patches from other PRs here

PR #81, opened 2016-06-28
Added vendoring

petemoore:
I looked at [the go wiki](https://github.com/golang/go/wiki/PackageManagementTools) to find recommended vendoring tools. The first one listed, [manul](https://github.com/kovetskiy/manul) looked nice in principle (because it uses git submodules to keep history of original project) but when I tried to freeze versions, I ran into all sorts of problems. I then started reading up on `git submodule` vs `git subtree` (e.g. see [these](https://medium.com/@porteneuve/mastering-git-subtrees-943d29a798ec#.fbub38zi5) [blogs](https://medium.com/@porteneuve/mastering-git-submodules-34c65e940407#.kqjkpkl2o)) but the whole story seemed very complicated. So in the end I resorted to good old [godep](https://github.com/tools/godep). It seems to still be pretty standard. It copies source, rather than preserving git history of dependencies, but hey, it is simple.

Having just broken `jsonschema2go`, it was a good way to try it out. In this initial revision, I've pinned `jsonschema2go` to an old revision, to overcome the bustage I caused by updating the library. It seems to work! Note, fresh builds not using vendoring at the moment would be broken, due to my `jsonschema2go` recent changes (which incidentally I'll be fixing shortly).

I've assigned all three of you just because it is major-impact - so I wanted everyone to be onboard. Don't feel obliged to provide a lot of feedback, as long as one person has a jolly good look, we should be fine. I just wanted to make sure you're all aware about it.

Thanks chaps!


jonasfj:
git submodule is way too complicated... I'm super happy we didn't go that way!

Makefile
```diff
@@ -28,8 +29,20 @@ test:
...
+	go vet github.com/taskcluster/taskcluster-worker/plugins/...
+	go vet github.com/taskcluster/taskcluster-worker/runtime/...
+	go vet github.com/taskcluster/taskcluster-worker/worker/...
+	golint github.com/taskcluster/taskcluster-worker/codegen/...
+	golint github.com/taskcluster/taskcluster-worker/commands/...
+	golint github.com/taskcluster/taskcluster-worker/config/...
+	golint github.com/taskcluster/taskcluster-worker/engines/...
+	golint github.com/taskcluster/taskcluster-worker/plugins/...
+	golint github.com/taskcluster/taskcluster-worker/runtime/...
+	golint github.com/taskcluster/taskcluster-worker/worker/...
```
petemoore: Note, this seems to be a limitation in the language (not related to the vendor tool). If you look at `go help packages` you'll see that `./...` would be expanded to match all packages under the current package, but strangely this also seems to include vendored packages, so `./...` or even `github.com/taskcluster/taskcluster-worker/...` are no good to us. For this reason I've had to list the packages explicitly, I haven't found a way around this.

gregarndt:
definitely +1 to vendoring, we've been bit by this before and it makes me indeed sad that there aren't better ways to do this. 

I've never used git submodules, but from the pains people exclaim, I'm definitely happy to not have to use them.

jonasfj:
# `r+`

PR #82, opened 2016-06-28
Just a nit docs improvement for readability

jonasfj:


petemoore:
Thanks - that is much better!

PR #83, opened 2016-06-29
Updated dependencies to latest versions

petemoore:
Basically I didn't want to have to support multiple versions of `jsonschema2go` library, so now that it is fixed to produce same content as before, I've updated the dependencies to be the latest.

Note the code generated by jsonschema2go is unchanged in any way (`git diff` will show no changes).

Since I updated all dependencies, it has brought in some other changes too, but I think we might as well run on the latest versions, unless we find a bug that makes us better off pinning.


gregarndt:
So it looks like there is one file change in this repo (godeps) where it's a different revision for one package, and a few other packages added resulting in a large PR.  If that's all that changed, I'm 👍

petemoore:
That's right. The vendor folder picked up the `testify` package, which then has its own vendored dependencies too, hence the rather large diff. This is nice, so test dependencies are also vendored, so we should be able to get predictable results in tests too, and not have tests break because some dependency introduced changes.

PR #84, opened 2016-07-06
Qemu

jonasfj:
@petemoore I'm having some issues with godep that removed a lot of stuff... And I couldn't get `godep save ./...` to work... basically, it would build fine, but when I moved it docker where I didn't have any deps it wouldn't work...

`godep save ./...` ignored the `jsonschema2go/text` for some reason... I see here that it also removed some other dependencies...

As for review, I think given that much what I'm landing is changing other things we should just land it...
I've disabled test that don't run without kvm to use a build flag and I only include the QEMU engine on linux....

@walac I've also disabled osxnative tests on platforms other than darwin using build tags...

@petemoore For review probably mainly look at things outside the `engines/qemu/` folder...You are welcome to review the insides... But I would prefer to land this rather than blocking on small things in the engine implementation that isn't done yet.


jonasfj:
@petemoore, if you can figure out godep please feel free to push to this branch... I tried but `go dep save ./...` doesn't seem to get me all the dependencies...

walac:
@jonasfj I disabled it as well in PR #77

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang:1.6
+MAINTAINER Jonas Finnemann Jensen <jopsen@gmail.com>
+
+RUN apt-get update -y
+RUN apt-get upgrade -y
+RUN apt-get install -y qemu dnsmasq-base build-essential liblz4-tool iptables
```
walac: I would run `apt-get autoremove` here as well, just for the sake of having a clean environment.

commands/qemu-build/cmd.go
```diff
@@ -0,0 +1,159 @@
...
+
+type cmd struct{}
+
+func (cmd) Summary() string {
+	return "Build an image for the QEMU engine"
+}
+
+func (cmd) Usage() string {
+	return `
+taskcluster-worker qemu-build takes
```
petemoore: takes what?

commands/qemu-guest-tools/guesttools.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/Sirupsen/logrus"
```
petemoore: Did you want to get rid of `logrus`?

commands/qemu-guest-tools/guesttools.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/taskcluster/go-got"
```
petemoore: nit: when the package name differs from the directory name, I like to be explicit, e.g. `import got "github.com/taskcluster/go-got"` - just to aid with visual understanding of code - but not a biggy

engines/osxnative/engine.go
```diff
@@ -20,11 +21,6 @@ type engineProvider struct {
 	extpoints.EngineProviderBase
 }
 
-func init() {
-	// Register the mac engine as an import side-effect
-	extpoints.EngineProviders.Register(engineProvider{}, "macosx")
-}
-
```
petemoore: Did you mean to commit this?

engines/qemu/config-schema.yml
```diff
@@ -0,0 +1,25 @@
+$schema:            http://json-schema.org/draft-04/schema#
+title:              "Engine Config"
+type:               object
+properties:
+  maxConcurrency:
+    type:           integer
+    minimum:        1
+    maximum:        100
+    description: |
```
petemoore: I think `|=` rather than `|` will avoid the trailing new line, iirc.

petemoore:
That was an extremely superficial review, but r+ from me.

@gregarndt @walac can you take a look too?

engines/qemu/config-schema.yml
jonasfj: Who cares about trailing lines in markdown... (description is rendered as markdown in our docs)...

But good point...

engines/osxnative/init_darwin.go
```diff
@@ -0,0 +1,8 @@
+package osxnative
+
+import "github.com/taskcluster/taskcluster-worker/engines/extpoints"
+
+func init() {
+	// Register the mac engine as an import side-effect
+	extpoints.EngineProviders.Register(engineProvider{}, "macosx")
+}
```
jonasfj: because the file is darwin only... the other files can compile on all platforms... at least for now...

@walac should totally overwrite what I did here when we merge his patch (hopefully soon too)...

.gitignore
```diff
@@ -1,3 +1,4 @@
 /taskcluster-worker
+/.vagrant
```
gregarndt: isn't this the same as line 4?

jonasfj: hmm... probably some merge issue... I'll fix that...

PR #85, opened 2016-07-07
Moved remaining dependencies into vendor folder

petemoore:
There were some `go get` commands in the `Makefile` which meant that build steps were potentially downloading code. Since all code should now be vendored, I changed `go get` statements for `go install` to make sure no new code was getting downloaded. I also simplified the `Makefile` using [this suggestion](https://github.com/golang/go/issues/11659#issuecomment-122139338).


petemoore:
Note, this is just a change to `Makefile` plus adding some new vendored dependencies.

.travis.yml
```diff
@@ -25,12 +25,12 @@ before_install:
   - go get github.com/pierrre/gotestcover
 
 install:
-  - "go get -d -t ./..."
+  - "echo No go get steps since we have vendored packages..."
```
petemoore: Note, I put an `echo` statement in here in case there was a default `install` target that did the `go get`, I wanted to make sure it is overridden. It also serves as a comment in case someone goes to add it, now they know not to.

petemoore:
r+ from jonasfj in #taskcluster-worker

PR #86, opened 2016-07-07
Convert volumes object into array of volumes

gregarndt:
This is to have a volumes array rather than an object keyed based on the
volume type.

### DIFF
--- a/plugins/volume/volumeset.go
+++ b/plugins/volume/volumeset.go
@@ -16,10 +16,23 @@ type volume interface {
 }
 
 type volumeOptions struct {
-	taskID     string
-	runID      int
-	name       string
-	mountPoint string
+	taskID string
+	runID  int
+	spec   spec
+}
+
+type spec struct {
+	// Path to mount volume into task environment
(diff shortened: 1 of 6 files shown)


plugins/volume/volumeset.go
```diff
@@ -38,28 +51,24 @@ type volumeSet struct {
...
 				"Unrecognized volume type. This should never happen and " +
 					"is considered a fatal error.",
 			)
 		}
+		opts := &volumeOptions{
+			taskID: context.TaskInfo.TaskID,
+			runID:  context.TaskInfo.RunID,
+			spec:   spec(volume),
+		}
+		v, err := plugin.newVolume(opts)
```
gregarndt: maybe implement opts as a "map[string]interface{}" and then the plugin will do what it needs to with the options.

petemoore:
@gregarndt is this PR ready for review?

gregarndt:
I think this is going to be closed.  @jonasfj is rewriting it.

jonasfj:
Yes, let's closet this and rethink volumes and how they take options.

PR #87, opened 2016-07-22
godep update

jonasfj:
@petemoore , this took forever to figure out... but `godep update ./...` is the right thing... `godep save ./...` only adds new dependencies...

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ The simplest is probably:
 go get -u github.com/tools/godep      # install godep tool
 godep restore ./...                   # copy vendored dependencies into your GOPATH
 go get -u -t ./...                    # update versions
-godep save ./...                      # save changes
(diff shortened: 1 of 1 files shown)


jonasfj:
Okay, now it makes more sense... `godep save ./...` will just add deps that aren't already there... and save existing dependencies to what the godep file says...

`godep update ./...` will update the godep file to latest version of everything... based on what is currently checked out...

petemoore:
Thanks for this Jonas! That explains a lot. =)

It sounds like we should have an explanation of both commands (since sometimes we'll add new dependencies) - can you document both, rather than removing the details about `godep save`?

jonasfj:
@petemoore, better?

petemoore:
> @petemoore, better?

beautiful! :+1:

PR #88, opened 2016-07-31
Qemu polling

jonasfj:
Artifact extraction and interactive shell for qemu engine...

Also improve a lot of test cases and scripted test image building... updated docker test env...

@petemoore, I know this is a big thing to review... Maybe we just do a meeting and talk about it :)


commands/qemu-guest-tools/actions_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"net/http/httptest"
+	"net/url"
+	"os"
+	"path/filepath"
+	goruntime "runtime"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
```
walac: Empty lines between imports make `gofmt` erroneously assume import list has finished.

jonasfj: I just let it do what my `go fmt` does automatically... maybe we should check that we're running the same formatting tools :)

And latest version of said formatting tools... I'll admit my editor runs a bunch, so I'm not even 100% sure which I'm using... But I would guess `go fmt`...

petemoore: This is unfortunately standard behaviour of [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) which varies slightly from `go fmt`, by introducing the line break and separating standard libraries from other imports. See e.g. [this issue](https://github.com/golang/go/issues/9922).

commands/qemu-guest-tools/actions_test.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		fmtPanic(append(a, err)...)
+	}
+}
+
+func assert(condition bool, a ...interface{}) {
+	if !condition {
+		fmtPanic(a...)
+	}
+}
```
petemoore: It is a shame go doesn't have such functions by default...

commands/qemu-guest-tools/guesttools.go
```diff
@@ -1,27 +1,35 @@
 package qemuguesttools
```
petemoore: Package comments would be useful here.

engines/qemu/image/README.md
```diff
@@ -10,51 +10,3 @@ following files:
...
-and copying the files onto the system.
-
-```sh
-./taskcluster-worker qemu-build \
-  --size 1 \
-  --boot tinycore.iso \
-  --cdrom data.iso \
-  from-new machine.json \
-  tinycore-worker.tar.lz4
-```
```
petemoore: Did you mean to delete these docs?

engines/qemu/metaservice/messages.go
```diff
@@ -0,0 +1,44 @@
...
+// to request that the executing be aborted.
+//
+// If [type] is MessageTypeExit then [data] = [exitCode], where exitCode is a
+// single byte 0 (success) or 1 (failed) indicating whether the command
+// terminated successfully.
+const (
+	MessageTypeData  = 0
+	MessageTypeAck   = 1
+	MessageTypeAbort = 2
+	MessageTypeExit  = 3
```
petemoore: Using [iota](https://golang.org/ref/spec#Iota) would be more idiomatic here.

engines/qemu/metaservice/messages.go
```diff
@@ -0,0 +1,44 @@
...
+	StreamStdin      = 0
+	StreamStdout     = 1
+	StreamStderr     = 2
```
petemoore: Here too, of course. :-)

engines/qemu/metaservice/messages.go
jonasfj: Nope, this is a format spec... Would be bad to have changes...

engines/qemu/metaservice/messages.go
jonasfj: Again, these constants are frozen

engines/qemu/image/README.md
jonasfj: Yes, I moved some of it to tedt-image folder

engines/qemu/metaservice/metaservice.go
```diff
@@ -176,3 +209,298 @@ func (s *MetaService) handleUnknown(w http.ResponseWriter, r *http.Request) {
...
+		// Check content-type
+		if r.Header.Get("Content-Type") != "application/json" {
+			reply(w, http.StatusBadRequest, Error{
+				Code:    ErrorCodeInvalidPayload,
+				Message: "Content-Type must be application/json",
+			})
+			return
+		}
+
+		// Read first 10 MiB of body (limited for safety)
```
petemoore: nice

engines/qemu/metaservice/messages.go
petemoore: Ah OK, that makes sense.

engines/qemu/metaservice/messages.go
petemoore: Yeah, ok.

commands/qemu-guest-tools/actions_test.go
```diff
@@ -0,0 +1,261 @@
...
+	go io.Copy(ioutil.Discard, shell.StderrPipe())
+	go func() {
+		if goruntime.GOOS == "windows" {
+			shell.StdinPipe().Write([]byte("type con\n"))
+		} else {
+			shell.StdinPipe().Write([]byte("exec cat -\n"))
+		}
+		// Give cat - some time to start, bash/busybox/dash won't work otherwise
+		// Or they will work, but only intermittently!!!
+		time.Sleep(250 * time.Millisecond)
```
petemoore: Will this cause test to last at least 250ms?

jonasfj: I boot a virtual machine like 10 times during testing... 250ms is nothing :) hehe

runtime/ioext/asyncpipe.go
```diff
@@ -0,0 +1,128 @@
...
+	capacity int
+}
+
+// AsyncPipe is similar to io.Pipe() except that writes isn't blocking, instead
+// data will be added to an internal buffer that can grow up to specified
+// capacity.
+// Additionally, you may supply a channel tell that will be told whenever
+// N bytes have been read, so that more bytes can be requested to be written.
+//
+// This pipe kind is useful when implementing congestion control.
```
petemoore: I think your `AsyncPipeWriter` could be achieved with a regular `byte` channel with the given `capacity`. This would certainly be more idiomatic, less code to manage/maintain, easier for contributors to reason about (as it would use standard low-level go mechanics) and probably would be similarly efficient. What are the drawbacks of using a regular channel with the given capacity?

runtime/ioext/readatmost.go
```diff
@@ -0,0 +1,73 @@
...
+var ErrMaxSizeExceeded = errors.New("MaxSize was exceeded before EOF was reached")
+
+// ReadAtmost will read at-most maxSize bytes from r and return an error if we
+// didn't reach EOF. Returns ErrMaxSizeExceeded if r contains more than maxSize
+// bytes. If maxSize is -1 ReadAtmost will read everything.
+//
+// This utility is useful when reading HTTP requests, in particular if reading
+// from untrusted sources. If using io.ReadAll it's easy to run the server out
+// of memory, a maxSize of 2MiB is usually sane and prevents such attacks.
+func ReadAtmost(r io.Reader, maxSize int64) ([]byte, error) {
```
petemoore: ReadAt**M**ost (capital 'M')

engines/qemu/metaservice/shell.go
```diff
@@ -0,0 +1,323 @@
...
+		if err != nil {
+			s.resolve.Do(func() {
+				debug("Resolving with internal-error, failed to send ping, error: %s", err)
+				s.success = false
+				s.err = engines.ErrNonFatalInternalError
+				s.dispose()
+			})
+			return
+		}
+	}
```
petemoore: Are you sure you don't want the `sendPings` method to spawn its own go routine? Seems like something that would always run as a background process, so should have the responsibility to spawn the go routine itself.

engines/qemu/metaservice/shell.go
```diff
@@ -0,0 +1,323 @@
...
+			s.err = engines.ErrNonFatalInternalError
+			s.dispose()
+		})
+	}
+}
+
+func (s *shell) sendPings() {
+	for {
+		// Sleep for ping interval time
+		time.Sleep(ShellPingInterval)
```
petemoore: [time.Tick](https://golang.org/pkg/time/#Tick) would be more idiomatic here (sorry to keep using that word!) 😜

engines/qemu/network/pool.go
```diff
@@ -362,8 +368,9 @@ func destroyNetwork(n *entry) error {
 		{"ip", "addr", "del", n.ipPrefix + ".1", "dev", n.tapDevice},
 		// Delete tap device
 		{"ip", "tuntap", "del", "dev", n.tapDevice, "mode", "tap"},
-	})
+	}, true)
 	if err != nil {
+		debug("Failed to destory tap device: %s, error: %s", n.tapDevice, err)
```
petemoore: This code is so scary, I think it warrants a few more comments to explain the finer points, or give an overview of how the network gets configured, and why. For most people coming fresh to this, it would take a lot of study/investigation to understand it.

engines/qemu/test-image/setup.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh -e
+
+# Install taskcluster-worker for the qemu-guest-tools
+sudo cp /mnt/sr0/taskcluster-worker /opt/
+sudo chmod +x /opt/taskcluster-worker
+
+# Install a startup script to launch taskcluster-worker after X.org
+cp /mnt/sr0/worker.sh /home/tc/.X.d/worker.sh
+chmod +x /home/tc/.X.d/worker.sh
```
petemoore: These look like hard coded paths that might not exist on some people's machines - does this need to be moved out into a configuration file / platform-specific file?

petemoore:
Where do the binary files such as `engines/qemu/test-image/tinycore-setup.tar.lz4` come from? Are the procedures for generating all binary files included in the repo?

petemoore:
I feel I've only scratched the surface with this first look through. @gregarndt can you also take a look?

I'll leave it like that for now, when you've replied to my current questions, I can add some more - rather than dumping too many questions all in one go.

petemoore:
(I've merged in master ^^^, to pick up test fixes)

commands/qemu-guest-tools/actions_test.go
walac: hrm, we should be consistent in the team how we do this. Currently I rely on `gofmt` through [vim-go](https://github.com/fatih/vim-go) for code formatting. I will check how I can change it to use `goimports`.

commands/qemu-guest-tools/actions_test.go
walac: I am using [this](https://github.com/stretchr/testify)

runtime/ioext/asyncpipe.go
jonasfj: no, it can't the semantics for sending when buffer is full is non-blocking, returns an error..
Also channels would be crazy slow... a channel has a cost per message, if each byte is a message.. well, let's just leave at that would be very very slow, unless go has some builtin magic for chan byte, which I doubt...

jonasfj: Also, you can't tell when stuff has been read from a channel... the `tell` feature is pretty essential here... It tells that bytes have been consumed... so it's safe to request additional bytes to be written to the pipe...

I don't want to store infinite bytes in memory, therefore there is a protocol for requesting additional bytes to be sent.

commands/qemu-guest-tools/actions_test.go
jonasfj: I'm too lazy to carry around a reference to the `t` object... I just panic instead - so simple :)

engines/qemu/metaservice/shell.go
jonasfj: I do, when I call `go sendPings()`

engines/qemu/metaservice/shell.go
jonasfj: yeah, that is nice... except it leaks! You can't stop the ticker ever... See:
https://golang.org/pkg/time/#Tick

engines/qemu/network/pool.go
jonasfj: Yeah, I'm not happy with the setup either... In particular I'm having some reliability issues...

So I might have to redo the network setup... Hence, I'm not going to go ahead with a huge documentation story on yet either,...

PR #89, opened 2016-08-01
Fix problems with travis-ci/gimme#42 and travis-ci/gimme#25

petemoore:
I think this should fix it.


petemoore:
It looks like this needs a bit of work, the go installation is ignoring `GOOS` and `GOARCH` - let me fix this and push another commit...

petemoore:
These are [genuine failures](https://travis-ci.org/taskcluster/taskcluster-worker/jobs/148913697#L265-L454) - 4 race conditions.

PR #90, opened 2016-08-03
VNC tests

jonasfj:
This adds a VNC server to mockengine... and a vnc client to `enginetests`...

And it does VNC with full tests for QEMU... + a bunch of test stability fixes, eliminating some leaks and other fun stuff...


petemoore:
r+

PR #93, opened 2016-08-13
Nested kvm

jonasfj:
The interesting thing to look at here is the interface changes I made to `Sandbox`..

You might also want to notice that I have a CLI client using the interactive shell... it supports terminal resize and everything, pretty cool :)

As for reviewing all this code, I suggest you just give a quick skim... I'm still iterating on some of it.. And a lot of it's actually refactoring existing code...

Ps. Yes, setting CPU to host, allows me run QEMU/kvm under QEMU/kvm, so we have hardware accelerated virtual machines inside our task :) Useful for image builds or crazy tests where people want to run two VMs next to each other or play with the network between two machines... hehe

Yes, I actually have QEMU engine tests running under qemu engine...


jonasfj:
@petemoore, I also moved all the test images to S3... And added scripts to download/update the images... I suspect that'll make you super happy...

If you and @walac  are up for it I'm open to modifying git history to remove the binary images.
I thought I would never to change them, but I will... and the repo is being huge (slow to clone), so maybe we should edit history filter them out and force push...

petemoore:
Yes, happy to adapt the git history. This isn't in production yet, so should only impact a couple of us, but for sure will be better for the long run.

petemoore:
I'm going to rubber stamp this as it __mostly__ looks like it only affects qemu engine, and we're not in production with that yet. And the tests pass.

r+
==

petemoore:
@walac Over to you - as you are the biggest user of taskcluster-worker at the moment...

commands/qemu-build/buildimage.go
```diff
@@ -78,6 +78,9 @@ func buildImage(
 		return err
 	}
 
+	// Setup logService so that http://169.254.169.254/v1/log
```
walac: Unfinished comment here?

commands/qemu-guest-tools/cmd.go
```diff
@@ -46,6 +60,29 @@ func (cmd) Execute(arguments map[string]interface{}) bool {
...
 	g := new(host, log)
+
+	if arguments["post-log"].(bool) {
+		logFile := arguments["<log-file>"].(string)
+		var r io.Reader
+		if logFile == "-" {
+			r = os.Stdin
+		} else {
+			f, err := os.Open(logFile)
+			if err != nil {
```
walac: Better logging the error somehow before return.

commands/qemu-guest-tools/cmd.go
```diff
@@ -46,6 +60,29 @@ func (cmd) Execute(arguments map[string]interface{}) bool {
...
+			defer f.Close()
+			r = f
+		}
+		w, done := g.CreateTaskLog()
+		_, err := io.Copy(w, r)
+		if err != nil {
+			err = w.Close()
+			<-done
+		}
+		return err == nil
```
walac: Better logging the error somehow before return.

commands/qemu-guest-tools/guesttools.go
```diff
@@ -304,23 +306,25 @@ func (g *guestTools) doExecShell(ID string) {
...
-	// Create a shell
-	shell := exec.Command("sh")
-	shell.Stdin = handler.StdinPipe()
-	shell.Stdout = handler.StdoutPipe()
-	shell.Stderr = handler.StderrPipe()
+	// Set command to standard shell, if no command is given
+	if len(command) == 0 {
+		command = []string{"sh"}
+		if goruntime.GOOS == "windows" {
+			command = []string{"cmd.exe"}
```
walac: Maybe Power Shell instead of the cmd?

commands/qemu-guest-tools/pty_unix.go
```diff
@@ -0,0 +1,72 @@
...
+		handler.Communicate(func(cols, rows uint16) error {
+			return setTTYSize(f, cols, rows)
+		}, func() error {
+			if cmd.Process != nil {
+				return cmd.Process.Kill()
+			}
+			return nil
+		})
+	}
+	// If pty wasn't supported for some reason we fall back to normal execution
```
walac: nit: log pty failed and we fallback to normal execution.

commands/shell/tty_unix.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Handle SIGWINCH signals (window resize signals from graphical terminals)
+	sigWinch := make(chan os.Signal, 1)
+	signal.Notify(sigWinch, syscall.SIGWINCH)
+	go func() {
+		for range sigWinch {
+			size, err := ttyutils.Winsize(os.Stdout)
+			if err == nil && setSize != nil {
+				setSize(size.Columns, size.Lines)
+			}
```
walac: I don't have much context here, but it feels like this code races. If the window size is modified between the calls to `WinSize` and `setSize`, it updates the size with invalid values. Not sure if it has practical consequences, or either if it has a non-racy solution.

engines/qemu/test-image/upload.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -e
+
+# Find location of the script no matter where it's located
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+echo "Skipping tinycore-setup.tar.lz4 as we don't change it"
+# aws s3 cp "$DIR/tinycore-setup.tar.lz4" "s3://public-qemu-images/test-images/tinycore-setup.tar.lz4"
```
walac: Don't we need to upload this?

plugins/interactive/shellclient/shellclient.go
```diff
@@ -278,25 +275,58 @@ func (s *shell) readMessages() {
 	}
 }
 
-func (s *shell) StdinPipe() io.WriteCloser {
+// StdinPipe returns a pipe to which stdin must be written.
+// It's important to close stdin, if you expect the remote shell to termiante.
```
walac: s/termiante/terminate

plugins/interactive/shellclient/shellclient.go
```diff
@@ -314,7 +344,8 @@ func (s *shell) Abort() error {
 	return s.err
 }
 
-func (s *shell) Wait() (bool, error) {
+// Wait will wait for remote shell to terminate and success, and error if any.
```
walac: nit: "Wait will wait for the remote shell to finish, by either succeeding or returning an error."

plugins/interactive/shellhandler.go
```diff
@@ -52,66 +54,71 @@ func newShellHandler(ws *websocket.Conn, log *logrus.Entry) *shellHandler {
 		tellIn:       tellIn,
 	}
 
-	ws.SetReadLimit(metaservice.ShellMaxMessageSize)
+	ws.SetReadLimit(ShellMaxMessageSize)
 
 	return s
 }
 
+// A SetSizeFunc can set the set the TTY size of a ter
```
walac: nit: "A SetSizeFunc can set the TTY size of a terminal."

plugins/interactive/shellserver.go
```diff
@@ -0,0 +1,192 @@
...
+		log:       log,
+	}
+	s.c.L = &s.m
+	return s
+}
+
+// Wait will wait for all active shells to be done and return
+func (s *ShellServer) Wait() {
+	s.m.Lock()
+	defer s.m.Unlock()
```
walac: What are we protecting here? `refCount`? If so, how is its value supposed to reach zero if the mutex is hold in the `for` iteration?

plugins/interactive/shellserver.go
```diff
@@ -0,0 +1,192 @@
...
+	for s.refCount > 0 {
+		s.c.Wait()
+	}
+}
+
+// WaitAndClose will wait for all active shells to be done and abort creation
+// of all new shells atomically.
+func (s *ShellServer) WaitAndClose() {
+	s.m.Lock()
+	defer s.m.Unlock()
```
walac: Same point as the previous comment.

plugins/interactive/shellserver.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+}
+
+func (s *ShellServer) updateRefCount(change int) {
+	s.m.Lock()
+	s.refCount += change
+	if s.refCount <= 0 {
+		s.c.Broadcast()
+	}
+	s.m.Unlock()
```
walac: Doesn't this deadlock with `Wait` and `WaitAndClose`? If we call `Wait` with ```refCount > 0```, then `Wait` will wait forever the value reaches zero, and `updateRefCount` will wait forever the mutex be released.

walac:
@jonasfj review done. I didn't put eyes on vendor stuff, however.

jonasfj:
> review done. I didn't put eyes on vendor stuff, however.
@walac, seriously, I would never expect you to review anything  in `vendor/`.. if that builds vendoring is great :)

commands/qemu-guest-tools/guesttools.go
jonasfj: yeah, maybe... I don't know yet... I don't think we have powershell on say windows xp...
I suspect that as I come further guest-tools will be configured with what shell to use, maybe if people install bash on windows that's an even better option for interactive shells :)

commands/shell/tty_unix.go
jonasfj: I don't think it has practical consequences... and it shouldn't leave the window in an incorrect size..
Because the `sigWinch` channel has capacity 1, so if a resize event happens, while processing the previous resize event... Then once the first resize event has been processed, the while-loop will get another message (the second resize event) from the `sigWinch` channel..

I didn't realize this until you pointed it out and forced me to consider it. But by having a capacity of one it doesn't matter how many resize events happen while we're processing the first resize event. It's true that only one even will be added to the channel, but since the message in the channel doesn't carry the window size it should all be okay...

Hmm... Perhaps the golang signal handling stuff isn't so crazy afterall.. I originally thought it was a bit weird.

engines/qemu/test-image/upload.sh
jonasfj: Only, need to update it if we change it... And changes are manual, so we won't do that very often... I'll add additional comment..

plugins/interactive/shellclient/shellclient.go
jonasfj: Wow, nice catch!

plugins/interactive/shellserver.go
jonasfj: If you haven't played with mutex/cond, you should have a look at:
https://golang.org/pkg/sync/#Cond

I used to think they were really hard concepts in C/C++. Similar primitives exists in pthreads.
Basically, `s.c.Wait()` in the for-loop below will release the lock and wait for `s.c.Signal()` or `s.c.Broadcast()` to be called... Then it'll reacquire the lock again and continue execution.

So by doing `s.c.Broadcast()` everytime I've changed `s.refCount` I ensure that after:
```go
	for s.refCount > 0 {
		s.c.Wait()
	}
```
`s.refCount` is zero and I hold the lock.

plugins/interactive/shellserver.go
jonasfj: `s.c.Wait()` releases the lock while it waits for `s.c.Broadcast()`.

Note: for this to work I must do `s.c.L = &s.m` in `NewShellServer`.

jonasfj: @walac, you might want to check out the sync package and play with this pattern.. It's pretty cool as it basically allows you to wait for any condition to occur. And it's rather efficient :)

It's a common pattern taken from pthreads, but the go implementation is a tiny bit simpler. Whenver I read the pthreads man-page about this I always get very confused, because it talks about threads waking up. But the golang docs are much simpler, unfortunately they could have a bit more examples.

walac: Can't you make this simpler using [sync/atomics](https://golang.org/pkg/sync/atomic/) package?

jonasfj: @walac, nope, If I used a atomic int I wouldn't be able to make something that waits for the int to reach zero..
Well, I could make a spin lock...
```go
for atomics.LoadInt32(&s.refCount) != 0 {
  time.Sleep(100 * time.Milisecond)
}
```

But let's just agree that's an extremely inefficient implementation :)

That's the beauty of this, I can wait for the refCount to reach zero... Note: similar thing might be possible using `sync.WaitGroup`, but the semantics around this when incrementing and decrementing aren't clearly documented.
And I can't ever check if it's zero, without waiting... I guess a generic semaphore implementation might have been nice for this.

PR #94, opened 2016-08-14
Use schema types

jonasfj:
What are your thoughts on replace `CompositeSchema` with `schematypes`, see:
https://github.com/taskcluster/go-schematypes

The idea is that we could actually construct the schema programmatically... Any type can be `Map`'ed to interface{}... So that way we can have nested plugin architectures for things like volumes plugins, where this is needed.

There is more work to be done in schematypes.. But this is a decent illustration of what it might look like.
And this is a huge architectural change so I would like you guys to think about it...


walac:
I loved this, I am much in favor on solving problems through language libraries rather than external tools.

engines/engine.go
```diff
@@ -127,16 +128,10 @@ type Capabilities struct {
...
-}
-
-// ConfigSchema returns an empty jsonschema indicating that no custom config is
-// required.
-func (EngineBase) ConfigSchema() runtime.CompositeSchema {
-	return runtime.NewEmptyCompositeSchema()
+// PayloadSchema returns an empty schematypes.Object indicating no contraints
+// on keys of the payload object.
+func (EngineBase) PayloadSchema() schematypes.Object {
+	return schematypes.Object{}
```
walac: Will `ConfigSchema` and `PayloadSchema` be the same thing? If so, `PayloadSchema` seems to be misnomer.

engines/enginetest/testutils.go
```diff
@@ -78,22 +77,22 @@ func (p *EngineProvider) ensureEngine() {
...
 	}
 
-	jsonConfig := map[string]json.RawMessage{}
+	var jsonConfig interface{}
 	err := json.Unmarshal([]byte(p.Config), &jsonConfig)
 	nilOrPanic(err, "Config parsing failed: ", p.Config)
-	config, err := engineProvider.ConfigSchema().Parse(jsonConfig)
-	nilOrPanic(err, "Config validation failed: ", p.Config)
+	err = engineProvider.ConfigSchema().Validate(jsonConfig)
+	nilOrPanic(err, "Config validation failed: ", p.Config, "\nError: ", err)
```
walac: Isn't better to change `nilOrPanic` to print the `error` when it isn't `nil`?

engines/osxnative/engine.go
```diff
@@ -19,36 +17,57 @@ type engine struct {
...
 }
 
-func (e engine) PayloadSchema() runtime.CompositeSchema {
+var payloadSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"link": schematypes.URI{
+			MetaData: schematypes.MetaData{
+				Title: "Executable to download",
+				Description: `Link to an script/executable to run. The file must still be
+		      explicitly be referenced by the command field.`,
```
walac: nit: remove the second "be"

engines/mock/payloadschema.go
```diff
@@ -0,0 +1,52 @@
...
+				"write-error-log",
+				"write-files",
+				"print-env-var",
+			},
+		},
+		"argument": schematypes.String{
+			MetaData: schematypes.MetaData{
+				Title: "Argument to be given to function",
+				Description: `This argument will be passed to function, notice that
+				not all functions takes an argument and may just choose to ignore it.`,
```
walac: nit: s/takes/take

engines/qemu/engine.go
```diff
@@ -5,34 +5,71 @@ package qemuengine
...
-	// Cast config to engineConfig
-	c, ok := options.Config.(*engineConfig)
-	if !ok {
+var configSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"maxConcurrency": schematypes.Integer{
+			MetaData: schematypes.MetaData{
+				Title: "Image Folder",
+				Description: `Path to folder to be used for image storage and cache.
+											Please ensure this has lots of space.`,
```
walac: ops, I think this description is wrong.

plugins/artifacts/artifacts.go
```diff
@@ -9,44 +7,49 @@ import (
...
-		payload:        *(options.Payload.(*payload)),
+		artifacts:      p.Artifacts,
 	}, nil
 }
 
 type taskPlugin struct {
 	plugins.TaskPluginBase
-	context *runtime.TaskContext
-	payload payload
+	context   *runtime.TaskContext
```
walac: nit: extra space

plugins/plugin.go
```diff
@@ -27,12 +28,13 @@ type TaskPluginOptions struct {
 //
 // All methods on this interface must be thread-safe.
 type Plugin interface {
-	// PayloadSchema returns the CompositeSchema that represents the payload.
+	// PayloadSchema returns the a schematypes.Object with the properties for
```
walac: nit s/the//

plugins/plugin.go
```diff
@@ -27,12 +28,13 @@ type TaskPluginOptions struct {
...
+	// for the TaskPluginOptions.Payload property.
 	//
-	// The Payload property on TaskPluginOptions given to NewTaskPlugin will be
-	// the result from CompositeSchema.Parse() on the CompositeSchema returned
-	// from this method.
-	PayloadSchema() (runtime.CompositeSchema, error)
+	// Note: this will be merged with payload schemas from engine and other
+	// plugins, thus, it cannot contain conflicting properties. Furthermore the
+	// metadata will be discarded and additionalProperties are not allowed.
```
walac: nit: s/are/will

plugins/plugin.go
```diff
@@ -146,15 +148,10 @@ type TaskPlugin interface {
 // new optional methods.
 type PluginBase struct{}
 
-// PayloadSchema returns an empty composite schema for plugins that doesn't
+// PayloadSchema returns schema for an empty object for plugins that doesn't
```
walac: nit: "a schema"

plugins/pluginmanager.go
```diff
@@ -57,70 +59,167 @@ func waitForErrors(errors <-chan error, count int) error {
...
+			Description: `Mapping from plugin name to plugin configuration.
+                    The 'disabled' key is special and lists plugins that are
+                    disabled. Plugins that are disabled do not require
+                    configuration.`,
+		},
+		Properties: schematypes.Properties{
+			"disabled": schematypes.Array{
+				MetaData: schematypes.MetaData{
+					Title: "Disabled Plugins",
+					Description: `List of disabled plugins. If a plugin then its
```
walac: If a plugin ?? then...

engines/engine.go
jonasfj: Nope...`plugin.ConfigSchema` is not a thing `pluginProvider.ConfigSchema` is the schema for configuration required when creating a plugin...

petemoore:
I think it could be a backwards step to go from having real json schema, to a go representation of a subset of jsonschema:

* We lose the generality of json schema
* We lose the auto generated types (and therefore the guarantee that types match schemas)
* It becomes a programming effort to change schema validation, rather than editing of a plain schema file

I also recognise the CompositeSchema design was poor and didn't provide a mechanism to "connect" schema excerpts together.

However, an alternative approach would have been to use the model provided by [jsonschema2go representation of json schema](https://github.com/taskcluster/jsonschema2go/blob/91b27eef252d8a07cf326374a4932be9df3b6be7/jsonschema.go#L28-L69) and build json schemas with the results of building objects from yaml source files.

It might have better to have this conversation earlier on, before such a major rewrite was undertaken, as maybe I could have pointed you to this design first, and it surely would have been a shorter path, and perhaps result in a superior solution requiring less code.

I do also see benefits to your approach. Probably overall it is simpler, and easier for a contributor to reason about without requiring a lot of mental investment. However, I think the better solution would have been to throw away CompositeSchema, which I think was the cause of the problems, rather than the go type generation.

I don't want to block on solving the CompositeSchema problems though, so if this is needed to get the worker on its feet, and we don't want to redesign it, so be it.

engines/enginetest/testutils.go
jonasfj: I think it might also do that :)

plugins/artifacts/artifacts.go
jonasfj: alignment by gofmt...

jonasfj:
@petemoore,
I'm starting to see loss of code-generation as a benefit... This isn't complicated, but code-gen is complicated. Perhaps we should even stop using `go-import-subtree` it really doesn't save us much...
Manually important all the sub-packages in the top-level package isn't error-prone or hard to do.. And we are just as likely to do it wrong as we are to forget to run `go generate`.

Notably, I also moved away from go-extpoints in this PR, it's simply too complicated...

@petemoore, If you take a look at the changes in plugin manager and worker package, you'll start to see where the power of this schema construction comes in...

-----
@walac, FYI; I haven't actually had the worker running after this major change, I'm sure that there are small configuration issues to fix... Error messages to improve. But if you guys support the direction this is taking I suggest that we start merging all of the out standing PRs and iterate on this stuff... I don't like having too many major PRs like this blocking everything, it might be better to have a minor breakage.

FYI; this is broken on travis as I haven't updated vendor/ yet, thinking I'll do that when i land it... It's aweful to review with so much noise...

worker/config.go
```diff
@@ -0,0 +1,180 @@
...
+			"engines": schematypes.Object{
+				MetaData: schematypes.MetaData{
+					Title: "Engine Configuration",
+					Description: `Mapping from engine name to engine configuration.
+						Even-though the worker will only use one engine at any given time,
+						the configuration file can hold configuration for all engines.
+						Hence, you need only update the 'engine' key to change which engine
+						should be used.`,
+				},
+				Properties: engineConfig,
```
jonasfj: Now isn't this cool... notice that in the type above `Engines` has the type `map[string]interface{}`, so it works even though the engines may specify whatever they want as config --- I can still unmap it into something well-typed.

walac:
> We lose the generality of json schema

I think what matters in this case is the user interface, aka yaml files, and it is just a custom yaml, I don't see it more general than go types.

> We lose the auto generated types (and therefore the guarantee that types match schemas)

The go type system should guarantee that, shouldn't it?

> It becomes a programming effort to change schema validation, rather than editing of a plain schema file

The syntax is as declarative as yaml, although I agree yaml is a bit cleaner. Anyway, any change in yaml would lead to a programming effort, since we need to implement the semantics of this yaml change.

jonasfj:
Okay, I'll make tests pass to day or tomorrow and land this...

PR #95, opened 2016-08-15
Update engine tests for new enginetest interface.

walac:
EngineProvider field is now a pointer instead of a object.


jonasfj:
# `r+`

I did this for test efficiency... since QEMU engine has a lot of setup/teardown that was slow...
Thanks for catching this...

PR #96, opened 2016-08-15
Include payload in validation error message. r=jonasfj

walac:
Dumping payload to logs when validation fails helps to diagnose what's
wrong with payload layout.


jonasfj:
# `r+`

PR #97, opened 2016-08-31
Travis to use go 1.7 instead of go 1.6

petemoore:
That should do it.


petemoore:
Opening a separate PR (#98) that also takes care of enforcing that compiler should be 1.7 or higher in the make file.

PR #98, opened 2016-08-31
Require go 1.7 compiler (or higher) to build

petemoore:
A big thank you ❤️ to @jhford for pointing out how to implement this in Make.


petemoore:
I've squashed into a single commit.

The changes to `Makefile` are to ensure that go compiler 1.7 or later is used. Note, you can ensure a file is compiled only with go 1.7 onwards with the build constraint [go1.7](https://golang.org/pkg/go/build/#hdr-Build_Constraints) - however, if you compile with an earlier version of go, the effect would be that that source file would just not be included. So putting this in the `Makefile` means you get an explanation, rather than just getting missing source files missing from the build.

The changes to `.travis.yml` are to make sure travis builds with exactly go1.7 (note: not go1.7 or later).

The changes to `vendor` are updates to `golint` and its dependencies in order for `golint` to function correctly with go1.7. The previous frozen version exists from a time before go1.7. See golang/lint#223.

Makefile
```diff
@@ -3,10 +3,24 @@
...
 
+# Ensure bash shell! Needed for checking go version...
+SHELL := /bin/bash
+
 # Ensure go-extpoints and go-import-subtree are available for go generate
 export PATH := $(GOPATH)/bin:$(PATH)
 
-all: build
+# For checking go compiler has a suitable version number
+GO_VERSION := $(shell go version 2>/dev/null | cut -f3 -d' ')
```
walac: I would go with `awk` instead of `cut` here because of its relaxed spacing matching pattern.

Makefile
```diff
@@ -3,10 +3,24 @@
...
+GO_MAJ := $(shell echo "$(GO_VERSION)" | cut -f1 -d'.')
+GO_MIN := $(shell echo "$(GO_VERSION)" | cut -f2 -d'.')
+
+all: rebuild
+
+prechecks:
+	@test -n "$(GO_VERSION)" || (echo "Could not find go compiler, 'go version' returned no output" && false)
+	@test "$(GO_MAJ)" == "go1" || (echo "Require go version 1.x, where x >= 7; however found '$(GO_VERSION)'" && false)
+	@test "$(GO_MIN)" -ge 7 || (echo "Require go version 1.x, where x>=7; however found '$(GO_VERSION)'" && false)
```
walac:
```bash
@test "0$(GO_MIN)" -ge 7
```
* notice the `0` before $(GO_MIN)


This way, if something goes wrong with `cut` and `GO_MIN` ends up empty, it still produces a valid expression.

walac:
I have mixed feelings about this. One thing is not care about go < 1.7, another one is prohibiting potential contributors to use 1.6, since they may rely on system package manager which didn't update to newest go yet.

Makefile
petemoore: Is there a problem it would solve, that this version doesn't? Trying to understand if it is purely personal preference, or if it affects behaviour.

Makefile
petemoore: Nice spot!

jhford: both `"0$(GO_MIN)"` and `"$(GO_MIN)"` will result in a falsy value for boolean logic:
```
jhford:~/taskcluster/sqs-simple $ test "" -gt 7
bash: test: : integer expression expected
jhford:~/taskcluster/sqs-simple $ echo $?
2
```
and adding the 0 doesn't account for the case that `GO_MIN` is assigned to some non-number, non-empty string value.
```
jhford:~/taskcluster/sqs-simple $ test "0john" -gt 7
bash: test: 0john: integer expression expected
jhford:~/taskcluster/sqs-simple $ echo $?
2
```
I think it reads cleaner without the 0, but no strong preference.

petemoore:
> I have mixed feelings about this. One thing is not care about go < 1.7, another one is prohibiting potential contributors to use 1.6, since they may rely on system package manager which didn't update to newest go yet.

It is relatively straightforward to install go 1.7 alongside any system package, as we do in `.travis.yml`. I agree though that we don't want to make lives harder for contributors. One advantage of forcing go1.7 is that we get expected results. Another issue is that we vendor dependencies, and these will soon be vendored to versions which might only work with 1.7 and we won't know (unless we also set up tests in travis for running against 1.6 etc). My feeling is we might be making more work for ourselves by supporting older go versions, without getting a lot of benefit for doing so. But I do see your point, so I'm also not sure.

@jonasfj what do you think?

gregarndt:
I think that since this is still very much in a developmental stage that upgrading frequently will not be an issue.  After that point we probably want to have a strategy for when we require certain version of the compiler and packages, so that we don't have various projects using all sorts of versions of Go (similar to to the mess we have with node right now 0.12, 4, 5, and 6).

PR #99, opened 2016-09-01
Enforces correct artifact name. r=jonasfj

walac:
osxnative engine creates an on-the-fly user to run its tasks. When the
artifact path is relative, the full path for the artifact is relative
to this temporary user home directory. As the artifact full path may not
have an intersection with artifact name, a string replacement may yield
a invalid artifact name.

This patch uses a concatenation of the artifact directory with artifact
file to generate the full artifact name.

### DIFF
--- a/plugins/artifacts/artifacts.go
+++ b/plugins/artifacts/artifacts.go
@@ -6,7 +6,6 @@ package artifacts
 import (
 	"mime"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/taskcluster/taskcluster-client-go/tcclient"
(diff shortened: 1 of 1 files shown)


jonasfj:
I suspect this is the wrong fix... FileHandler should be called with path being the path of the file...

If you in the engine prefix all paths, then you must remove that prefix before returning it..

walac:
The file path is unchanged, only the artifact name was patched.

jonasfj:
@walac, I think we should clarify the semantics of `path` in:
https://github.com/taskcluster/taskcluster-worker/blob/master/engines/resultset.go#L13

Perhaps it would better to say that `path` should be relative to the path we're listing under...
I'm also tempted to suggest the path shoudn't start with a slash, and use slashes as separator (regardless of underlying platform).

Without being specific about what the value of `path` should be, we can easily end up doing all sorts of weird fixes. Ideally, we can even have enginetest ensure that this is done correctly.

walac:
If we are going to imposing anything to `path`, I would rather say it must be the complete artifact file path. It is much more clear IMO and easier for engines to deal with.

We should discuss this in a separate bug, as it is unrelated to this PR. This PR fixes a completely different bug: assuming the path and artifact name have the same prefix, which is wrong. For directories, the file artifact name is "artifact name + file name". And this patch makes it explicitly.

PR #100, opened 2016-09-01
Bug 1298801: Includes the daemon command. r=jonasfj

walac:
This commits introduces the ability for taskcluster-worker to run as a
daemon. Before running it, you need to install it as a daemon, with the
"install" command:

$ taskcluster-worker daemon install --engine macosx --logging-level info

The --engine argument is mandatory for the install command. To start the
daemon, run:

$ taskcluster-worker daemon start

To stop it:

$ taskcluster-worker daemon stop

And to uninstall it:

$ taskcluster-worker daemon remove

Remember taskcluster-worker reads taskcluster credentials from
environment variables. You need to make sure taskcluster-worker daemon
has these environment variables set.


commands/daemonize/daemonize.go
```diff
@@ -0,0 +1,196 @@
...
+
+func init() {
+	cmd_ep.CommandProviders.Register(cmd{}, "daemon")
+}
+
+// service has embedded daemon
+type service struct {
+	daemon.Daemon
+	args map[string]interface{}
+	log  *logrus.Logger
```
jonasfj: Perhaps service should be defined in a separate file.. just a nit..

commands/daemonize/daemonize.go
```diff
@@ -0,0 +1,196 @@
...
+		"workerGroup":   config.WorkerGroup,
+		"provisionerID": config.ProvisionerID,
+	})
+
+	w, err := worker.New(config, engine, runtimeEnvironment, l)
+	if err != nil {
+		return "Could not create worker", err
+	}
+
+	w.Start()
```
jonasfj: you need to handle `SIGTERM` and call `w.Stop()` right?

commands/daemonize/daemonize.go
```diff
@@ -0,0 +1,70 @@
...
+	return "Run as a daemon."
+}
+
+func usage() string {
+	return `Usage:
+  taskcluster-worker daemon <action> [--engine <engine>] [--logging-level <level>]
+
+Options:
+  -l --logging-level <level>  	Set logging at <level>.
+  -e --engine <engine>			Set the engine at <engine>.
```
petemoore: spacing looks a bit off here

commands/daemonize/daemonize.go
```diff
@@ -0,0 +1,70 @@
...
+`
+}
+
+func (cmd) Usage() string {
+	return usage()
+}
```
petemoore: What is the reason to have both `func usage() string` and `func (cmd) Usage() string`?

commands/daemonize/daemonize.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (cmd) Execute(args map[string]interface{}) bool {
+	// Setup logger
```
petemoore: nit: set up ("setup" is a noun)

commands/daemonize/daemonize.go
```diff
@@ -0,0 +1,70 @@
...
+const (
+	// name of the service
+	name        = "taskcluster-worker"
+	description = "Taskcluster worker"
+)
+
+var dependencies = []string{}
+
+func init() {
+	cmd_ep.CommandProviders.Register(cmd{}, "daemon")
```
petemoore: nit: the package name is `daemonize`, the directory is `daemonize`, the file is `daemonize.go` but the command is `daemon`. I think they should all be one or the other for consistency. I see we import a package called `daemon` but we could import it as something else if needed.

commands/daemonize/daemonize.go
petemoore: Can you specify the possible value for `<action>` too?

I think I prefer to list them explicitly in the usage - this way, each action can have its own parameter options (as different actions require different parameters). For example, like [this](https://github.com/taskcluster/generic-worker/blob/a4784866719f50ff9e47b63c0e444839f8dc4e56/main.go#L70-L81).

commands/daemonize/service.go
```diff
@@ -0,0 +1,143 @@
...
+
+// Manage by daemon commands or run the daemon
+func (svc *service) Manage() (string, error) {
+	action := svc.args["<action>"].(string)
+	// if received any kind of command, do it
+	switch action {
+	case "install":
+		if svc.args["--engine"] == nil {
+			return "Install failed", errors.New("--engine is mandatory for install command")
+		}
```
petemoore: You won't need this check if you specify each action separately in the usage (see my comment above).

commands/daemonize/syslog_mock.go
```diff
@@ -0,0 +1,9 @@
+// +build windows
+
+package daemonize
+
+import "github.com/Sirupsen/logrus"
+
+func setupSyslog(logger *logrus.Logger) error {
+	return nil
+}
```
petemoore: What is the plan for logging on Windows?

petemoore:
> Remember taskcluster-worker reads taskcluster credentials from environment variables. You need to make sure taskcluster-worker daemon has these environment variables set.

I think we should include that in the help text.

Can you confirm this is only for `taskcluster-worker daemon start` and e.g. not for `taskcluster-worker daemon start` ?

Ultimately I guess this will be resolved when we have proper config loading in place...

commands/daemonize/daemonize.go
walac: it is used in service.go file when an invalid command is specified, although we can `panic`

commands/daemonize/syslog_mock.go
walac: No plan for windows log for now, since we can't run the worker as a service.

walac:
We need it on `taskcluster-worker daemon run`

PR #101, opened 2016-09-08
Clarification of path in FileHandler

jonasfj:
We need to fix engine tests and mock to do this... as well as all the other engines, but I think it's the right direction...


jonasfj:
let's also add to `ExtractFolder` that this is recursive...

walac:
r+

PR #102, opened 2016-09-09
Honor artifacts relative paths.

walac:
From PR #101, the path passed to the file handler must be relative to
the artifact path.

If did the math correctly, this breaks qemu engine.

### DIFF
--- a/engines/osxnative/resultset.go
+++ b/engines/osxnative/resultset.go
@@ -3,13 +3,14 @@
 package osxnative
 
 import (
-	"github.com/taskcluster/taskcluster-worker/engines"
-	"github.com/taskcluster/taskcluster-worker/runtime"
-	"github.com/taskcluster/taskcluster-worker/runtime/ioext"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"regexp"
+
+	"github.com/taskcluster/taskcluster-worker/engines"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+	"github.com/taskcluster/taskcluster-worker/runtime/ioext"
 )
 
 type resultset struct {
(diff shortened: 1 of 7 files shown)


engines/osxnative/resultset.go
```diff
@@ -90,6 +91,12 @@ func (r resultset) ExtractFolder(path string, handler engines.FileHandler) error
 				return engines.ErrResourceNotFound
 			}
 
+			p, err = filepath.Rel(path, p)
+			if err != nil {
+				r.context.LogError(err)
+				return engines.ErrResourceNotFound
```
jonasfj: Isn't this like an internal error of some kind? Ie. should we not crash the application? I'm not sure here...

jonasfj: Hmm... maybe this is okay...

plugins/artifacts/artifacts.go
```diff
@@ -116,9 +116,9 @@ func (tp taskPlugin) errorHandled(name string, expires tcclient.Time, err error)
 	return false
 }
 
-func (tp taskPlugin) createUploadHandler(name, prefix string, expires tcclient.Time) func(string, ioext.ReadSeekCloser) error {
+func (tp taskPlugin) createUploadHandler(name string, expires tcclient.Time) func(string, ioext.ReadSeekCloser) error {
 	return func(path string, stream ioext.ReadSeekCloser) error {
-		return tp.attemptUpload(stream, path, filepath.Join(name, filepath.Base(path)), expires)
+		return tp.attemptUpload(stream, path, filepath.Join(name, path), expires)
```
jonasfj: use `path` not `filepath` which is platform specific.. .This should always use slashes...

jonasfj:
Looks decent, not sure about the error handling... but definitely sure we need to use slash based paths in the artifact plugin or it won't be platform independent.

walac:
If we are not 100% sure, I rather prefer nicely return an error than `panic`.

walac:
Fixed issue + osx engine tests fixes.

PR #103, opened 2016-09-09
Fixed a few shadowned err variables -- and a rare condition in test -…

jonasfj:
…- we should wait for reclaim request that are in the air


walac:
Closing this as it dups #104

PR #104, opened 2016-09-09
Fix gometalinter

jonasfj:
After this we should run `gometalinter --vendor --disable-all --enable=misspell --tests  ./...` on travis...

I'm not sure how add gometalinter, as it has a lot of dependencies, it might not be appropriate to vendor it...

### DIFF
--- a/engines/osxnative/resultset_test.go
+++ b/engines/osxnative/resultset_test.go
@@ -106,11 +106,11 @@ func TestExtractFolder(t *testing.T) {
 
 	err = r.ExtractFolder("test-data", func(p string, stream ioext.ReadSeekCloser) error {
 		expected := path.Base(p) + "\n"
-		data, err := ioutil.ReadAll(stream)
+		data, err2 := ioutil.ReadAll(stream)
 		sdata := string(data)
 
-		if err != nil {
-			return err
+		if err2 != nil {
+			return err2
 		}
 
 		if sdata != expected {
--- a/engines/osxnative/sandbox.go
+++ b/engines/osxnative/sandbox.go
@@ -162,19 +162,19 @@ func (s *sandbox) WaitForResult() (engines.ResultSet, error) {
 			}
 		}()
 
-		userInfo, err := osuser.Lookup(u.name)
-		if err != nil {
+		userInfo, err2 := osuser.Lookup(u.name)
+		if err2 != nil {
 			s.context.LogError("Error looking up for user \""+u.name+"\": ", err, "\n")
 		} else {
-			uid, err := strconv.ParseUint(userInfo.Uid, 10, 32)
-			if err != nil {
-				s.context.LogError("ParseUint failed to convert ", userInfo.Uid, ": ", err, "\n")
+			uid, err2 := strconv.ParseUint(userInfo.Uid, 10, 32)
+			if err2 != nil {
+				s.context.LogError("ParseUint failed to convert ", userInfo.Uid, ": ", err2, "\n")
 				return nil, engines.ErrNonFatalInternalError
 			}
 
-			gid, err := strconv.ParseUint(userInfo.Gid, 10, 32)
-			if err != nil {
-				s.context.LogError("ParseUint failed to convert ", userInfo.Gid, ": ", err, "\n")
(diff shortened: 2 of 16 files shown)


jonasfj:
@petemoore, I'm not sure we should activate all options from gometalinter.
I'm currently using: `gometalinter --vendor, --disable-all, --enable=vet, --enable=misspell, --enable=vetshadow, --enable=golint, --enable=ineffassign, --enable=goconst, --tests `
In my IDE...

But we could probably run more of these... but some of them like: `dupl` which reports duplicated code has both pros and cons, sometimes the simplest way to write tests is to duplicate code... and just tweak the duplicated code a bit... This is not pretty but solves the problem, and IMO test don't have to be high quality.

Anyways, we can enable these one at the time. Obviously the ones that don't have too many errors right now are good candidates for being enabled :)

engines/osxnative/resultset_test.go
```diff
@@ -106,11 +106,11 @@ func TestExtractFolder(t *testing.T) {
...
 	err = r.ExtractFolder("test-data", func(p string, stream ioext.ReadSeekCloser) error {
 		expected := path.Base(p) + "\n"
-		data, err := ioutil.ReadAll(stream)
+		data, err2 := ioutil.ReadAll(stream)
 		sdata := string(data)
 
-		if err != nil {
-			return err
+		if err2 != nil {
+			return err2
```
walac: You don't need to rename the variable in this case, it is expected to be a different `err` that will be returned to the outer function.

engines/osxnative/sandbox.go
```diff
@@ -162,19 +162,19 @@ func (s *sandbox) WaitForResult() (engines.ResultSet, error) {
...
+				s.context.LogError("ParseUint failed to convert ", userInfo.Uid, ": ", err2, "\n")
 				return nil, engines.ErrNonFatalInternalError
 			}
 
-			gid, err := strconv.ParseUint(userInfo.Gid, 10, 32)
-			if err != nil {
-				s.context.LogError("ParseUint failed to convert ", userInfo.Gid, ": ", err, "\n")
+			gid, err2 := strconv.ParseUint(userInfo.Gid, 10, 32)
+			if err2 != nil {
+				s.context.LogError("ParseUint failed to convert ", userInfo.Gid, ": ", err2, "\n")
```
walac: deferred function depends on setting `err` to check if it needs to cleanup. It is better to declare the variables explicitly and use `=` operator for assignment.

engines/osxnative/sandbox.go
```diff
@@ -196,17 +196,17 @@ func (s *sandbox) WaitForResult() (engines.ResultSet, error) {
...
+			s.context.LogError(err2)
 			return nil, engines.ErrNonFatalInternalError
 		}
 
 		defer os.Remove(filename)
 
-		if err = os.Chmod(filename, 0777); err != nil {
-			s.context.LogError(err, "\n")
+		if err2 = os.Chmod(filename, 0777); err2 != nil {
+			s.context.LogError(err2, "\n")
```
walac: Same as previous comment.

engines/osxnative/resultset_test.go
jonasfj: Yes, but if I don't rename the variable in `data, err2 := ioutil.ReadAll(stream)` line 109 (where the rename is) then go-metalinter will complain.
The reason is that `err` will overshadow `err` from outer scope.

It's a decent warning to have as we could easily make a mistake by assuming the inner err is the same as outer err. We can also disable the overshadowing rule, but I suspect it'll catch a lot of small bugs for us...

engines/osxnative/sandbox.go
jonasfj: Hmm... Fixed it, but this seems like a rather fragile control flow... We should probably rethink that.

Perhaps set a boolean `cleanup = true` or something... to trigger the deferred function.. Or even better not use defer like this.. -- I'm not immediately sure how to structure this an elegant manner.

engines/osxnative/resultset_test.go
walac: Good then.

PR #105, opened 2016-09-12
Improve work

jonasfj:

PR #106, opened 2016-09-13
Fix gometalinter

jonasfj:


jonasfj:
carries with r+ from #105

PR #112, opened 2016-09-20
Run gc before each task

jonasfj:
To do this I had to update godep... which was horribly broken for some reason.. .Maybe I've messed up my go install... Anyways, I just deleted `vendor/`, updated all package and revendored everything.

This included an upgrade to tcclient, everything else was smooth...

Note: this is based on interactive branch... The last commit is only vendoring stuff, so feel free to largely ignore that... Mostly the only interesting part here is in the garbagecollector and how run it before each task.


runtime/gc/gc.go
```diff
@@ -61,11 +93,59 @@ func (gc *GarbageCollector) Unregister(resource Disposable) bool {
...
+		return err
+	dispose:
+		err = r.Dispose()
+		if err != nil {
+			if err != ErrDisposableInUse {
+				goto abort
+			}
+			goto keep
+		}
+	}
```
walac: I am not a *don't goto* guy, but we got a little spaghetti here. With the creation of a helper function containing the code of `dispose` and `keep` branches, we remove most of the `goto`s. `abort` branch is fine, though.

runtime/gc/gc.go
```diff
@@ -89,3 +169,31 @@ func (gc *GarbageCollector) CollectAll() error {
...
 }
+
+// needDiskSpace returns true if we need to free diskspace
+func (gc *GarbageCollector) needDiskSpace() bool {
+	// If we have no metrics or minimum diskspace we remove everything
+	if gc.minimumDiskSpace == 0 || gc.storageFolder == "" {
+		return true
+	}
+	stat, err := disk.Usage(gc.storageFolder)
+	if err != nil {
```
walac: nit: log a warning here

runtime/gc/gc.go
```diff
@@ -89,3 +169,31 @@ func (gc *GarbageCollector) CollectAll() error {
...
+}
+
+// needMemory returns true if we need to free memory
+func (gc *GarbageCollector) needMemory() bool {
+	// If we have no metrics or minimum memory we remove everything
+	if gc.minimumMemory == 0 {
+		return true
+	}
+	stat, err := mem.VirtualMemory()
+	if err != nil {
```
walac: nit: log a warning here

runtime/gc/gc_test.go
```diff
@@ -82,3 +87,108 @@ func TestGarbageCollector(t *testing.T) {
...
+}
+
+func TestCollectDiskOnlyInUse(t *testing.T) {
+	gc := &GarbageCollector{
+		storageFolder:    "...",
+		minimumDiskSpace: math.MaxInt64,
+		minimumMemory:    1,
+	}
+
+	// Add two resources only r1 should be disposed
```
walac: This test doesn't expect any resource be disposed, does it?

plugins/artifacts/artifacts.go
```diff
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	schematypes "github.com/taskcluster/go-schematypes"
-	"github.com/taskcluster/taskcluster-client-go/tcclient"
+	"github.com/taskcluster/taskcluster-client-go"
```
jonasfj: `tcclient` still I think... we previously had a really old version of taskcluster-client in `vendor/`...

runtime/gc/gc.go
jonasfj: hmm.. yeah probably, we don't have a log context, I'll leave a todo and do it later... I have too many PRs built ontop of this to do major refactoring..

runtime/gc/gc_test.go
jonasfj: true, fixed comments...

runtime/gc/gc.go
jonasfj: Yeah, I think this is something we'll refactor later... We should possibly use the estimate resource sizes to decide how many we release... Things like that...

jonasfj:
Merge other PRs into this... Didn't make it any prettier. But I was loosing track, having too many stacked was starting to scare me..

PR #115, opened 2016-09-22
Added configuration options for max memory per vm

jonasfj:
This added an engine option to QEMU engine so we can configure max allowed memory per VM.

If  a machine is specified to require more memory than is allowed, it'll result in an malformed-payload error.

### DIFF
--- a/engines/qemu/vm/machine.go
+++ b/engines/qemu/vm/machine.go
@@ -15,6 +15,7 @@ import (
 // This only allows certain arguments to be specified.
 type Machine struct {
 	UUID    string `json:"uuid"`
+	Memory  int    `json:"memory,omitempty"`
 	Network struct {
 		Device string `json:"device"`
 		MAC    string `json:"mac"`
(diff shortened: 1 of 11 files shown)


engines/qemu/vm/machine.go
```diff
@@ -41,6 +42,10 @@ const machineSchemaString = `{
 			"pattern":								"^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$",
 			"description":						"System UUID for the virtual machine"
 		},
+		"memory": {
+			"type":										"integer",
```
walac: is this some indent issue or a github thing?

engines/qemu/vm/options.go
```diff
@@ -0,0 +1,35 @@
...
+                  these limitations will be resolved 'malformed-payload'.`,
+	},
+	Properties: schematypes.Properties{
+		"maxMemory": schematypes.Integer{
+			MetaData: schematypes.MetaData{
+				Title: "Max Memory",
+				Description: `Maximum allowed virtual machine memory in MiB. This is
+										 	also the default memory if the machine image doesn't
+											specify memory requirements. If the machine specifies a
+											memory higher than this, the task will fail to run.`,
```
walac: nit: indent problems here

engines/qemu/vm/vm.go
```diff
@@ -42,11 +43,25 @@ type VirtualMachine struct {
 	domain       *qemu.Domain
 }
 
-// NewVirtualMachine constructs a new virtual machine.
+// NewVirtualMachine constructs a new virtual machine using the given
+// machineOptions, image, network and cdroms.
+//
+// Returns engines.MalformedPayloadError, if machineOptions and image definition
```
walac: nit s/,//

engines/qemu/vm/machine.go
jonasfj: oh, my go editor uses tabs in strings and this happens... :)

PR #116, opened 2016-09-23
Don't add prefix message when dumping command output.

walac:
When logging out the output of a command run in a context of a task, we
should not add any prefix string, we this will make treeherder parsing
to fail.

### DIFF
--- a/engines/osxnative/sandbox.go
+++ b/engines/osxnative/sandbox.go
@@ -52,7 +52,7 @@ type stderrLogWriter struct {
 }
 
 func (w stderrLogWriter) Write(p []byte) (int, error) {
-	w.context.LogError(string(p))
+	w.context.LogRaw(string(p))
 	return len(p), nil
 }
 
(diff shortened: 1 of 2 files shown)


engines/osxnative/sandbox.go
```diff
@@ -52,7 +52,7 @@ type stderrLogWriter struct {
 }
 
 func (w stderrLogWriter) Write(p []byte) (int, error) {
-	w.context.LogError(string(p))
+	w.context.LogRaw(string(p))
```
jonasfj: hmm... Maybe I should kick myself for not seeing this earlier... but you want a `io.Writer` interface.. you don't need to wrap `TaskContext` the `LogError` and `Log` methods are really only for non-raw messages...

The raw messages should go to `TaskContext.LogDrain()` which returns an `io.Writer` that is buffered and all..

jonasfj:
Note: At some point we should look into ensure that output is line buffered... either adding a wrapper to `TaskContext.LogDrain()` ensuring this... Or check that all engines does line buffered output... Otherwise, the output could easily get mangled.

Anyways, just use `TaskContext.LogDrain()`... risk of mangling is low, and we'll fix it when we get there.

PR #118, opened 2016-09-25
Fix newlines for osxnative engine logs.

walac:


engines/osxnative/sandbox.go
```diff
@@ -144,10 +126,10 @@ func (s *sandbox) WaitForResult() (engines.ResultSet, error) {
 	// run successfully.
 	u := user{}
 	if err = u.create(); err != nil {
-		s.context.LogError("Could not create user: ", err, "\n")
+		s.context.LogError("Could not create user: ", err)
```
jonasfj: Doesn't `context.LogError` prefix with `[taskcluster:error]` ? Which seems to imply to me that it's intended to print lines... hence, how about we move the `\n` into `context.LogError` ?

jonasfj: Ignore more... I read htis backwards. r+

PR #119, opened 2016-09-27
Config loading

jonasfj:
Sorry, I had to touch `vendor/` again.. but it's not so bad as last time :)

Things worth looking at is `config/`.


config/provider.go
```diff
@@ -0,0 +1,46 @@
...
+// Providers returns a map of the registered TransformationProvider.
+func Providers() map[string]TransformationProvider {
+	mProviders.Lock()
+	defer mProviders.Unlock()
+
+	// Clone providers
+	m := map[string]TransformationProvider{}
+	for name, provider := range providers {
+		m[name] = provider
+	}
```
walac: Why do we need to return a copy?

config/secrets/transform.go
```diff
@@ -0,0 +1,94 @@
...
+func injectSecrets(
+	s *secrets.Secrets,
+	cache map[string]map[string]interface{},
+	val interface{},
+) (interface{}, error) {
+	switch val := val.(type) {
+	case []interface{}:
+		for i, v := range val {
+			v, err := injectSecrets(s, cache, v)
+			val[i] = v
```
walac: Shouldn't we only assign the value if there is no error?

config/secrets/transform_test.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+		Result: map[string]interface{}{
+			"key": "hello-world",
+			"credentials": map[string]interface{}{
+				"clientId":    "no-client",
+				"accessToken": "no-secret",
+			},
+			"secretsBaseUrl": s.URL,
+		},
+	}.Test()
```
walac: We need a test case for array of interfaces, doesn't we?

config/configtest/transform.go
```diff
@@ -0,0 +1,35 @@
...
+		panic(fmt.Sprintf("Unknown transform: %s", c.Transform))
+	}
+
+	err := transform.Transform(c.Input)
+	if err != nil {
+		panic(fmt.Sprintf("Transform(Input) failed, error: %s", err))
+	}
+
+	if !reflect.DeepEqual(c.Input, c.Result) {
+		panic(fmt.Sprintf("c.Result != c.Input, after transform, result: %#v ", c.Input))
```
jonasfj: Probably, but I'm too lazy to carry around `testing.T`... It's just tests who cares about code quality.

config/env/transform_test.go
```diff
@@ -0,0 +1,21 @@
...
+	os.Setenv("CFG_TEST_ENV_VAR", "hello-world")
+	configtest.Case{
+		Transform: "env",
+		Input: map[string]interface{}{
+			"key": map[string]interface{}{"$env": "CFG_TEST_ENV_VAR"},
+		},
+		Result: map[string]interface{}{
+			"key": "hello-world",
+		},
+	}.Test()
```
jonasfj: nope, good catch

config/provider.go
jonasfj: To avoid concurrency issues

config/secrets/transform.go
jonasfj: Sure, it's not important... I think modification of the input is undefined when there is an error.

config/secrets/transform_test.go
jonasfj: true

PR #120, opened 2016-09-27
Bug 1305707: gzip compress live log artifact.

walac:
Treeherder log parser expects log file be gzipped compressed. This also
saves bandwidth resources.

### DIFF
--- a/plugins/livelog/livelog.go
+++ b/plugins/livelog/livelog.go
@@ -4,7 +4,9 @@ package livelog
 
 import (
 	"bufio"
+	"compress/gzip"
 	"fmt"
+	"io"
 	"net/http"
 	"time"
 
@@ -20,19 +22,24 @@ type pluginProvider struct {
 
 type plugin struct {
 	plugins.PluginBase
-	log *logrus.Entry
+	log         *logrus.Entry
+	environment *runtime.Environment
 }
 
 type taskPlugin struct {
 	plugins.TaskPluginBase
-	context    *runtime.TaskContext
-	url        string
-	expiration tcclient.Time
-	log        *logrus.Entry
+	context     *runtime.TaskContext
+	url         string
+	expiration  tcclient.Time
+	log         *logrus.Entry
+	environment *runtime.Environment
 }
 
 func (pluginProvider) NewPlugin(opts plugins.PluginOptions) (plugins.Plugin, error) {
-	return plugin{log: opts.Log}, nil
+	return plugin{
+		log:         opts.Log,
+		environment: opts.Environment,
+	}, nil
 }
 
 func (p plugin) NewTaskPlugin(opts plugins.TaskPluginOptions) (plugins.TaskPlugin, error) {
@@ -42,6 +49,7 @@ func (p plugin) NewTaskPlugin(opts plugins.TaskPluginOptions) (plugins.TaskPlugi
 			"taskID": opts.TaskInfo.TaskID,
 			"runID":  opts.TaskInfo.RunID,
 		}),
+		environment: p.environment,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


plugins/livelog/livelog.go
```diff
@@ -100,11 +104,46 @@ func (tp *taskPlugin) Finished(success bool) error {
 		return err
 	}
 	defer file.Close()
+
+	tempFile, err := ioutil.TempFile(os.TempDir(), "livelog")
```
jonasfj: Maybe use `runtime.Environment.TemporaryStorage` to get a temporary file in the temporary directory.

We mostly introduced the `TemporaryStorage` abstraction to ensure that files could be moved to something other than `/tmp`.. Notably because `/tmp` is a bad choice for docker images, cache volumes, etc...

So I'm not sure... if we should use it here.

plugins/livelog/livelog.go
```diff
@@ -100,11 +104,46 @@ func (tp *taskPlugin) Finished(success bool) error {
...
+
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err2 := tempFile.Close(); err2 != nil {
+			tp.context.LogError(err2)
```
jonasfj: Don't log this error to the task log!

1) You can't log to task log in Finished() - the log is closed!
2) This is clearly not the tasks fault, it's an internal worker error, the task should not ever get system error messages from the worker. These should go into logrus, for syslog...

plugins/livelog/livelog.go
```diff
@@ -100,11 +103,42 @@ func (tp *taskPlugin) Finished(success bool) error {
 		return err
 	}
 	defer file.Close()
+
+	tempDir, err := runtime.NewTemporaryStorage(os.TempDir())
```
jonasfj: sorry, no...
In `NewPlugin` you get a copy of `runtime.Environment` from the options... Keep that, it contains an instance of temporary storage that you can use.

plugins/livelog/livelog.go
```diff
@@ -100,11 +103,42 @@ func (tp *taskPlugin) Finished(success bool) error {
...
+	zip := gzip.NewWriter(tempFile)
+	if _, err = io.Copy(zip, file); err != nil {
+		return err
+	}
+
+	if err = zip.Close(); err != nil {
+		return err
+	}
+
+	compressedLog, err := os.Open(tempFile.Path())
```
jonasfj: temp file already implements `ioext.ReadWriteSeeker`

It just embeds `*os.File`, see: https://github.com/taskcluster/taskcluster-worker/blob/2619dd09aeb69107e70f5a7dc9d788d8552f6954/runtime/tempfolder.go#L41

plugins/livelog/livelog.go
```diff
@@ -100,11 +103,42 @@ func (tp *taskPlugin) Finished(success bool) error {
...
+	if err != nil {
+		return err
+	}
+
+	tempFile, err := tempDir.NewFile()
+	if err != nil {
+		return err
+	}
+
+	defer tempFile.Close()
```
jonasfj: If you really wanted to log the error, which honestly isn't critical, but if you wanted to... you would use the `Log` from `TaskPluginOptions`, keep a reference to the `options.Log` from `NewTaskPlugin`

plugins/livelog/livelog.go
jonasfj: actually, it's not crazy to embed `runtime.Environment` in your plugin.

plugins/livelog/livelog.go
walac: Not a big deal, actually.

jonasfj: agreed, we should probably not even return an error, and instead teach the tempFile to log errors, because it should always work... anyways that's for another day.

plugins/livelog/livelog.go
```diff
@@ -20,19 +22,24 @@ type pluginProvider struct {
 
 type plugin struct {
 	plugins.PluginBase
-	log *logrus.Entry
+	log         *logrus.Entry
+	environment *runtime.Environment
```
jonasfj: Just a note: I want to pass environment by value in the future. Not now, just a note.

PR #121, opened 2016-09-29
Dockerfile cleanup

jonasfj:
I probably still have more dockerfile work to do before I can deploy with this...

But as illustrated in `qemu-config.yml` I don't need to have any secrets in the config file.
I already put statelessdns secrets into secrets :)


examples/qemu-config.yml
```diff
@@ -1,33 +1,48 @@
...
+  serverPort:         443
+  tlsCertificiate:
+    $secret:  project/taskcluster/taskcluster-worker/stateless-dns
+    key:      certificate
+  tlsKey:
+    $secret:  project/taskcluster/taskcluster-worker/stateless-dns
+    key:      tlsKey
+  statelessDNSSecret:
+    $secret:  project/taskcluster/taskcluster-worker/stateless-dns
+    key:      secret
```
jonasfj: So this is what I wanted to do with the config transforms... Take a look at the secret: `project/taskcluster/taskcluster-worker/stateless-dns` at https://tools.taskcluster.net/secrets/

I haven't tested that everything is there... last I used golang for certificate I had to include the entire chain, but that's just a slight extension of this secret.

tc-worker.Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Install zstd 1.0.0
+RUN apt-get update -y \
+ && apt-get install -y curl build-essential \
+ && curl -L https://github.com/facebook/zstd/archive/v1.0.0.tar.gz > /tmp/zstd.tar.gz \
+ && tar -xvf /tmp/zstd.tar.gz -C /tmp \
+ && make -C /tmp/zstd-1.0.0/programs install \
+ && rm -rf /tmp/zstd-1.0.0/ /tmp/zstd.tar.gz \
+ && apt-get purge -y curl build-essential \
+ && apt-get auto-remove -y \
```
jonasfj: By removing in the same `RUN` step, files aren't present in the layers... This image is actually only 283MB, where as the `tc-worker-env` used for testing is 766MB (it includes build dependencies like gcc, make and golang).

config/secrets/transform.go
```diff
@@ -38,57 +40,32 @@ func (provider) Transform(config map[string]interface{}) error {
...
-		key, ok2 := val["key"].(string)
-		if ok1 && ok2 && len(val) == 2 {
-			// If secret isn't in the cache we try to load it
-			if _, ok := cache[name]; !ok {
-				secret, err := s.Get(name)
-				if err != nil {
-					return nil, err
-				}
-				value := map[string]interface{}{}
-				_ = json.Unmarshal(secret.Secret, &value)
```
petemoore: We should probably `nilOrPanic(...)` rather than ignore errors - what do you think? I think we should definitely have something to recover from panics at the bottom of the stack, which can initiate a worker shutdown. It feels non-deterministic to ignore errors, even if they should never occur... I also don't like bloating code with error handling, so it is definitely a balance.

jonasfj: there is certainly no harm in panicing on error here :)

jonasfj: Actually, I decided to just return an error because that was just as easy...

Note: code has been refactored a bit... to deduplicate some logic.

tc-worker.Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ && apt-get clean -y \
+ && rm -rf /var/lib/apt/lists/ \
+    ;
```
walac: indent nit?

tc-worker.Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# Install runtime dependencies
+RUN apt-get update -y \
+ && apt-get upgrade -y \
+ && apt-get install -y qemu-system-x86 qemu-utils dnsmasq-base iptables \
+ && apt-get clean -y \
+ && rm -rf /var/lib/apt/lists/ \
+    ;
+
+# Install zstd 1.0.0
+RUN apt-get update -y \
```
walac: Do we need to run `apt-get update` twice?

docker-tests.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 
 touch .bash_history
-ARGS="--tty --interactive --rm --privileged -e DEBUG -v `pwd`/.bash_history:/root/.bash_history -v `pwd`:/src taskcluster/tc-worker-env";
+ARGS="--tty --interactive --rm --privileged -e DEBUG";
+ARGS="$ARGS -v `pwd`/.bash_history:/root/.bash_history";
+ARGS="$ARGS -v `pwd`:/go/src/github.com/taskcluster/taskcluster-worker/";
+ARGS="$ARGS taskcluster/tc-worker-env";
```
jonasfj: No, reason... any reason not to?

I'll fix these nits... and land...

PR #122, opened 2016-09-30
Add type support for values loaded from env vars, mainly to support $…

jonasfj:
…PORT

### DIFF
--- a/config/env/transform.go
+++ b/config/env/transform.go
@@ -1,9 +1,23 @@
 // Package configenv implements a TransformationProvider that replaces objects on
 // the form: {$env: "VAR"} with the value of the environment variable VAR.
+//
+// An additional key 'type' may be given as {$env: "VAR", type: "TYPE"}, which
+// determines how the environment variable value should be parsed. By default
+// the TYPE is 'string', but the following types are supported:
+//   - string, value is return as string (default),
+//   - number, value is parsed a number,
+//   - json, value is parsed as JSON,
+//   - bool, value becomes true if it matches 'true' (case insensitive), and,
+//   - list, value is parsed as space separated list of strings.
 package configenv
 
 import (
+	"encoding/json"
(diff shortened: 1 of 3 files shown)


config/env/transform.go
```diff
@@ -16,6 +30,47 @@ func init() {
 
 func (provider) Transform(cfg map[string]interface{}) error {
 	return config.ReplaceObjects(cfg, "env", func(val map[string]interface{}) (interface{}, error) {
-		return os.Getenv(val["$env"].(string)), nil
+		env := val["$env"].(string)
+		value := os.Getenv(env)
+		typ, ok := val["type"]
```
walac: nit: `valueType` is a more descriptive variable name than `typ`.

config/env/transform.go
```diff
@@ -1,9 +1,23 @@
 // Package configenv implements a TransformationProvider that replaces objects on
 // the form: {$env: "VAR"} with the value of the environment variable VAR.
+//
+// An additional key 'type' may be given as {$env: "VAR", type: "TYPE"}, which
+// determines how the environment variable value should be parsed. By default
+// the TYPE is 'string', but the following types are supported:
+//   - string, value is return as string (default),
+//   - number, value is parsed a number,
```
walac: I think we should have different types for scalars and integers.

jonasfj: No, the object hierarchy uses types matching what you get after `var t interface{}; json.Unmarshal(..., &t)`

So all integers will be loaded in as `float64`.. Consider these config transformers as functions that operate on JSON equivalent data structures:
 * `map[string]interface{}`
 * `interface{}`
 * `bool`
 * `float64`
 * `string`
 * `nil`

Maybe I should make some checks for that...

jonasfj: Will add a recursive check to fix this later..

PR #123, opened 2016-10-03
Deployment fixes

jonasfj:
A collection of small bugs I fixed in the process of deploying tc-worker  on packet.

### DIFF
--- a/config/load.go
+++ b/config/load.go
@@ -11,7 +11,7 @@ import (
 )
 
 // Schema returns the configuration file schema
-func Schema() schematypes.Schema {
+func Schema() schematypes.Object {
 	transformations := []string{}
 	for name := range Providers() {
 		transformations = append(transformations, name)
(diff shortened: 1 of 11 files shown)


config/load.go
```diff
@@ -61,10 +61,21 @@ func Load(data []byte) (map[string]interface{}, error) {
 		return nil, fmt.Errorf("Expected 'config' property to be an object")
 	}
 
+	// Ensure that we have a simple JSON compatible structure
+	if err := jsonCompatTypes(result); err != nil {
+		panic(fmt.Sprintf("YAML loaded wrong types, error: %s", err))
+	}
+
 	// Apply transforms
 	if _, ok := c["transforms"]; ok {
```
walac: instead of ignoring the result, shouldn't we save it in a variable, since we will need this in line 72?

jonasfj: I would, but I couldn't come up with a good variable name :)

PR #124, opened 2016-10-05
Add configuration for creation to macosx engine.

walac:
In order to create some kind of isolation among tasks, macosx engine has the ability to create an on-fly user to run tasks. This commit adds configurations to the engine to tell if a new user should be created or not to run a task, as well as the user group.

### DIFF
--- a/engines/osxnative/engine.go
+++ b/engines/osxnative/engine.go
@@ -12,7 +12,8 @@ import (
 
 type engine struct {
 	engines.EngineBase
-	log *logrus.Entry
+	config *configType
+	log    *logrus.Entry
 }
 
 type engineProvider struct {
@@ -20,12 +21,26 @@ type engineProvider struct {
 }
 
 func (e engineProvider) NewEngine(options engines.EngineOptions) (engines.Engine, error) {
-	return engine{log: options.Log}, nil
+	var c configType
+	if err := schematypes.MustMap(configSchema, options.Config, &c); err != nil {
+		options.Log.WithError(err).Error("Invalid configuration")
+		return nil, engines.ErrContractViolation
+	}
+
+	return &engine{
+		config: &c,
+		log:    options.Log,
+	}, nil
 }
 
 type payloadType struct {
 	Link    string   `json:"link"`
 	Command []string `json:"command"`
+}
+
+type configType struct {
+	CreateUser bool     `json:"createUser"`
+	UserGroups []string `json:"userGroups"`
 }
 
 var payloadSchema = schematypes.Object{
@@ -46,10 +61,36 @@ var payloadSchema = schematypes.Object{
 			Items: schematypes.String{},
 		},
 	},
-	Required: []string{"command"},
 }
 
-func (e engine) PayloadSchema() schematypes.Object {
+var configSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"createUser": schematypes.Boolean{
+			MetaData: schematypes.MetaData{
+				Title: "Tells if a user should be created to run a command",
+				Description: `When set to true, a new user is created on the fly to run
+				the command. It runs the command from whitin the user's home directory.`,
+			},
+		},
+		"userGroups": schematypes.Array{
+			MetaData: schematypes.MetaData{
+				Title: "List of groups to assign user to",
+				Description: `It contains the list of groups the user will belong to.
+				The first item is the user's primary group, the rest will be user's
+				supplementary groups.`,
+			},
+			Items:  schematypes.String{},
+			Unique: true,
+		},
+	},
+	Required: []string{"createUser"},
+}
+
+func (e engineProvider) ConfigSchema() schematypes.Schema {
+	return configSchema
+}
+
+func (e *engine) PayloadSchema() schematypes.Object {
 	return payloadSchema
 }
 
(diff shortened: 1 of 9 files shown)


config/replaceobj.go
```diff
@@ -59,7 +59,7 @@ func jsonCompatTypes(val interface{}) error {
 				return err
 			}
 		}
-	case string, float64, bool, nil:
+	case string, float64, bool, int, nil:
```
jonasfj: Sorry, I've formalized config structure to be: string, float64, bool, nil, map[string]interface{}, and []interface{} types.
Checkout: https://github.com/taskcluster/taskcluster-worker/pull/125

I fixed things there so YAML doesn't unmarshal integers to int, but to float64, like json would...

engines/osxnative/engine.go
```diff
@@ -12,22 +12,40 @@ import (
...
 func (e engineProvider) NewEngine(options engines.EngineOptions) (engines.Engine, error) {
-	return engine{log: options.Log}, nil
+	var c configType
+	err := configSchema.Map(options.Config, &c)
+	if err == schematypes.ErrTypeMismatch {
+		panic("Type mismatch")
+	} else if err != nil {
+		options.Log.WithError(err).Error("Invalid configuration")
+		return nil, engines.ErrContractViolation
+	}
```
jonasfj: I haven't done this everywhere yet, but the pattern:
```
	var c configType
	if schematypes.MustMap(configSchema, options.Config, &c) != nil {
		return nil, engines.ErrContractViolation
	}
```
Is a bit simple... and panic is done is `MustMap`

See: https://github.com/taskcluster/taskcluster-worker/blob/master/plugins/interactive/interactive.go#L112-L115

engines/osxnative/engine.go
```diff
@@ -49,6 +67,33 @@ var payloadSchema = schematypes.Object{
...
+			},
+		},
+		"userGroups": schematypes.Array{
+			MetaData: schematypes.MetaData{
+				Title: "List of groups to assign user to",
+				Description: `It contains the list of groups the user will belong to.
+				The first item is the user's primary group, the rest will be user's
+				supplementary groups.`,
+			},
+			Items:  schematypes.String{},
```
jonasfj: I would specify a `Pattern` here... as you probably don't allow all strings...

engines/osxnative/engine.go
```diff
@@ -49,6 +67,33 @@ var payloadSchema = schematypes.Object{
...
+			Unique: true,
+		},
+	},
+	Required: []string{"createUser"},
```
jonasfj: You don't have to require it... You can, but booleans always default to false...
I guess it's totally fair enough to require that people consider if they want this or not when configuring..

engines/osxnative/engine.go
```diff
@@ -49,6 +67,33 @@ var payloadSchema = schematypes.Object{
...
+}
+
+func (e engineProvider) ConfigSchema() schematypes.Schema {
+	return configSchema
+}
+
 func (e engine) PayloadSchema() schematypes.Object {
```
jonasfj: Since engine now has data members... I would recommend that you pass it around as a pointer to avoid copying...

engines/osxnative/engine.go
```diff
@@ -12,22 +12,40 @@ import (
...
+	var c configType
+	err := configSchema.Map(options.Config, &c)
+	if err == schematypes.ErrTypeMismatch {
+		panic("Type mismatch")
+	} else if err != nil {
+		options.Log.WithError(err).Error("Invalid configuration")
+		return nil, engines.ErrContractViolation
+	}
+
+	return engine{
```
jonasfj: Giving engine data members, I would suggest you pass it as a pointer... to avoid copying for every method call... also if you were to mutate members it wouldn't work unless it's by reference...

jonasfj:
I don't know too much about the user group stuff on OS X... But from what I can see this looks good..
Feel free to land once you address nits outlined.

config/replaceobj.go
jonasfj: Not doing this is probably my only requirement here... everything else is nits, that you probably should address because it would be really easy :)

engines/osxnative/engine.go
walac: Actually the restriction in the string is the allowed group names, I prefer to leave the system decide. If the group doesn't exist, it will fail ahead.

engines/osxnative/engine.go
walac: `PayloadSchema` expects it by value in `EngineBase`. If I use a pointer here, there will be a conflict in the method signature. I had problems with this in the past. We can address it in another PR.

engines/osxnative/engine.go
jonasfj: I was just suggesting that you restrict it to a pattern like: `Pattern: "^[a-zA-Z0-9_][a-zA-Z0-9_-]*$"`

So that the group names can't start with dash, contain spaces or other "fun" things that could potentially cause issues if used as command line arguments... Anyways, I guess it's not really an attack vector as this is part of engine config, and anyone writing engine config has full system access.

jonasfj: yeah, nvm, it's probably a totally overkill thing.. when it's in engine config.

engines/osxnative/engine.go
jonasfj: I mean `engine` by pointer... ie. `func (e *engine)...`

walac: Made a test here. This used to be a problem in go 1.6, not the case for 1.7.

walac:
Addressed comments...

PR #127, opened 2016-10-10
Script engine

jonasfj:
Another engine to our collection...

This one is really simple, and when using it most plugins except maybe "livelog" should probably be disabled.


gregarndt:
I think the only plugin that will be needed by the jobs that will be used for this is the artifact plugin for test result logs.

jonasfj:
Nope, artifact plugin doesn't create logs... and won't be used...

The script will run in a temporary folder, any files left behind by it in this folder will be uploaded as artifacts.

There is no where in the task payload where you list what artifacts to extract.

engines/script/config.go
```diff
@@ -0,0 +1,60 @@
...
+var configSchema = schematypes.Object{
+	MetaData: schematypes.MetaData{
+		Title:       "Script Engine Configuration",
+		Description: `Configuration properties for the 'scriptengine'.`,
+	},
+	Properties: schematypes.Properties{
+		"script": schematypes.Array{
+			MetaData: schematypes.MetaData{
+				Title: "Script to Execute",
+				Description: `Script and arguments to execute. This script will be fed
```
walac: s/will be fed/will feed/ ?

engines/script/engine.go
```diff
@@ -0,0 +1,70 @@
...
+	engines.Register("script", engineProvider{})
+}
+
+func (engineProvider) ConfigSchema() schematypes.Schema {
+	return configSchema
+}
+
+func (engineProvider) NewEngine(options engines.EngineOptions) (engines.Engine, error) {
+	var config configType
+	if schematypes.MustMap(configSchema, options.Config, &config) != nil {
```
walac: nit: log the error

engines/script/sandbox.go
```diff
@@ -0,0 +1,138 @@
...
+	if _, ok := err.(*exec.ExitError); err != nil && !ok {
+		s.log.Error("Script execution failed, error: ", err)
+	}
+
+	// Upload artifacts if not aborted
+	if !s.aborted.Get() {
+		err2 := s.uploadArtifacts()
+		if err2 != nil {
+			success = false
+			s.context.LogError("Failed to upload artifacts")
```
walac: nit: add `err2` to the error string

engines/script/sandbox.go
```diff
@@ -0,0 +1,138 @@
...
+	resolve     atomics.Once
+	resultSet   engines.ResultSet
+	resultError error
+	resultAbort error
+	log         *logrus.Entry
+	aborted     atomics.Bool
+	done        chan struct{}
+}
+
+func (s *sandbox) run() {
```
walac: Why not put this code in `WaitForResult`?

engines/script/sandbox.go
jonasfj: No, I strongly disagree.
`err2` is an internal error.

---

So I see us as having two kinds of errors.
 1. errors that we blame the task author for
 2. internal errors... (all other errors)

If we are not 100% sure it's the task authors fault we have no business telling the task author what the internal error is.
In fact I view doing so as a security risk, error messages could contain headers with auth stuff etc... so we should **never** blindly log errors to task logs.

engines/script/engine.go
jonasfj: No, validation error is a contract error.

engines/script/config.go
jonasfj: No, the script is subject to the feeding.. :) hehe

ie. this is `echo "{...}" | script` so the payload is _fed_ to the script.

gregarndt:
Yea I was referring to the logs and files that the script leaves behind, not the task log of actual execution.  Great to know that the folder can just be uploaded.

engines/script/sandbox.go
walac: Is this only applicable for this engine or it is a general thing?

jonasfj: General thing...
Well, for all practical intends and purposes WaitForResult should always be called...
But I prefer not to have that call be driving the application... Furthermore, I totally expect that there could be more than one call to WaitForResult.

@walac, as a general rule of thumb, I've come to realize that methods like WaitForResult and Abort aren't good places to do be doing logic. Rather they should just initiate something and then wait for the result to be done.

Note: that execution should start before WaitForResult is called.

engines/script/engine.go
jonasfj: No, payload and config options are validated against the schema before being passed in here...
The validation part of map is therefor unnecessary, but nice. It enforces the contract that's all..

IMO, we could panic on the error if we wanted to...

PR #128, opened 2016-10-11
Added interactive shell-server for testing client implementations

jonasfj:
Some of the pty stuff is a bit messy... I want to make that nicer later...

Note: this is just for testing shell client implementations... Particularly those in node.js, so there is no need for detailed review...


walac:
Windows is down 👎

commands/shell-server/cmd.go
```diff
@@ -0,0 +1,73 @@
...
+type cmd struct{}
+
+func (cmd) Summary() string {
+	return "Host an interactive shell-server on localhost"
+}
+
+func (cmd) Usage() string {
+	return `
+taskcluster-worker shell-server will open a websocket to an interactive task, start
+a shell and expose it in your terminal. This is similar to using an SSH client.
```
walac: nit: s/an SSH/a SSH/

commands/shell-server/cmd.go
```diff
@@ -0,0 +1,73 @@
...
+  -s --shell <SHELL>      Default shell to use [default: ` + defaultShell + `].
+     --log-level <level>  Log level [default: INFO].
+  -h --help               Show this screen.
+`
+}
+
+func (cmd) Execute(args map[string]interface{}) bool {
+	log, err := runtime.CreateLogger(args["--log-level"].(string))
+	if err != nil {
+		fmt.Printf("Invalid log-level, error: %s", err)
```
walac: nit: print error to `fmt.Fprintf(os.Stderr, ...`

commands/shell-server/cmd.go
jonasfj: SSH is an abbreviation, and when pronouncing the letter "S" it sounds as if it's spelled "Ess". Hence, it's "an SSH" and not "a SSH"...

Okay, I could still be wrong, but in that case let's leave to for a naive English speaker  :) haha

commands/shell-server/cmd.go
jonasfj: decent point...

PR #134, opened 2016-10-13
Disabled cgo build on traivs

jonasfj:


jonasfj:
landing to make sure it actually works...

jonasfj:
@walac, @petemoore, please have a quick a look at these changes..

I basically just don't install everything, and I use `CGO_ENABLED=0` ensuring that we:
1) don't accidentally rely on something that needs CGO
2) statically link everything! Making deployment a LOT easier...

Note: couldn't set this globally as the race detector needs CGO inorder to work.

petemoore:
Seems good to me. Although we might want to test, by making a release.

jonasfj:
I already made a release... actually I made a few... At this point I'm marking them as prereleases...

PR #135, opened 2016-10-13
Script artifact folder

jonasfj:
One temporary folder per script task.. Files left in `artifacts/` of that folder will be uploaded as artifacts.


jonasfj:
@jmaher, have a look... this updates example `script.py` to create artifacts too..
(one day maybe we'll get around to writing proper tests for script engine, for now artifact upload isn't covered, but exit code and log is covered).

Also got it link statically, releases from now on will be statically linked.

PR #138, opened 2016-10-19
Streaming livelog

jonasfj:


runtime/client/queue.go
```diff
@@ -65,7 +75,93 @@ func (m *MockQueue) CreateArtifact(taskID, runID, name string, payload *queue.Po
...
+	if json.Unmarshal(*r, &s3req) != nil {
+		return false
+	}
+	return s3req.StorageType == "s3"
+})
+
+// ExpectS3Artifact will setup queue to expect an S3 artifact with given
+// name to be created for taskID and runID using m and returns
+// a channel which will receive the artifact.
+func (m *MockQueue) ExpectS3Artifact(taskID string, runID int, name string) <-chan []byte {
```
jonasfj: This method is super nice for when we have to test that artifacts are created :)

jonasfj:
@gregarndt, in case you're curious as to what the fix was... I suspect the `bufio.Reader` from before was buffering more than a single line.

PR #139, opened 2016-10-19
Implement context

jonasfj:
I think `TaskContext` should implement `context.Context`...

In future we should also not have `TaskContext` track status, it's totally unnecessary and only `worker/` package uses it. Plugins and engines should only care if a task is aborted/cancelled, and not whether it was an abort or a cancel.

They should only care that execution should stop...

### DIFF
--- a/commands/qemu-guest-tools/cmd.go
+++ b/commands/qemu-guest-tools/cmd.go
@@ -18,9 +18,10 @@ import (
 
 	"github.com/taskcluster/taskcluster-worker/commands"
 	"github.com/taskcluster/taskcluster-worker/runtime"
+	"github.com/taskcluster/taskcluster-worker/runtime/util"
 )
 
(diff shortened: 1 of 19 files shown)


jonasfj:
Note: this effectively two ways for a sandbox to be aborted...
1) Sandbox.Abort() or SandboxBuilder.Abort()
2) TaskContext.Abort()

We need two because (1) doesn't return until the task have been aborted... Where as (2) started the aborting proccess. Naturally, calling (1) should also start the aborting process.

We need to wait for abortion to be done, because it may hold a resource that needs to be released before we can start the next task...

jonasfj:
I hope that makes some sense... I hate it, that we have two methods of signaling aborting..
But I suspect that if always call both things will be okay... Granted we'll need to work on `worker/` to make sure that happens.

PR #141, opened 2016-10-26
Native engine

jonasfj:
Initial draft of native engine...

There is a still a few TODOs in there.
Notably:
 * Engine tests
 * Context URL extraction isn't done yet
 * Proxy support

Also there is no look at volume support yet.


jonasfj:
Regardless, please give this a quick review and let's land it... So we can iterate on missing features.

@walac, if you get bored you can explore feasibility of implementing the `system` package for OS X.
Tests are already written you just have to define a few commands, see comments.

engines/native/nativelinux_test.go
```diff
@@ -0,0 +1,117 @@
...
+		}`,
+	}
+
+	c.TestPingProxyPayload()
+	c.TestPing404IsUnsuccessful()
+	c.TestLiveLogging()
+	c.TestParallelPings()
+	c.Test()
+}
+*/
```
petemoore: Should this be deleted?

engines/native/system/process_darwin.go
```diff
@@ -0,0 +1,204 @@
...
+
+	if err != nil {
+		return nil, fmt.Errorf("Unable to execute binary, error: %s", err)
+	}
+
+	// Go wait for result
+	go p.waitForResult()
+
+	return p, nil
+}
```
petemoore: Is this a modified version of the system package library? How come we can't use the standard library natively? (I believe it has support for running processes as a different user on darwin/linux - even if not on Windows)

engines/native/system/process_windows.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// StartProcess starts a new process with given arguments, environment variables,
+// and current working folder, running as given user.
+//
+// Returns an human readable error explaining why the sub-process couldn't start
+// if not successful.
+func StartProcess(options ProcessOptions) (*Process, error) {
+	panic("Not implemented")
+}
```
petemoore: Nice. :-)

engines/native/system/user_darwin.go
```diff
@@ -0,0 +1,98 @@
...
+	"os/exec"
+	"os/user"
+	"strconv"
+	"strings"
+
+	"github.com/taskcluster/slugid-go/slugid"
+)
+
+const defaultShell = "/bin/bash"
+const systemUserAdd = "/usr/sbin/useradd"
```
petemoore: This doesn't exist on my Mac - are you sure this is right?

From memory, on darwin there are different commands for creating users, rather than `useradd`.

runtime/atomics/waitgroup.go
```diff
@@ -0,0 +1,113 @@
...
+
+// WaitGroup is similar to sync.WaitGroup, except it can enter a draining state
+// at which point additional calls to Add will fail and returns
+// ErrWaitGroupDraining
+type WaitGroup struct {
+	m        sync.Mutex
+	c        sync.Cond
+	count    int
+	draining bool
+}
```
petemoore: A custom `WaitGroup` implementation?

jonasfj: With a draining mode needed when tracking shells.

engines/native/system/user_windows.go
```diff
@@ -0,0 +1,21 @@
...
+// User is a representation of a system user account.
+type User struct {
+	name       string
+	homeFolder string
+}
+
+// CreateUser will create a new user, with the given homeFolder, set the user
+// owner of the homeFolder, and assign the user membership of given groups.
+func CreateUser(homeFolder string, groups []*Group) (*User, error) {
+	panic("Not implemented")
```
jonasfj: Note: it really was my hope that it would be simple to implement the system package on Windows and OS X... So we can quickly do it later.

engines/native/nativelinux_test.go
jonasfj: No, I plan to implement proxies later.. with some luck...

engines/native/system/process_darwin.go
jonasfj: We need some cross platform abstractions for setting users... And other things, figured the simplest thing was to hide the internals, so they can wrap exec.Command as needed...

PR #142, opened 2016-10-31
Fetcher sub-system

jonasfj:
The idea with this system is that we can defined different way to fetch something.
"something" always being a file of some sorts...

Now a plugin like caches that maybe allow tasks to reference external resources to be fetched and used to preload the cache, can just use `fetcher.URL.Schema()` as the schema...
In particular, they can use `fetcher.Combine(...)` to combine multiple fetcher implementations... So we can have a fetcher implementation for each of:
 * `{taskId, artifact}`
 * `{taskId, runId, artifact}`
 * `{index, artifact}`
 * `{index, runId, artifact}`
 * `{bucket, key}` // for fetching from S3 using temporary credentials from auth
And what's more we can easily add additional fetchers...

For QEMU images I might want one that supports fetching from S3, but elevates credentials from `worker:image:<name>` to some specific S3 bucket. So people can have access to the use the images, but not access to download the images.

Anyways, this PR just introduces a generic subsystem for fetching things, and combing manners for fetching things, so external reference can be fairly consistent.

Note: the `HashKey` and `Scopes` methods are for future, where I hope to implement  a caching sub-system that can wrap resources that have been fetched...


jonasfj:
@petemoore, btw, please have a look at:
https://github.com/taskcluster/taskcluster-worker/projects/1?fullscreen=true

runtime/fetcher/mocks_test.go
```diff
@@ -0,0 +1,120 @@
...
+	context.Context
+	queue client.Queue
+}
+
+func (c *mockContext) Queue() client.Queue {
+	return c.queue
+}
+
+func (c *mockContext) Progress(description string, percent float64) {
+	debug("Progress: %s - %d %", description, percent)
```
walac: percent symbol should be `%%`

PR #147, opened 2016-12-21
Port native engine to Mac OSX. r=jonasfj

walac:


engines/native/system/system_test.go
```diff
@@ -17,21 +19,81 @@ import (
...
 //       testSleep should be defined per platform
 
 func TestSystem(t *testing.T) {
+	var err error
+	var u *User
+
+	currentUser, err := user.Current()
+	assert.NoError(t, err)
+
+	if currentUser.Username != "root" {
```
jonasfj: This won't work on windows... Hmm, the reason I had the `system` build constraint was so that we would only run the tests when in a proper context. I guess there could also be contexts where tests are running in root, but we're testing something else and don't want to pollute with garbage users. Considering that the tests are likely to mess up your system, I think hiding then behind a build constraint might be sane.

But I see how auto detecting if the test can run is super nice :)

---
I guess we don't have to fix this now... We can just leave it like this, maybe add a TODO saying something about how this has to be refactored with testing on windows. Or add the build constraint back...

engines/native/system/system_test.go
```diff
@@ -17,21 +19,81 @@ import (
...
+	for _, s := range y {
+		if s == x {
+			return true
+		}
+	}
+
+	return false
+}
+
 // Note: Constants testGroup, testCat, testTrue, testFalse, testPrintDir, and
```
jonasfj: I love the addition of extra tests for user groups :)

Let's document the `testGroups` variable here. Just a small thing.

engines/native/system/process_darwin.go
```diff
@@ -143,20 +136,22 @@ func StartProcess(options ProcessOptions) (*Process, error) {
 		err = p.cmd.Start()
 	} else {
 		p.pty, err = pty.Start(p.cmd)
-		if options.Stdin != nil {
+		if err == nil {
```
walac: If `pty.Start` fails, worker will panic with memory access error.

engines/native/system/system_test.go
```diff
@@ -17,21 +19,81 @@ import (
...
+
+			gid, err = strconv.Atoi(g.Gid)
+			require.NoError(t, err)
+
+			groups = append(groups, &Group{gid})
+		}
+
+		u, err = CreateUser(homeDir, groups)
+		require.NoError(t, err)
+		su, err := user.Lookup(u.name)
```
jonasfj: Again this is platform specific.

engines/native/system/system_test.go
```diff
@@ -17,21 +19,81 @@ import (
...
+		require.NoError(t, err)
+
+		// some sanity checks
+		require.EqualValues(t, u.name, su.Username)
+
+		uid, err := strconv.ParseUint(su.Uid, 10, 32)
+		require.NoError(t, err)
+		require.EqualValues(t, u.uid, uint32(uid))
```
jonasfj: platform specific and accessing private variables...

If we want to do whitebox tests, let's make a `system_darwin_test.go` file and add darwin specific test cases in there...
Ideally, we can have both generic blackbox tests and platform specific whitebox tests. IMO, the whitebox testing is probably overkill, but it doesn't hurt to have it :)

engines/native/system/variables_darwin_test.go
```diff
@@ -0,0 +1,10 @@
+package system
+
+// test variables
```
jonasfj: Ah, this is nice... so much easier to track variables :)

walac:
@jonasfj I think I fixed all issues.

walac:
Added an extra commit to fix Mac OSX issues.

engines/native/sandbox.go
```diff
@@ -35,6 +38,13 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
 		return nil, fmt.Errorf("Failed to temporary folder, error: %s", err)
 	}
 
+	if b.payload.Context != "" {
+		if err = fetchContext(b.payload.Context, homeFolder.Path()); err != nil {
+			b.context.LogError(err)
+			return nil, err
```
jonasfj: If there is a problem with the payload you should return a `MalformedPayloadError` not arbitrary errors... We can't gracefully handling arbitrary errors.

engines/native/sandbox.go
```diff
@@ -76,6 +86,39 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
...
 }
 
+func fetchContext(context, destdir string) error {
+	filename, err := util.Download(context, destdir)
+	if err != nil {
+		return fmt.Errorf("Error downloading '%s': %v", context, err)
+	}
+
+	// TODO: verify if this will harm Windows
+	if err = os.Chmod(filename, 0700); err != nil {
```
jonasfj: Add a TODO to abstract this away with a method in the system package... We'll have to add such a method, I'm okay with doing that later...

engines/native/sandbox.go
```diff
@@ -35,6 +38,13 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
 		return nil, fmt.Errorf("Failed to temporary folder, error: %s", err)
 	}
 
+	if b.payload.Context != "" {
+		if err = fetchContext(b.payload.Context, homeFolder.Path()); err != nil {
```
jonasfj: Add a TODO to start downloading at creation of the sandbox builder, and wait for download to finish here.
Add another TODO to make that abortable.

engines/native/sandbox.go
```diff
@@ -76,6 +86,39 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
...
+	unpackedFile := ""
+	switch filepath.Ext(filename) {
+	case ".zip":
+		err = runtime.Unzip(filename)
+	case ".gz":
+		unpackedFile, err = runtime.Gunzip(filename)
+	}
+
+	if err != nil {
+		return fmt.Errorf("Error unpacking '%s': %v", filename, err)
```
jonasfj: Don't expose absolute filepath to users... They have no need to know that. The URL is more important.

engines/native/sandbox.go
```diff
@@ -76,6 +86,39 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
 	return s, nil
 }
 
+func fetchContext(context, destdir string) error {
+	filename, err := util.Download(context, destdir)
```
jonasfj: Add TODO to use the fetcher subsystem that I have PR for... This is exactly what it is intended for.

Add another TODO to use the cache subsystem when such system has been architected.

engines/native/sandbox.go
walac: We actually have no such option. Since we download the context to the user home directory, we cannot start the download before the user is actually created. And we have to wait to the download finishing before start the process. Maybe moving the user creation to an earlier stage as well.

jonasfj: In a world where we cache a reuse downloaded resources, we wouldn't download to the user folder.
Instead we would download to some temporary folder and extract from there every time we need it.

Like said, I think this is just TODOs as the subsystems for this isn't in place yet.

engines/native/sandbox.go
walac: Fixed.

engines/native/sandbox.go
walac: Added.

engines/native/sandbox.go
walac: Added.

engines/native/sandbox.go
walac: Fixed.

walac:
Regarding the unpack thing, indeed we can make it fancier by applying the factory design pattern.

PR #148, opened 2017-01-02
Bug 1266763: Implement maxruntime plugin. r=jonasfj

walac:
maxruntime plugin provides the maxRunTime payload field that defines a
timeout value for task execution. If a timeout happens, the task is
aborted with Expired status.


plugins/maxruntime/maxruntime.go
```diff
@@ -0,0 +1,88 @@
...
+		done:           make(chan bool),
+		log:            options.Log,
+	}, nil
+}
+
+func (tp *taskPlugin) Started(sandbox engines.Sandbox) error {
+	go func() {
+		select {
+		case <-time.After(time.Duration(tp.maxRunTime) * time.Millisecond):
+			tp.log.Error("Task expired")
```
jonasfj: Log message is not correct...
Please explain what happened here. MaxRunTime was excedded and therefore the task was killed... etc...

plugins/maxruntime/maxruntime.go
```diff
@@ -0,0 +1,88 @@
...
+
+type payloadType struct {
+	MaxRunTime int `json:"maxRunTime"`
+}
+
+var payloadSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"maxRunTime": schematypes.Integer{
+			MetaData: schematypes.MetaData{
+				Title:       "Task execution timeout (ms)",
```
jonasfj: I wonder if we should change this to seconds or minutes...
@petemoore What do think?

I think ms is very JS and JSON friendly... so I'm not oppose to it, but now is a great time to discuss this.

petemoore: @jonasfj Good question! Although ms has the advantage that you can specify sub-second intervals, it comes with the disadvantage that numbers just get bigger, e.g. one hour is 3600000 which at a glance is not easy to recognise as an hour (you have to carefully count the 0's). Since the time overhead of a task is at least a second, I think it isn't optimal if we plan for tasks that run for less than a second (i.e. we wouldn't optimise chunking so that tasks last less than a second) so it feels unnecessary to support this level of accuracy. I think `maxRunTime` should really be considered a safety net, so if your task runs for a crazy amount of time, it can be aborted, rather than it being used as a mechanism to intentionally fail tasks that don't execute in a precise amount of time - and as a safety mechanism, I would say seconds would be the absolute smallest unit we'd want to consider - I would also not be put off using minutes, although I maybe seconds is a safer middle ground, and is consistent with current worker implementations (at least, that is what generic worker uses).

jonasfj: Honestly, I'm tempted to want support for strings like `'5 hours'` and `'35 min'`, and then just defined a regexp for it. But that's probably crazy overkill :)
(and something we can play with later)

Yeah, let's just go to seconds.

plugins/maxruntime/maxruntime.go
```diff
@@ -0,0 +1,90 @@
...
+	return payloadSchema
+}
+
+func (plugin) NewTaskPlugin(options plugins.TaskPluginOptions) (plugins.TaskPlugin, error) {
+	var p payloadType
+	if err := schematypes.MustMap(payloadSchema, options.Payload, &p); err != nil {
+		options.Log.WithFields(logrus.Fields{
+			"error":   err,
+			"payload": options.Payload,
+		}).Error("Payload has invalid format")
```
jonasfj: Don't log here, this is an contract error... honestly, we might as well just panic..

jonasfj: But let's return `ErrContractViolation` as we do everywhere else... :)

walac: This is an error caused by the task, I don't think it is a good idea just panic it. Also, the caller doesn't have much detail of what exactly happened (like the `error` object), so I think it is a good idea to log it here.

jonasfj: this is a schema mismatch, the caller is responsible for validating against the schema.

That way we don't have to implement this in every engine/plugin...

The only thing you have to check here is conditions that aren't encoded in the schema.

PR #149, opened 2017-01-07
A bunch of fixes and improvements.

walac:
For details, look at individual commits.

### DIFF
--- a/engines/native/sandbox.go
+++ b/engines/native/sandbox.go
@@ -54,11 +54,20 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
 		return nil, fmt.Errorf("Failed to create temporary system user, error: %s", err)
 	}
 
+	env := map[string]string{}
+	for k, v := range b.env {
+		env[k] = v
+	}
+
+	env["HOME"] = homeFolder.Path()
+	env["USER"] = user.Name()
+	env["LOGNAME"] = user.Name()
+
 	// Start process
 	debug("StartProcess: %v", b.payload.Command)
 	process, err := system.StartProcess(system.ProcessOptions{
 		Arguments:     b.payload.Command,
-		Environment:   b.env,
+		Environment:   env,
 		WorkingFolder: homeFolder.Path(),
 		Owner:         user,
 		Stdout:        ioext.WriteNopCloser(b.context.LogDrain()),
--- a/examples/macosx-config.yml
+++ b/examples/macosx-config.yml
@@ -20,7 +20,16 @@ config:
       userGroups: ['staff']
   logLevel:         info
   plugins:
-    disabled:       ['success', 'interactive']
+    disabled:       ['interactive', 'maxruntime']
+    env:
+      extra:
+        PATH: {$env: PATH}
+        TMPDIR: {$env: TMPDIR}
+        LANG: {$env: LANG}
+        LC_ALL: {$env: LC_ALL}
+        TERM: {$env: TERM}
+        TERM_PROGRAM: {$env: TERM_PROGRAM}
+        SHELL: /bin/bash
   pollingInterval:  10
   queueBaseUrl:     https://queue.taskcluster.net/v1
   reclaimOffset:    120
--- a/plugins/artifacts/artifacts.go
+++ b/plugins/artifacts/artifacts.go
@@ -92,7 +92,15 @@ func (tp *taskPlugin) Stopped(result engines.ResultSet) (bool, error) {
 	}
 
 	if len(nonFatalErrs) > 0 {
-		return false, engines.MergeMalformedPayload(nonFatalErrs...)
+		// Only report an exception if the command executed successfully.
+		// The logic behind this it is that no artifact upload failure is
+		// expected for a succeeded run, but failed tasks might be missing
+		// some artifacts.
+		if result.Success() {
+			return false, engines.MergeMalformedPayload(nonFatalErrs...)
+		}
+
+		return false, nil
 	}
 	return true, nil
 }
--- a/plugins/env/env.go
+++ b/plugins/env/env.go
@@ -13,10 +13,27 @@ import (
 
 type plugin struct {
 	plugins.PluginBase
+	extraVars map[string]string
 }
 
 type payloadType struct {
 	Env map[string]string `json:"env"`
+}
+
+type config struct {
+	Extra map[string]string `json:"extra"`
+}
+
+var configSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"extra": schematypes.Map{
+			MetaData: schematypes.MetaData{
+				Title:       "Extra environment variables",
+				Description: "This defines extra environment variables to add to the engine.",
+			},
+			Values: schematypes.String{},
+		},
+	},
 }
 
 var payloadSchema = schematypes.Object{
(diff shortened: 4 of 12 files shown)


engines/native/sandbox.go
```diff
@@ -30,6 +30,25 @@ type sandbox struct {
 	wg         atomics.WaitGroup
 }
 
+// white list of environment variables that must
+// be passed to the child process
+var environmentWhitelist = []string{
+	"PATH",
+	"TMPDIR",
+	"PWD",
```
jonasfj: what is this, and it is good to pass down, if we're creating a sub-user...?

in docker images we ususally set:
```
# Set variable normally configured at login, by the shells parent process, these
# are taken from GNU su manual
ENV           HOME          /home/worker
ENV           SHELL         /bin/bash
ENV           USER          worker # username
ENV           LOGNAME       worker # username
ENV           HOSTNAME      taskcluster-worker
ENV           LANG          en_US.UTF-8
ENV           LC_ALL        en_US.UTF-8
```
based on GNU su manual.

engines/native/sandbox.go
```diff
@@ -30,6 +30,25 @@ type sandbox struct {
...
+	"EDITOR",
+	"LANG",
+	"LOGNAME",
+	"TERM",
+	"TERM_PROGRAM",
+	"TASK_ID",
```
jonasfj: Shouldn't these be set based on config variables rather than being passed through

engines/native/sandbox.go
```diff
@@ -30,6 +30,25 @@ type sandbox struct {
 	wg         atomics.WaitGroup
 }
 
+// white list of environment variables that must
+// be passed to the child process
+var environmentWhitelist = []string{
+	"PATH",
+	"TMPDIR",
```
jonasfj: PATH and TMPDIR makes sense to me to passthrough, the rest not much...

Setting a `HOME`, `EDITOR`, `LANG`, `LOGNAME`,..... variables as outlined in my command referrring to GNU su, also makes sense... probably with some of them being configurable.

plugins/artifacts/artifacts.go
```diff
@@ -92,7 +92,15 @@ func (tp *taskPlugin) Stopped(result engines.ResultSet) (bool, error) {
 	}
 
 	if len(nonFatalErrs) > 0 {
-		return false, engines.MergeMalformedPayload(nonFatalErrs...)
+		// Only report an exception if the command executed successfully.
+		// The logic behind this it is that no artifact upload failure is
+		// expected for a succeeded run, but failed tasks might be missing
+		// some artifacts.
```
jonasfj: Nice,

plugins/livelog/livelog.go
```diff
@@ -90,6 +90,14 @@ func (tp *taskPlugin) Prepare(context *runtime.TaskContext) error {
 }
 
 func (tp *taskPlugin) Finished(success bool) error {
+	return tp.uploadLog(success)
+}
+
+func (tp *taskPlugin) Exception(runtime.ExceptionReason) error {
+	return tp.uploadLog(false)
```
jonasfj: Nice, but let's add a boolean to ensure we don't upload the logs twice, that'll likely cause the queue to return an error for no good reason...

Actually, I'm not sure that it's not a bug, if `Finished` doesn't get called. (we might fix that later)

jonasfj: Note: adding the boolean, means it won't cause issues if we later ensure that `Finished` is always called.

plugins/livelog/livelog.go
```diff
@@ -90,6 +90,14 @@ func (tp *taskPlugin) Prepare(context *runtime.TaskContext) error {
...
+}
+
+func (tp *taskPlugin) uploadLog(success bool) error {
```
jonasfj: What is the parameter good for?

plugins/env/env.go
```diff
@@ -13,12 +13,29 @@ import (
...
 	Env map[string]string `json:"env"`
 }
 
+type config struct {
+	Extra map[string]string `json:"extra"`
+}
+
+var configSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"extra": schematypes.Map{
```
jonasfj: Can't we list these as `static` ie. static environment variables... Or `constant`

examples/macosx-config.yml
```diff
@@ -20,7 +20,22 @@ config:
...
+    env:
+      extra:
+        PATH: {$env: PATH}
+        TMPDIR: {$env: TMPDIR}
+        LANG: {$env: LANG}
+        LC_ALL: {$env: LC_ALL}
+        LOGNAME: {$env: LOGNAME}
+        TERM: {$env: TERM}
+        TERM_PROGRAM: {$env: TERM_PROGRAM}
+        TASK_ID: {$env: TASK_ID}
```
jonasfj: You can't do this :)

The worker runs multiple tasks right?

examples/macosx-config.yml
```diff
@@ -20,7 +20,22 @@ config:
...
+        RUN_ID: {$env: RUN_ID}
```
jonasfj: Same with this...

plugins/env/env.go
```diff
@@ -44,6 +61,10 @@ func (plugin) NewTaskPlugin(options plugins.TaskPluginOptions) (plugins.TaskPlug
 		return nil, engines.ErrContractViolation
 	}
 
+	for k, v := range pl.extraVars {
+		p.Env[k] = v
+	}
```
jonasfj: you should set `TASKID` and `RUNID` based on `options.Context`, I think it's in there somewhere in the `TaskContext`

examples/macosx-config.yml
```diff
@@ -20,7 +20,22 @@ config:
...
+        TASKCLUSTER_WORKER_TYPE: {$env: TASKCLUSTER_WORKER_TYPE}
```
jonasfj: WorkerType can probably be exposed through `TaskContext` instead... Or maybe `runtime.Environment` as passed in on plugin creation. Since this value doesn't change at runtime.

examples/macosx-config.yml
walac: Yeah, I already removed this locally.

plugins/env/env.go
walac: Sure, np.

examples/macosx-config.yml
walac: Removed, too.

plugins/env/env.go
walac: Ok. I actually not sure if this is relevant for the target task.

examples/macosx-config.yml
walac: Another useless set. Just removed it.

examples/macosx-config.yml
```diff
@@ -20,7 +20,17 @@ config:
...
   plugins:
-    disabled:       ['success', 'interactive']
+    disabled:       ['interactive', 'maxruntime']
+    env:
+      extra:
+        PATH: {$env: PATH}
+        TMPDIR: {$env: TMPDIR}
+        LANG: {$env: LANG}
+        LC_ALL: {$env: LC_ALL}
+        LOGNAME: {$env: LOGNAME}
```
jonasfj: `LOGNAME` is set in the native engine.

plugins/env/env.go
```diff
@@ -82,8 +103,21 @@ type pluginProvider struct {
 	plugins.PluginProviderBase
 }
 
-func (pluginProvider) NewPlugin(plugins.PluginOptions) (plugins.Plugin, error) {
-	return plugin{}, nil
+func (pluginProvider) NewPlugin(options plugins.PluginOptions) (plugins.Plugin, error) {
+	var c config
+	if err := schematypes.MustMap(configSchema, options.Config, &c); err != nil {
+		options.Log.WithError(err).Error("Invalid configuration")
```
jonasfj: Again no need to log on contract violations, caller is responsible for logging errors from schema mismatch

PR #154, opened 2017-01-27
Avoid using go:generate

jonasfj:
Let's just manually import stuff, that's all we use `go generate` for anyways.

It seems this will make things simpler... removing another step in the build process is always nice.


jonasfj:
This also adds some lovely top-level package docs for godoc :)

PR #157, opened 2017-02-05
Bug 1336585: do not create a user if CreateUser is false

djmitche:
This adds a new CreateUser method to the system.User struct.

If not creating a user, the engine avoids deleting the user or its home
directory, or killing its processes.

---

@walac: this passes tests and so on, but it's not clear how good the test coverage is for this code.  So:

1. Does this seem right?
1. How can I test this better?
1. What do you think about the "TODO" item in `resultset.go`?


engines/native/engine.go
```diff
@@ -32,6 +32,10 @@ func (engineProvider) ConfigSchema() schematypes.Schema {
 
 func (engineProvider) NewEngine(options engines.EngineOptions) (engines.Engine, error) {
 	var c config
+
+	// apply default values
+	c.CreateUser = true
+
```
walac: I have the feeling that `MustMap` will override it. As my suggestion to enforce explicitly configuration of this item, this is irrelevant anyway.

engines/native/resultset.go
```diff
@@ -140,6 +140,12 @@ func (r *resultSet) ExtractFolder(path string, handler engines.FileHandler) erro
 }
 
 func (r *resultSet) Dispose() error {
+	// if we didn't create a user for this sandbox, we shouldn't destroy it either.
+	// TODO: why is this cleanup here, and not in Sandbox?
+	if !r.engine.config.CreateUser {
+		return nil
+	}
```
walac: But we still need to remove `homeFolder`, right?

engines/native/config.go
```diff
@@ -3,7 +3,8 @@ package nativeengine
 import schematypes "github.com/taskcluster/go-schematypes"
 
 type config struct {
-	Groups []string `json:"groups,omitempty"`
+	Groups     []string `json:"groups,omitempty"`
+	CreateUser bool     `json:"createUser,omitempty"`
```
walac: I think it is safer if we enforce configuration here. aka remove `,omitempty`

engines/native/resultset.go
```diff
@@ -140,6 +140,12 @@ func (r *resultSet) ExtractFolder(path string, handler engines.FileHandler) erro
 }
 
 func (r *resultSet) Dispose() error {
+	// if we didn't create a user for this sandbox, we shouldn't destroy it either.
+	// TODO: why is this cleanup here, and not in Sandbox?
```
walac: Because we need to user folder to persist after task is done to extract artifacts after `WaitForResult` returns.

engines/native/config.go
```diff
@@ -28,6 +29,14 @@ var configSchema = schematypes.Object{
...
 			},
 		},
+		"createUser": schematypes.Boolean{
+			MetaData: schematypes.MetaData{
+				Title: "Tells if a user should be created to run a command",
+				Description: `When set to true, a new user is created on the fly to run
+				the command. It runs the command from whitin the user's home directory.
+				If false, the command runs without changing userid`,
+			},
+		},
```
walac: nit: s/command/task/

engines/native/sandbox.go
```diff
@@ -47,11 +47,20 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
...
+		user, err = system.CreateUser(homeFolder.Path(), b.engine.groups)
+		if err != nil {
+			homeFolder.Remove() // best-effort clean-up this is a fatal error
+			return nil, fmt.Errorf("Failed to create temporary system user, error: %s", err)
+		}
+	} else {
+		user, err = system.CurrentUser()
+		if err != nil {
+			return nil, err
+		}
```
walac: Maybe we should assign `nil` to `user` as a flag to say "we are running as current user", so, in case someone mistakenly tries to delete the current user, we get a `panic`, instead of a dead host.

engines/native/sandbox.go
```diff
@@ -207,10 +216,12 @@ func (s *sandbox) Abort() error {
 		// Remove temporary user (this will panic if unsuccessful)
 		s.user.Remove()
```
walac: Don't we need to add this inside the `s.engine.config.CreateUser` clause?

engines/native/system/user_darwin.go
```diff
@@ -21,6 +21,31 @@ type User struct {
...
+func CurrentUser() (*User, error) {
+	osUser, err := user.Current()
+	if err != nil {
+		return nil, err
+	}
+
+	// Uid and Gid are always decimal numbers on posix systems
+	uid, err := strconv.Atoi(osUser.Uid)
+	if err != nil {
+		panic("user id is not an integer")
```
walac: Maybe `panic` with `err` and `osUser.Uid` in the error string.

engines/native/system/user_darwin.go
```diff
@@ -21,6 +21,31 @@ type User struct {
...
+	}
+	gid, err := strconv.Atoi(osUser.Gid)
+	if err != nil {
+		panic("user id is not an integer")
```
walac: Maybe `panic` with `err` and `osUser.Gid` in the error string.

engines/native/system/user_linux.go
```diff
@@ -22,6 +22,31 @@ type User struct {
...
+func CurrentUser() (*User, error) {
+	osUser, err := user.Current()
+	if err != nil {
+		return nil, err
+	}
+
+	// Uid and Gid are always decimal numbers on posix systems
+	uid, err := strconv.Atoi(osUser.Uid)
+	if err != nil {
+		panic("user id is not an integer")
```
walac: Maybe `panic` with `err` and `osUser.Uid` in the error string.

engines/native/system/user_linux.go
```diff
@@ -22,6 +22,31 @@ type User struct {
...
+	}
+	gid, err := strconv.Atoi(osUser.Gid)
+	if err != nil {
+		panic("user id is not an integer")
```
walac: Maybe `panic` with `err` and `osUser.Gid` in the error string.

walac:
Another topic I forgot, process killing on abort. We can't safely rely on `KillByOwner` operation for this.

walac:
@djmitche do you want I pick this, since you will be on work trip this week?

djmitche:
That's probably a good idea -- I can easily envision this stretching until
Friday, and AIUI we need it sooner than that.

engines/native/config.go
```diff
@@ -28,6 +29,14 @@ var configSchema = schematypes.Object{
...
 	},
 	Required: []string{},
```
jonasfj: Maybe require that `createUser` is defined, rather that relying on `false` is the default...
Or modify `go-schema-types` to support default values :)

engines/native/native_test.go
```diff
@@ -36,6 +39,35 @@ func TestLogging(t *testing.T) {
...
 
+func TestLoggingNoUserCreation(t *testing.T) {
+	p := &enginetest.EngineProvider{
+		Engine: "native",
+		Config: `{
+            "createUser": false
+        }`,
+	}
+
+	user, err := system.CurrentUser()
```
jonasfj: Nice trick...

engines/native/sandbox.go
```diff
@@ -60,8 +77,8 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
 	}
 
 	env["HOME"] = homeFolder.Path()
```
jonasfj: Is this a good choice if we're using `system.currentUser` ?

jonasfj: Maybe instead of calling this variable for `homeFolder, err := b.engine.environment.TemporaryStorage.NewFolder()` we should call it for `workFolder`, still set it as working directory, and use it for homeDir if we're creating a new user...

engines/native/sandbox.go
```diff
@@ -47,11 +47,28 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
...
+
+	if b.engine.config.CreateUser {
+		// Create temporary user account
+		user, err = system.CreateUser(homeFolder.Path(), b.engine.groups)
+		if err != nil {
+			homeFolder.Remove() // best-effort clean-up this is a fatal error
+			return nil, fmt.Errorf("Failed to create temporary system user, error: %s", err)
+		}
+		username = user.Name()
+	} else {
```
jonasfj: If we're introducing `system.CurrentUser()` I propose that we allow for that to be passed as `Owner` property, and instead do:
```go
var user *system.User
if b.engine.config.CreateUser {
  user, err = system.CreateUser(homeFolder.Path(), b.engine.groups)
  ...
} else {
  user, err = system.CurrentUser()
  ...
}

....
env["USER"] = user.Username()
```

engines/native/sandbox.go
```diff
@@ -47,11 +47,28 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
...
+		var curUser *system.User
+
+		user = nil
+
+		curUser, err = system.CurrentUser()
+		if err != nil {
+			return nil, err
```
jonasfj: Don't forget to clean up `homeFolder.Remove() // best-effort clean-up this is a fatal error` as in the error handling for if-branch.

engines/native/sandbox.go
```diff
@@ -195,20 +214,36 @@ func (s *sandbox) Abort() error {
...
+		// We must kill the children after the parent
+		// to avoid the case of the parent creating new
+		// processes after we kill the descendents.
+		// On the other hand, we must kill children before
+		// the parent process is `waited`, because then
+		// the children process will be inherited by the init
+		// process in posix systems, and KillChildren won't
+		// work anymore.
+		// So, we run process.Kill() in a goroutine to
+		// "parallelize" the killing process.
```
jonasfj: This sounds like a super racy solution...

So in `process_posix.go` we should only do `go p.waitForResult()` the first time `process.Wait()` is called...
Instead of starting to wait on the process in `StartProcess`.... Though after thinking I don't think that'll help as `sandbox.Abort()`  maybe called while `sandbox.WaitForResult()` is waiting. So we would need some wait to abort the pending wait operation.

Either way, it seems like this is platform specific information... it belongs in `system`.
So let's create a `system.KillProcessTree(s.process)` instead... Where `KillProcessTree` terminates a process and all it's children. Exactly, how we get that working and avoid races can be handled in `system` where the other `Wait` call is likely blocking in parallel.

---
Otherwise, like the idea of doing `system.KillProcessTree` followed by `KillByOwner(s.user)` if `CreateUser` is true...

engines/native/system/process_posix.go
```diff
@@ -164,36 +155,59 @@ func StartProcess(options ProcessOptions) (*Process, error) {
...
+func KillByOwner(user *User) error {
+	// Create pkill command
+	uid := strconv.FormatUint(uint64(user.uid), 10)
+	return pkill("-u", uid)
+}
+
+// KillChildren will kill all process that are child of the given process.
+func KillChildren(process *Process) error {
+	pid := strconv.Itoa(process.cmd.Process.Pid)
+	return pkill("-P", pid)
```
jonasfj: from man:
> -P ppid,...
> Only match processes whose parent process ID is listed.

So what about grand children ?

We might have to use process groups see:
https://medium.com/@felixge/killing-a-child-process-and-all-of-its-children-in-go-54079af94773#.dft0n4rey

But I suspect the safe thing to do is: `SIGSTOP` then traverse children... and `SIGSTOP` all of them recursively... Until you hit the leafs of the trees and then do `SIGKILL` for all of them.

jonasfj: According to wikipedia STOP and KILL can't be intercepted.
> There are two signals which cannot be intercepted and handled: SIGKILL and SIGSTOP.

See: https://en.wikipedia.org/wiki/Unix_signal#Handling_signals

jonasfj: (I have no idea how to do this on windows, probably it's almost the same)

We definitely should kill children, grandchildren and the entire process tree... Of course there is no way to prevent tasks to create a childen B, have B fork creating C and then kill B, making C a child of pid 1 (init), hence, C survives this cleanup attempt.

engines/native/system/system_test.go
```diff
@@ -222,6 +222,26 @@ func TestSystem(t *testing.T) {
...
+		done := make(chan bool)
+		go func() {
+			p.Wait()
+			done <- waited.Get()
+		}()
+		time.Sleep(100 * time.Millisecond)
+		go func() {
+			require.NoError(t, KillChildren(p))
+		}()
+		waited.Set(true)
```
jonasfj: You need to move this into the `go func() {.... ` above if you want to test what the assert says...

djmitche:
Here's how we do the process cleanup now:
  https://github.com/mozilla/build-cleanslate/blob/master/cleanslate.py
I would suggest changing $HOME as well, as there are likely a lot of
utilities and other scripts that look in $HOME for dotfiles, so otherwise a
host might accumulate dotfiles over time, or even be poisoned by a
malicious dotfile.

Dustin
​

engines/native/sandbox.go
walac: We assign `user = nil` by design, so if some code path accidentally tries to delete the user, taskcluster-worker will panic but the host system is safe.

engines/native/resultset.go
```diff
@@ -140,17 +140,21 @@ func (r *resultSet) ExtractFolder(path string, handler engines.FileHandler) erro
...
 
 func (r *resultSet) Dispose() error {
-	// Halt all other sub-processes
-	err := system.KillByOwner(r.user)
-	if err != nil {
-		r.log.Error("Failed to kill all processes by owner, error: ", err)
-	}
+	var err error
+
+	if r.engine.config.CreateUser {
```
jonasfj: We need an `else` part to this right? killing the process tree if we didn't use `CreateUser`.

engines/native/sandbox.go
```diff
@@ -16,59 +16,77 @@ import (
...
+			workingFolder.Remove() // best-effort clean-up this is a fatal error
+			return nil, fmt.Errorf("Failed to create temporary system user, error: %s", err)
+		}
+		username = user.Name()
+	} else {
+		var curUser *system.User
+
+		user = nil
+
+		curUser, err = system.CurrentUser()
```
jonasfj: Can't we modify `system.StartProcess(system.ProcessOptions{` to accept `Owner:  system.CurrentUser()` ?

That way this becomes:
```go
var user *system.User
if b.engine.config.CreateUser {
  user, err = system.CreateUser(workingFolder.Path(), b.engine.groups)
} else {
  user, err = system.CurrentUser()
}
if (err != nil) {
  ....
}

env["LOGNAME"] = user.Name()
```

It would reduce a lot of special cases, and seems rather reasonable that `Owner` could be the user object returned from `CurrentUser()`.

engines/native/sandbox.go
```diff
@@ -170,16 +188,18 @@ func (s *sandbox) waitForTermination() {
 
 	s.resolve.Do(func() {
 		// Halt all other sub-processes
-		system.KillByOwner(s.user)
+		if s.engine.config.CreateUser {
+			system.KillByOwner(s.user)
+		}
```
jonasfj: Again, don't we need an alternative here for the other case? If we kill the process tree... I guess the process might no longer exist, or?

engines/native/system/process_posix.go
```diff
@@ -164,36 +201,63 @@ func StartProcess(options ProcessOptions) (*Process, error) {
...
-	return nil
+// KillByOwner will kill all process with the given owner.
+func KillByOwner(user *User) error {
+	// Create pkill command
+	uid := strconv.FormatUint(uint64(user.uid), 10)
+	return pkill("-u", uid)
+}
+
+// KillProcesses will kill root and all of its descendents.
+func KillProcesses(root *Process) error {
```
jonasfj: This is pretty neat, can we rename this `KillProcessTree` ? or `KillProcessAndDecendents`

engines/native/resultset.go
jonasfj: Hmmm... I guess we can't, am I right? Because the process is already dead at this point, so we can't do further attempts to cleanup... Children were either killed or now have parent id 1.

engines/native/sandbox.go
jonasfj: This is probably a nvm mind too... because you can am I right? The process is already dead and children have have parent pid 1.

PR #158, opened 2017-02-07
Switch to govendor. r=jonasfj

walac:
govendor is much easier to deal with than godep, and we don't need to
maintain the dependencies inside the repo.


petemoore:
@walac this has deleted most of the vendored libraries - this introduced the following lines changes:

```
+401 −80,716
```

I'll revert this change.

petemoore:
My mistake, ignore me! All fine - I see it is intentional. :-)

PR #160, opened 2017-02-09
Bug fixes for process startup without new user creation.

walac:


engines/native/sandbox.go
```diff
@@ -84,11 +84,15 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
 		// Stderr defaults to Stdout when not specified
 	})
 	if err != nil {
+		if b.engine.config.CreateUser {
+			user.Remove()
+		} else {
+			workingFolder.Remove()
+		}
```
jonasfj: it's nit an else... shouldn't we always remove the workingFolder?

engines/native/system/process_posix.go
```diff
@@ -135,6 +135,23 @@ func StartProcess(options ProcessOptions) (*Process, error) {
...
 
+	if options.Owner != nil {
+		currentUser, err := CurrentUser()
+		if err != nil {
+			return nil, err
+		}
+
+		// If we pass an owner to exec.Start, it will end up calling
+		// setngroups (even if we don't set any groups set), a syscall
+		// only that root is allowed to execute, causing non-privileged
```
jonasfj: that only root is allowed

PR #161, opened 2017-02-14
Bug 1339196: Make $HOME and os/user.Lookup().HomeDir equal. r=jonasfj

walac:
We can get user's home directory either but looking at the $HOME
environment variable or read the OS specific user database.

As is common for scripts to look for home directory just by reading $HOME
environment variable, our first approach was to not change user database
and modify the $HOME variable to a temporary directory when spawning
the task command without creating a new user. This way, we could clobber
the task just by removing the $HOME directory.

This approach worked for most tests but xpcshell. Particularly, it checks
if $HOME matches user database, making this test to fail.

As we may not be allowed to change even the current user home directory,
we fix this by using the user's true home directory.

This implies that we only need to create a temporary directory if we
create a new user to run the task, otherwise it is not necessary. We
also need to explicitly set the correct owner of the context file.

### DIFF
--- a/engines/native/sandbox.go
+++ b/engines/native/sandbox.go
@@ -30,37 +30,51 @@ type sandbox struct {
 	wg            atomics.WaitGroup
 }
 
-func newSandbox(b *sandboxBuilder) (*sandbox, error) {
-	// Create temporary home folder for the task
-	workingFolder, err := b.engine.environment.TemporaryStorage.NewFolder()
-	if err != nil {
-		b.log.Error("Failed to create temporary folder: ", err)
-		return nil, fmt.Errorf("Failed to temporary folder, error: %s", err)
-	}
-
-	if b.payload.Context != "" {
-		if err = fetchContext(b.payload.Context, workingFolder.Path()); err != nil {
-			b.context.LogError(err)
-			return nil, engines.NewMalformedPayloadError(
-				fmt.Sprintf("Error downloading %s: %v", b.payload.Context, err),
-			)
-		}
-	}
-
+func newSandbox(b *sandboxBuilder) (s *sandbox, err error) {
 	var user *system.User
+	var workingFolder runtime.TemporaryFolder
+
+	defer func() {
+		if err != nil {
+			if b.engine.config.CreateUser && user != nil {
+				user.Remove()
+			}
+
+			if workingFolder != nil {
+				_ = workingFolder.Remove()
+			}
+		}
+	}()
 
 	if b.engine.config.CreateUser {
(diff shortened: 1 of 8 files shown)


engines/native/sandbox.go
```diff
@@ -78,44 +91,40 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
...
+		return
 	}
 
-	s := &sandbox{
+	s = &sandbox{
 		engine:        b.engine,
 		context:       b.context,
 		log:           b.log,
-		workingFolder: workingFolder,
+		workingFolder: user.Home(),
```
jonasfj: okay, this makes sense...

engines/native/sandbox.go
```diff
@@ -19,7 +21,7 @@ type sandbox struct {
 	engine        *engine
 	context       *runtime.TaskContext
 	log           *logrus.Entry
-	workingFolder runtime.TemporaryFolder
+	workingFolder string
```
jonasfj: I propose keeping `runtime.TemporaryFolder` but leaving it `nil` if we're not creating a folder...

That way we reuse the cleanup logic in `runtime`

engines/native/sandbox.go
```diff
@@ -230,7 +254,7 @@ func (s *sandbox) Abort() error {
 		}
 
 		// Remove temporary home folder
-		err := s.workingFolder.Remove()
+		err := os.RemoveAll(s.workingFolder)
```
jonasfj: So now you're remove the home directory that the current user has... I think that can have negative side-effects :)
If reusing a home directory, I don't think we can do any cleanup.

engines/native/sandbox.go
```diff
@@ -125,6 +137,21 @@ func fetchContext(context, destdir string) error {
 	if err = os.Chmod(filename, 0700); err != nil {
 		return fmt.Errorf("Error setting file '%s' permissions: %v", filename, err)
 	}
+	u, err := osuser.Lookup(user.Name())
```
jonasfj: @walac shouldn't make these things part of the `system` package... I doubt this will work well across platforms...

jonasfj: Make we need something like `system.ChangeOwner(user, path)` ?

walac: I think golang implements chown as a noop in platforms that doesn't support it. But better put inside system.

jonasfj: Especially, since you might want to do ACL kind of stuff on windows..

If we move it to `system` then it's very clear that native engine has no platform dependent code..

PR #163, opened 2017-02-14
Native home

jonasfj:
This won't delete the home directory...
I think it's wrong to delete a home directory that we didn't create.

### DIFF
--- a/engines/native/sandbox.go
+++ b/engines/native/sandbox.go
@@ -3,7 +3,9 @@ package nativeengine
 import (
 	"fmt"
 	"os"
+	osuser "os/user"
 	"path/filepath"
+	"strconv"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/taskcluster/taskcluster-worker/engines"
@@ -19,7 +21,7 @@ type sandbox struct {
 	engine        *engine
 	context       *runtime.TaskContext
 	log           *logrus.Entry
-	workingFolder runtime.TemporaryFolder
+	workingFolder string
 	user          *system.User
 	process       *system.Process
 	env           map[string]string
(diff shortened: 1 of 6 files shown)


engines/native/sandbox.go
```diff
@@ -78,26 +90,22 @@ func newSandbox(b *sandboxBuilder) (*sandbox, error) {
...
-
-		workingFolder.Remove()
-
 		// StartProcess provides human-readable error messages (see docs)
 		// We'll convert it to a MalformedPayloadError
-		return nil, engines.NewMalformedPayloadError(
+		err = engines.NewMalformedPayloadError(
 			"Unable to start specified command: ", b.payload.Command, "error: ", err,
 		)
+		b.context.LogError(err)
```
jonasfj: Note. things like this is not useful...
If we return malformed payload, the worker package is responsible for logging that error...
That said it would make a lot of sense to use `b.context.LogError(...` to log things that aren't part of the malformed payload error... such as a text explaining what happened... But I suspect that's not something we do initially.

engines/native/sandbox.go
```diff
@@ -30,37 +32,47 @@ type sandbox struct {
...
+				_ = workingFolder.Remove()
+			}
+		}
+	}()
 
 	if b.engine.config.CreateUser {
+		// Create temporary home folder for the task
+		workingFolder, err = b.engine.environment.TemporaryStorage.NewFolder()
+		if err != nil {
+			err = fmt.Errorf("Failed to temporary folder, error: %s", err)
```
walac: nit: "Failed to create..."

engines/native/sandbox.go
```diff
@@ -30,37 +32,47 @@ type sandbox struct {
...
 		user, err = system.CreateUser(workingFolder.Path(), b.engine.groups)
 		if err != nil {
-			workingFolder.Remove() // best-effort clean-up this is a fatal error
-			return nil, fmt.Errorf("Failed to create temporary system user, error: %s", err)
+			err = fmt.Errorf("Failed to create temporary system user, error: %s", err)
+			return
 		}
 	} else {
 		user, err = system.CurrentUser()
 		if err != nil {
```
walac: nit: this can be one line:

```go
if user, err = system.CurrentUser(); err != nil {
```

engines/native/sandbox.go
walac: Does it log inside task context?

walac:
Closing as I added these changes to #161

PR #164, opened 2017-02-15
Introduce a config setting to limit the number of tasks to be claimed

petemoore:
This makes integration testing easier (so tests can run worker to just pick up as many tasks as they have generated) and also makes it easier to customise worker behaviour from outside the worker (e.g. to execute arbitrary steps between each task).

### DIFF
--- a/worker/config.go
+++ b/worker/config.go
@@ -35,6 +35,7 @@ type configType struct {
 	MaxLifeCycle     int                    `json:"maxLifeCycle"`
 	MinimumDiskSpace int64                  `json:"minimumDiskSpace"`
 	MinimumMemory    int64                  `json:"minimumMemory"`
+	MaxTasksToRun    int                    `json:"maxTasksToRun"`
 }
 
 type credentials struct {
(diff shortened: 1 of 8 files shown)


worker/config.go
```diff
@@ -238,13 +239,23 @@ func ConfigSchema() schematypes.Object {
...
 			},
+			"maxTasksToRun": schematypes.Integer{
+				MetaData: schematypes.MetaData{
+					Title: "Number of tasks the worker should run before exiting",
+					Description: `If set to 0, the worker does not limit the number of tasks
+						it will claim and execute. For positive values > 0, the worker will
+						exit if it completes the given number of tasks.`,
+				},
+				Minimum: 0,
+				Maximum: math.MaxInt32,
```
jonasfj: Nit but no limit is signaled with `MaxInt64`

worker/queueservice.go
```diff
@@ -85,10 +87,11 @@ func (q *queueService) Start() <-chan *taskClaim {
 
 	go func() {
 		for !q.halt.Get() {
-			q.mu.RLock()
-			capacity := q.capacity
-			q.mu.RUnlock()
-			tasks := q.retrieveTasksFromQueue(capacity)
+			claimSize, finished := q.claimSize()
+			if finished {
+				break
```
jonasfj: Shouldn't it just skip claiming? I might be wrong here... ie. Sleep a second and then `continue`.

Ideally, we probably should be doing Cond.Wait or something, but let's not complicate with that yet...

worker/config.go
petemoore: See https://golang.org/pkg/builtin/#int

worker/queueservice.go
petemoore: We should break from the loop once we have claimed all of the tasks we are allowed to (i.e. once we hit `maxTasksToRun`).

worker/config.go
petemoore: See https://golang.org/pkg/builtin/#int

worker/queueservice.go
petemoore: We should break from the loop once we have claimed all of the tasks we are allowed to (i.e. once we hit `maxTasksToRun`). At this point we know we cannot claim any more tasks, so no point looping.

PR #166, opened 2017-02-17
Make sure Env is not nil but an empty map, if no env supplied in task payload

petemoore:
@jonasfj seem reasonable? Basically, if no env is in task payload, you'd end up having a nil map, which you then can't later modify. Maybe `Map` should take care of this, this was just a quick fix.

### DIFF
--- a/plugins/env/env.go
+++ b/plugins/env/env.go
@@ -48,11 +48,11 @@ var payloadSchema = schematypes.Object{
 	},
 }
 
-func (plugin) PayloadSchema() schematypes.Object {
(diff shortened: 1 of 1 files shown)


jonasfj:
Maybe leave a comment after the line explaining that you need if it no env vars are set as default values will be compiled over....

jonasfj:
@walac  we should change this to be reference now... If *plugin, because it now contains fields... And we don't want to copy those around.. even if array/map field are by reference...

If we as a mutex it won't be fun... By value is only fun for field less plugins

PR #169, opened 2017-02-23
Add initial .taskcluster.yml

petemoore:
Initial taskcluster-github integration


petemoore:
Note, the failing windows test is fixed in #171

petemoore:
Note, at the time I started this, I had problems running taskcluster-worker CI inside a taskcluster-worker. It may be that these issues have disappeared. I chose to use generic-worker for running the CI on OS X as it seemed to be stable, but @djmitche has quite rightly pointed out, it might not be wise to be deploying generic-worker on mac if we can avoid it, since it is just another thing that would need to be migrated over later to taskcluster-worker. So I'll play again with this tomorrow and see if I can get at least the mac CI tests running in a taskcluster-worker environment.

Note, Windows CI will have to be in generic-worker for the time being, and the linux integration tests will have to continue to run in docker-worker too.

jonasfj:
If this doesn't fail... we can talk about landing it :)

.taskcluster.yml
```diff
@@ -0,0 +1,169 @@
...
+    metadata:
+      name: "Build/test 64 bit taskcluster-worker on Debian GNU/Linux 8 (jessie)"
+      description: "This builds the 64 bit linux version of taskcluster-worker"
+      owner: "{{ event.head.user.email }}" # the user who sent the pr/push e-mail will be inserted here
+      source: "{{ event.head.repo.url }}"  # the repo where the pr came from will be inserted here
+    extra:
+      github:
+        # Events that will trigger this task
+        events:
+          - pull_request.*
```
jonasfj: I'm not a super big fan of this... anyone can open PRs..

petemoore: I've removed for now, we can discuss if we want to add it back in. I believe it would use the base revision of the `.taskcluster.yml` file rather than the head revision, so I don't think it is such a security concern, although someone could maliciously tie up our workers (although we could report them to github).

PR #170, opened 2017-02-23
Make interrupt channel size match number of unique interrupts it can receive

petemoore:
This change is based on my interpretation of the godocs - let me know if you disagree! I wasn't entirely sure.

### DIFF
--- a/commands/work/work.go
+++ b/commands/work/work.go
@@ -44,7 +44,7 @@ func (cmd) Execute(args map[string]interface{}) bool {
 		w.Start()
 		close(done)
 	}()
-	c := make(chan os.Signal, 1)
+	c := make(chan os.Signal, 2)
 	signal.Notify(c, os.Interrupt, os.Kill)
 	select {
 	case <-c:


walac:
iiuc, the buffer size is set according to the signal receiving rate (oh my, now I realized go programs may loose signal notifications), not the number of signal types handled. Anyway, the `os.Kill` in notify parameter means nothing, since process cannot handle `SIGKILL`.

petemoore:
So maybe buffer size 1 is ok after all? The [docs](https://golang.org/pkg/os/signal/#Notify) say:

> Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient. 

So since we would gracefully shutdown if any signal comes in, I think we don't need to worry if we get more than one signal. We just need to make sure we have at least a buffer of one, so that if a signal comes in, it doesn't get dropped, and then we can handle it, and any future signals we receive we can ignore as we are shutting down the go program anyway.

If this is the case, we can probably close this PR - @walac @jonasfj do you interpret it like this too?

walac:
I think this can be closed

PR #172, opened 2017-02-23
Add first integration test

petemoore:
This is hopefully the first of many to come. This is a very basic one at the moment, just to get ball rolling. I'll be updating it to check for task status etc - I'd just like to get some stuff landed before it bitrots.

Although the test itself is quite simple, it puts the pieces in place in order for us to be able to add more integration tests.


integrationtest/helper.go
```diff
@@ -0,0 +1,163 @@
...
+	// graph later for troubleshooting
+	taskGroupID string = slugid.Nice()
+	testdata    string = testdataDir()
+)
+
+func testdataDir() string {
+	// some basic setup...
+	cwd, err := os.Getwd()
+	if err != nil {
+		panic(fmt.Errorf("Could not get current directory during test package initialisation: %v", err))
```
walac: instead of `panic`ing the test, I think it is better to pass `*testing.T` to the function and fail the test in case of error.

integrationtest/helper.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	return filepath.Join(cwd, "testdata")
+}
+
+type TaskPayload struct {
+
```
walac: nit: remove empty line

integrationtest/helper.go
```diff
@@ -0,0 +1,163 @@
...
+		//
+		// Syntax:     ^([\x20-\x2e\x30-\x7e][\x20-\x7e]*)[\x20-\x2e\x30-\x7e]$
+		Name string `json:"name"`
+
+		// File system path of artifact
+		//
+		// Syntax:     ^.*[^/]$
+		Path string `json:"path"`
+
+		// Artifacts can be either an individual `file` or a `directory` containing potentially multiple files with recursively included subdirectories
```
walac: nit: break the long line

integrationtest/helper.go
```diff
@@ -0,0 +1,163 @@
...
+		// Possible values:
+		//   * "file"
+		//   * "directory"
+		Type string `json:"type"`
+	} `json:"artifacts,omitempty"`
+
+	// Command to execute
+	Command []string `json:"command"`
+
+	// Optional URL for a gzipped tar-ball to downloaded and extracted in the HOME directory for running the command.
```
walac: nit: "to be download.."
nit: break the long line

integrationtest/helper.go
```diff
@@ -0,0 +1,163 @@
...
+		Scopes:        []string{},
+		Tags:          json.RawMessage(`{}`),
+		Priority:      "normal",
+		TaskGroupID:   taskGroupID,
+		WorkerType:    workerType,
+	}
+	return
+}
+
+func SubmitTask(t *testing.T, td *queue.TaskDefinitionRequest, payload TaskPayload) (taskID string, myQueue *queue.Queue) {
```
walac: nit: break the long line

integrationtest/helper.go
```diff
@@ -0,0 +1,163 @@
...
+	taskID = slugid.Nice()
+	// check we have all the env vars we need to run this test
+	if os.Getenv("TASKCLUSTER_CLIENT_ID") == "" || os.Getenv("TASKCLUSTER_ACCESS_TOKEN") == "" {
+		t.Skip("Skipping test since TASKCLUSTER_CLIENT_ID and/or TASKCLUSTER_ACCESS_TOKEN env vars not set")
```
walac: should we skip or fail the test?

integrationtest/helper.go
```diff
@@ -0,0 +1,163 @@
...
+	creds := &tcclient.Credentials{
+		ClientID:    os.Getenv("TASKCLUSTER_CLIENT_ID"),
+		AccessToken: os.Getenv("TASKCLUSTER_ACCESS_TOKEN"),
+		Certificate: os.Getenv("TASKCLUSTER_CERTIFICATE"),
+	}
+	myQueue = queue.New(creds)
+
+	b, err := json.Marshal(&payload)
+	if err != nil {
+		t.Fatalf("Could not convert task payload to json")
```
walac: how about using [testify](https://github.com/stretchr/testify) package like other tests do?

integrationtest/helper.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	payloadJSON := json.RawMessage{}
+	err = json.Unmarshal(b, &payloadJSON)
+	if err != nil {
+		t.Fatalf("Could not convert json bytes of payload to json.RawMessage")
```
walac: better including `err` in error messages, unless you use `testify`, and then you get that for free :)

integrationtest/helper.go
petemoore: This is executed during package initialisation, not during a test, so we have to panic - no test is running at this point.

integrationtest/helper.go
petemoore: Note: the `TaskPayload` definition is generated from the exported json schema, but I can tweak it here.

integrationtest/helper.go
petemoore: I'll adjust this too...

integrationtest/helper.go
petemoore: In automation, we'll have the env vars set - so for me it was mostly a question for what we want to do when we run locally, or a contributor runs locally.

An advantage of skipping is that you can still see, for the tasks that you ran, that they passed. I'm guessing as this package grows, sometimes we'll just want to run the (quick) unit tests, and other times we'll want to also run integration tests.

Maybe we should introduce two separate make targets, for running unit tests, and for running integration tests, and then if you run integration tests, you are required to have these env vars set, but if you run unit tests, the integrationtest package test suite doesn't run... Hmmmm.

Should I implement as two different `make` targets?

Also curious what @jonasfj thinks...

integrationtest/helper.go
petemoore: Would that offer an advantage here?

integrationtest/helper.go
petemoore: whoops - good spot!

integrationtest/helper.go
walac: Ah, in this case, we are safe skipping tests...

integrationtest/helper.go
walac: In general it makes code simpler. In this case it would just be `require.NoError(err)`.
And it is also a praise for consistency, as we are writing new tests using it.

jonasfj:
How about we put integration tests in the `test/` or `tests/` folder... That's usually where all integration tests go... Because normal tests are just part of whatever package.

integrationtest/helper.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// If true, reboot the machine after task is finished.
+	Reboot bool `json:"reboot,omitempty"`
+}
+
+func RunTestWorker(workerType string) {
+	os.Setenv("TASKCLUSTER_CAPACITY", "1")
+	os.Setenv("TASKCLUSTER_WORKER_TYPE", workerType)
+	os.Setenv("TASKCLUSTER_WORKER_ID", workerType)
+	os.Setenv("TASKCLUSTER_MAX_TASKS", "1")
```
jonasfj: This is a cool idea... I guess we could use this to overwrite baseUrls... like `QUEUE_BASE_URL`, etc...

jonasfj: Obviously, we would then have to add them to the config, but then we wouldn't have to run against production queue.

integrationtest/helper.go
jonasfj: also don't use the name `myQueue`, please...

integrationtest/helper.go
jonasfj: yes, you'll love testify: `require.NoError(t, err)` it's super nice...

integrationtest/helper.go
jonasfj: i don't mind the panic, it's simple...

But perhaps wrap it in a function... like `pathToTestData() string` and have that panic if there is issues..

Randomly note: http://stackoverflow.com/a/18537419/68333

petemoore: > But perhaps wrap it in a function... like pathToTestData() string and have that panic if there is issues..

@jonasfj Did you look at the code? That is exactly what I have done! ;-)

integrationtest/helper.go
```diff
@@ -0,0 +1,163 @@
...
+	commands.Run(
+		[]string{
+			"work",
+			filepath.Join("testdata", "worker-config.yml"),
```
petemoore: I can conclusively say that `"testdata"` is indeed a string - not an int, nor a struct, nor any other type. The compiler determines this based on the `"` characters. Or am I missing something? :-)

petemoore: Ah - maybe you are asking if this literal string should instead be an absolute path, determined at runtime? No - tests are guaranteed to be executed from the directory they are contained in, so the relative path to the `testdata` subdirectory is guaranteed to work.

integrationtest/helper.go
petemoore: ok, you've convinced me! I'll do that.

petemoore:
> But can I ask that we maybe put this into test/native/
and then put the helper methods for making integration tests into test/ or test/helper/ or test/integration/ ?

> I prefer to not have too many top-level folders... This repo is bound to get huge. So if we can hide all integration tests in one folder that would be nice.

@jonasfj Let's discuss when you're online later today. I don't really understand, because hiding integration tests in one folder is what this PR does - is it the choice of name `integrationtest` that you don't like? I think nothing here should go in `test/native` since I think the unit tests that don't submit real tasks to the queue should go in there, and the time-consuming integration tests that hit the production queue should all be co-located in the `integrationtest` directory, even if it is testing the native engine (since each integration test runs against a single engine).

petemoore:
> But can I ask that we maybe put this into test/native/
and then put the helper methods for making integration tests into test/ or test/helper/ or test/integration/ ?
>
> I prefer to not have too many top-level folders... This repo is bound to get huge. So if we can hide all integration tests in one folder that would be nice.

@jonasfj Let's discuss when you're online. I can rename `/integrationtest` to `/test` if you prefer. I think for the most part, the integration tests we'll write will be engine-independent, so if we write an integration test, and we have multiple engines, we'll probably want to run the same integration test against all engines. There may be some slight change to the test payload and worker config based on the engine payload/config requirements, but we can probably localise that to a config file transformation and task payload transformation per engine (something like test "aspects").

In that case, probably we'd have `/test/<engine>` per engine, where that directory would contain the payload/config transformations required for that engine, and then the `/test` directory could contain subdirectories for different integration test suites, as we see fit. Does that seem reasonable to you?

PR #173, opened 2017-02-23
Look for ineffective assignments

petemoore:
Has been useful for me, for discovering otherwise hard-to-discover obscure bugs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,9 @@ check: test
 	# tests should fail if go fmt results in uncommitted code
 	git status --porcelain
 	/bin/bash -c 'test $$(git status --porcelain | wc -l) == 0'
+	go get github.com/gordonklaus/ineffassign
+	"${GOPATH}/bin/ineffassign" .
+
 test:
 	go get github.com/golang/lint/golint
 	go test -v -race $$(go list ./... | grep -v /vendor/)


petemoore:
Note, the reported ineffective assignment in travis is fixed in another PR, so let's not fix it in here (otherwise we'll hit a merge conflict later).

walac:
Ok, but this PR needs to rebase to tests pass and merge button be enabled

Makefile
```diff
@@ -32,6 +32,9 @@ check: test
 	# tests should fail if go fmt results in uncommitted code
 	git status --porcelain
 	/bin/bash -c 'test $$(git status --porcelain | wc -l) == 0'
+	go get github.com/gordonklaus/ineffassign
+	"${GOPATH}/bin/ineffassign" .
```
jonasfj: let's use metalinter instead:
https://github.com/alecthomas/gometalinter

Then we can specify the rules we want to run...

petemoore: That's being done in #179 but may take a while to land. Let's go with this for now, until #179 lands.

jonasfj:
Hmm, okay... But we really should land #173 soon instead.. it'll create
lots of merge conflicts...

Den 24. feb. 2017 2.51 AM skrev "Pete Moore" <notifications@github.com>:

petemoore:
True! OK, will close. :-)

PR #174, opened 2017-02-23
Prefer standard library log.Panic(...) over fmtPanic(...)

petemoore:
`fmtPanic` was unnecessary as there is already `log.Panic` in the standard library. This also has advantage of writing to standard error rather than standard out.

### DIFF
--- a/engines/enginetest/attachvolume.go
+++ b/engines/enginetest/attachvolume.go
@@ -59,8 +59,10 @@ func (c *VolumeTestCase) TestWriteReadVolume() {
 		log.Panic("Running with writeVolumePayload didn't finish successfully")
 	}
 	if !c.readVolume(volume, false) {
-		log.Panic("Running with CheckVolumePayload didn't finish successfully, ",
-			"after we ran writeVolumePayload with same volume (writing something)")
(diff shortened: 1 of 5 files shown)


engines/enginetest/attachvolume.go
```diff
@@ -55,10 +56,10 @@ func (c *VolumeTestCase) TestWriteReadVolume() {
 	nilOrPanic(err, "Failed to create a new cache folder")
 	defer evalNilOrPanic(volume.Dispose, "Failed to dispose cache folder")
 	if !c.writeVolume(volume, false) {
-		fmtPanic("Running with writeVolumePayload didn't finish successfully")
+		log.Panic("Running with writeVolumePayload didn't finish successfully")
 	}
 	if !c.readVolume(volume, false) {
-		fmtPanic("Running with CheckVolumePayload didn't finish successfully, ",
+		log.Panic("Running with CheckVolumePayload didn't finish successfully, ",
 			"after we ran writeVolumePayload with same volume (writing something)")
```
jonasfj: you have to fix alignment here... and other places... please :)

PR #178, opened 2017-02-23
Clean up pluginmanager

petemoore:
Some simplifications...

### DIFF
--- a/plugins/pluginmanager.go
+++ b/plugins/pluginmanager.go
@@ -41,26 +41,6 @@ func mergeErrors(errs ...error) error {
 	return nil
 }
 
-// waitForErrors executes function f against every plugin passed in taskPlugins
-// and returns an error which represents the merge of all errors which occurred
-// against any plugin.
-//
-// Note, that errors might be nil, if all are nil it'll return nil otherwise
-// it'll merge the errors.
-func waitForErrors(taskPlugins []TaskPlugin, f func(p TaskPlugin) error) (err error) {
-	errors := make([]error, len(taskPlugins))
-	var wg sync.WaitGroup
-	for i, j := range taskPlugins {
-		wg.Add(1)
-		go func(i int, j TaskPlugin) {
-			defer wg.Done()
-			errors[i] = f(j)
-		}(i, j)
-	}
-	wg.Wait()
-	return mergeErrors(errors...)
-}
-
 // PluginManagerConfigSchema returns configuration for PluginOptions.Config for
 // NewPluginManager.
 func PluginManagerConfigSchema() schematypes.Object {
@@ -234,7 +214,9 @@ func (m *pluginManager) NewTaskPlugin(options TaskPluginOptions) (manager TaskPl
 	return
 }
 
-func (m *taskPluginManager) executePhase(f func(p TaskPlugin) error) error {
+type TaskPluginPhase func(TaskPlugin) error
+
+func (m *taskPluginManager) executePhase(f TaskPluginPhase) error {
 	// Sanity check that no two methods on plugin is running in parallel, this way
 	// plugins don't have to be thread-safe, and we ensure nothing is called after
 	// Dispose() has been called.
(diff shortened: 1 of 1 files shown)


plugins/pluginmanager.go
```diff
@@ -41,26 +41,24 @@ func mergeErrors(errs ...error) error {
...
 // it'll merge the errors.
-func waitForErrors(errors <-chan error, count int) error {
-	if count == 0 {
-		return nil
+func waitForErrors(taskPlugins []TaskPlugin, f func(p TaskPlugin) error) (err error) {
+	errors := make([]error, len(taskPlugins))
+	var wg sync.WaitGroup
+	for i, j := range taskPlugins {
+		wg.Add(1)
+		go func(i int, j TaskPlugin) {
```
walac: `j` shadows the outer scoped `j`
`i` shadows the outer scoped `i`

You can access the outer variables inside `func`, so you can do
```go
go func () {
```

plugins/pluginmanager.go
```diff
@@ -201,41 +199,39 @@ func (m *pluginManager) PayloadSchema() schematypes.Object {
...
+	taskPlugins := make([]TaskPlugin, len(m.plugins))
+	errors := make([]error, len(m.plugins))
 
-	errors := make(chan error)
+	var wg sync.WaitGroup
 	for i, p := range m.plugins {
-		go func(name string, p Plugin) {
-			taskPlugin, err := p.NewTaskPlugin(TaskPluginOptions{
+		wg.Add(1)
+		go func(i int, p Plugin) {
```
walac: `p` shadows outer scoped `p`
`i` shadows outer scoped `i`

plugins/pluginmanager.go
```diff
@@ -302,17 +286,14 @@ func (m *taskPluginManager) Stopped(r engines.ResultSet) (bool, error) {
...
-			if !success {
-				result.Set(false)
-			}
-			errors <- err
-		}(p)
-	}
-	return result.Get(), waitForErrors(errors, len(m.taskPlugins))
+	err := waitForErrors(m.taskPlugins, func(p TaskPlugin) error {
+		success, err := p.Stopped(r)
+		if !success {
```
walac: is `success` valid if `err != nil`?

plugins/pluginmanager.go
petemoore: I'm afraid that's not quite right - you need them, as the go routines run concurrently.

plugins/pluginmanager.go
petemoore: See above - this is needed.

plugins/pluginmanager.go
petemoore: No idea, this was just a refactor - but this preserves behaviour.

plugins/pluginmanager.go
walac: As we discussed in irc, just change parameter names to make gometalinter happy.

plugins/pluginmanager.go
jonasfj: yeah, golang doesn't capture variables from `for` loops properly... this is like javascript without `let`

petemoore:
@jonasfj @walac While applying Wander's linting changes, I spotted some more simplifications to make, so the code is even shorter now. Could you review again? Thanks!

plugins/pluginmanager.go
```diff
@@ -247,120 +225,59 @@ func (m *taskPluginManager) Prepare(c *runtime.TaskContext) error {
...
 	defer m.working.Set(false)
 
-	// Run method on plugins in parallel
-	errors := make(chan error)
-	for _, p := range m.taskPlugins {
-		go func(p TaskPlugin) { errors <- p.Prepare(c) }(p)
+	// Execute phase on plugins in parallel
+	errors := make([]error, len(m.taskPlugins))
+	var wg sync.WaitGroup
+	for i, j := range m.taskPlugins {
```
jonasfj: Nit but variables I or index and to or taskPlugin seems more descriptive.

walac: @petemoore is using different variable names to avoid variable shadowing...

PR #181, opened 2017-02-25
support rebooting after a configurable number of hours

djmitche:
I think this nicely works around all of the race conditions.

I don't really know how to test this -- advice is appreciated!

### DIFF
--- a/plugins/reboot/reboot.go
+++ b/plugins/reboot/reboot.go
@@ -1,26 +1,50 @@
-// This plugin allows a task to reboot the machine after it is finished.
-// It must add the boolean "reboot" payload attribute.
-
+// Package reboot is a plugin that supports rebooting the host running the worker.
+//
+// In some cases, especially running without a very tight
+// sandbox, this is desirable after specific tests.
+//
+// In other cases, reboots are useful after a configured uptime, to
+// cycle the host's configuration.
+//
+// To reboot after tasks, users must add the boolean "reboot" payload
+// attribute.
 package reboot
 
 import (
 	"fmt"
 
+	"sync"
+	"time"
+
 	schematypes "github.com/taskcluster/go-schematypes"
 	"github.com/taskcluster/taskcluster-worker/plugins"
+	"github.com/taskcluster/taskcluster-worker/runtime"
 )
 
 type plugin struct {
 	plugins.PluginBase
+	environment *runtime.Environment
+
+	// the command to reboot the host
+	rebootCommand []string
+
+	// while this is held for reading, there are tasks running
+	runningTasks sync.RWMutex
 }
 
 type payloadType struct {
-	Reboot bool `json:"reboot"`
+	RebootWhenDone bool `json:"reboot"`
+}
+
+type config struct {
+	RebootAfter   int      `json:"rebootAfter"`
+	RebootCommand []string `json:"rebootCommand"`
 }
 
 type taskPlugin struct {
 	plugins.TaskPluginBase
-	reboot bool
+	plugin         *plugin
+	rebootWhenDone bool
 }
 
 type pluginProvider struct {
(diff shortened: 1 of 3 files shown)


plugins/reboot/reboot.go
```diff
@@ -50,24 +102,51 @@ func (plugin) NewTaskPlugin(options plugins.TaskPluginOptions) (plugins.TaskPlug
...
+	pl.runningTasks.RLock()
 	return taskPlugin{
 		TaskPluginBase: plugins.TaskPluginBase{},
-		reboot:         p.Reboot,
+		plugin:         pl,
+		rebootWhenDone: p.RebootWhenDone,
 	}, nil
 }
 
 func (tp taskPlugin) Dispose() error {
```
jonasfj: We need to change the reciever type to `func (tp *taskPlugin) ...` in all the methods... As we no want to pass by reference...
I guess we should stop using the other pattern, just because of the footgun it introduces later...

Note: before `tp` was strictly read-only.

plugins/reboot/reboot.go
```diff
@@ -50,24 +102,51 @@ func (plugin) NewTaskPlugin(options plugins.TaskPluginOptions) (plugins.TaskPlug
...
+		return nil, engines.ErrContractViolation
+	}
+
+	plugin := &plugin{
+		PluginBase:   plugins.PluginBase{},
+		environment:  options.Environment,
+		runningTasks: sync.RWMutex{},
+	}
+
+	if c.RebootAfter > 0 {
```
jonasfj: Given that this is so far out into the future... why not just set `rebootAfter` as a `time.Time` timestamp...
Then do `if time.Now().Before(tp.plugin.rebootAfter) {  reboot(); }` we probably don't need a timer when we're waiting 96 hours... or so...

djmitche: Where would that conditional go?  *Something* needs to execute at that time.  I could make a timer that runs that check every hour but that would be a waste of time for the first 95 invocations. 

And we need to avoid race conditions between checking time since startup and checking for idleness.  The RWLock actually models that quite cleanly.

plugins/reboot/reboot.go
djmitche: I was curious why it was passed by value.  All the work I've been doing in Rust made me think the compiler would take care of alerting me if I was copying some value that shouldn't be copied (like, in this case, the RWLock).

plugins/reboot/reboot.go
jonasfj: oh, I was thinking we would run it after clean up following a task... But yeah, if we're 100% idle that never happens...

plugins/reboot/reboot.go
```diff
@@ -50,24 +102,51 @@ func (plugin) NewTaskPlugin(options plugins.TaskPluginOptions) (plugins.TaskPlug
...
+		// after rebootAfter hours, call the reboot method
+		go func() {
+			select {
+			case <-time.After(time.Duration(c.RebootAfter) * time.Hour):
+				plugin.environment.Log.Warning("Host rebootAfter has expired; scheduling reboot")
+				plugin.reboot()
+			}
+		}()
+	}
```
walac: hrm, shouldn't this go inside `pluginProvider`? I think `NewPlugin` can be called multiple times during worker life time, and only when a task is created.

@jonasfj you know better the ins and outs of task starting flow, am I right here?

djmitche: I don't think this is necessary.  The [docs](https://docs.taskcluster.net/reference/workers/taskcluster-worker/engine-plugin) say "[aside from testing], a single Plugin will be created at worker startup, and live until the worker stops/dies."

djmitche:
OK, massively updated.  This now allows the reboot command to be specified (since we need sudo now).  I tested this by setting the reboot command to something that doesn't actually reboot, and changing time.Hour to time.Second, then seeing that it runs that command and exits when run locally.  I didn't test the performance of the RWLock, but hopefully one of you can confirm my logic there.

plugins/reboot/reboot.go
jonasfj: Agree... Rebooting from a plugin will always be weird... As it's outside the scope of what plugins control.

petemoore:
The current design has some problems:

1) All plugins should be disposed of before the reboot command is issued, otherwise, the shutdown could be executed before other plugins have completed their `Dispose` call (remember the plugins are disposed of concurrently).
2) It feels cleaner for the worker to exit, and whatever executes the worker to issue a reboot command, rather than a reboot being issued from inside the worker. Ideally the worker would also not `os.Exit(0)`, but rather, all go routines would be cleanly terminated (e.g. signalled via closing of done channels, or using the standard library context package) and the main go routine would return. This provides a more controlled shutdown sequence, since all go routines then gracefully exit, and unit tests can be written that start a worker, with control returned after the worker completes (so go runtime does not exit in unit test).
3) Note, even though you block on the creation of a runtime task plugin starting, there is nothing stopping the other task plugins from initialising, and thus other task plugins would be created, and not given the chance to dispose. The mutex only takes care of blocking other reboot task plugins, not the other non-reboot task plugins.

In general, I think the reboot plugin should not trigger the shutdown execution directly, since the plugin is inherently tied to task execution workflow - it would be better for such a plugin to signal via a runtime package mechanism that the worker should reboot (or even better, gracefully stop) at the next available opportunity, and all components that are concerned with this runtime status would take care to respect this flag (such as the task manager) such that all go routines naturally complete when they recognise that this runtime flag has been set. This way, a flag is set, and then all routines naturally conclude in a graceful manner, and control is returned to whatever function started the worker. This makes writing unit tests also a lot easier, since the go runtime does not exit, but control is simply returned to the calling code (i.e. unit test). Such unit tests can of course operate with timeouts, in order to have hanging tests.

And before we land this PR, we should definitely add unit tests. :-)

Thanks @djmitche !

petemoore:
Note, there is already a method for gracefully terminating the worker [here](https://github.com/taskcluster/taskcluster-worker/blob/0dec87046cce27d8f3d316d9ebd330e1ca1e08ca/worker/worker.go#L163), so part of this work is already done. Probably this just needs to be exposed via a `runtime` package function, which the reboot plugin can call.

If we really want the reboot to be issued from inside the worker, rather than e.g. a calling script which starts the worker, maybe we should expose it as a configuration setting, so that tests don't actually issue a reboot command, but in production it can be called.

jonasfj:
I agree we want a shutdown manager thing in the environment... or a
requestRebootFunction in runtime.Environment...

But implementing the other side of that still lacks a little... and I'm not
sure how that control flow works. So for the reboot plugin to send a UNIX
signal to its own pid, while not ideal, works for now.
Please leave a comment to fix this using a generic mechanism.

But for now, let accept a hack here. I hope to get around to offer a
runtime API mechanism for requesting a reboot/shutdown.

Den 28. feb. 2017 2.12 AM skrev "Pete Moore" <notifications@github.com>:

plugins/reboot/reboot.go
walac: Well, my question was about multiple instances of a plugin. I have no objection of rebooting from the plugin. The only issue I see is that the the worker will always shutdown with a `SIGKILL` due to the `for { select {} }` statement.

djmitche:
I'm a little confused about the last two comments -- as written, the reboot command is configurable, and I've removed the self-signal (I didn't really see the point of it, and it is racy).

I think the bigger plan is to introduce some kind of lifecycle manager that can handle a cessation of task acceptance, graceful shutdown of all goroutines, and then exit.  That sounds like a big project, though!  I'm also worried about bugs in that manager: if just one goroutine's shutdown doesn't work correctly, we'll quickly have 300 hung workers that we're madly ssh'ing into to reboot.

walac:
`reboot` command will send SIGTERM to all process, wait some amount of time, and then send SIGKILL to all processes still alive. The reason for the self kill was to notify the worker it should shutdown sooner, giving it more time for cleanup. Anyway, `reboot` will always notify the worker to shutdown, the point is 1) we can't start a reboot when there tasks running 2) when a reboot is issued, we can't pull new tasks from the queue.

djmitche:
I suppose the risk is that a timed reboot will commence after a task
completes, but with this use of the lock, we will reboot immediately on
disposing the reboot TaskPlugin, which means other plugins may still be
shutting down.

I could add a delay to the TP disposal so that the other TP's can complete
first.  Yay for fixing issues with sleeps :(

On task startup, I bet the plugins are initialized *after* the work is
claimed.  Does that mean that, in the common case where a task completes
after the reboot time but there are more tasks in the queue, this
implementation will always leave a task run in the `claim-expired` state?
That would be pretty bad.

This is going to block roll-out on OS X, so we need something fairly
quick.  My understanding of this app is not such that I could write a
WorkerLifecycleManager in a reasonable amount of time.  Is there another,
simpler solution that won't leave claim-expired bombs all over?

petemoore:
> This is going to block roll-out on OS X, so we need something fairly quick. My understanding of this app is not such that I could write a WorkerLifecycleManager in a reasonable amount of time. Is there another, simpler solution that won't leave claim-expired bombs all over?

@djmitche If you can get something to call `worker.GracefulStop()`, it should prevent it claiming more tasks, and complete the existing tasks it is running.

petemoore:
or even `tm.GracefulStop()` where `tm` is the task manager... (if that proves easier to achieve in the short term)

walac:
> I could add a delay to the TP disposal so that the other TP's can complete first. Yay for fixing issues with sleeps :(

This is just an extension of what `reboot` already does, therefore I see no value on "just waiting some time more".

> This is going to block roll-out on OS X, so we need something fairly quick. My understanding of this app is not such that I could write a WorkerLifecycleManager in a reasonable amount of time. Is there another, simpler solution that won't leave claim-expired bombs all over?

Bear in mind that the cases we will do a reboot are those in which only one task run each time. Therefore, I am in favor of delaying a solution to when this actually become an issue. Just `panic` reboot plugin if `capacity > 1`.

djmitche:
Updated per discussion in irc

---

* use a RWLock to avoid race conditions in shutting down the worker
* add required plugin config
  * reboot delay (in hours)
  * reboot command (to allow use of sudo where necessary)

* update POSIX reboot technique:
  * Wait for the reboot command to complete and panic if it fails

Note that in my testing, this terminates the worker before it can 
execute the reboot command.

djmitche:
Those test failures all look intermittent to me..?

petemoore:
> Those test failures all look intermittent to me..?

Let's merge in the latest master, to pick up the taskcluster-github integration. We'll get more reliable results there, and cover more platforms.

petemoore:
@djmitche New failures are due to the new Monitor that replaces previous log/sentry implementation, that was added in #182. Sorry!

djmitche:
I'd like to get this landed - is there anything further to address, or is this a sufficient solution for now?

plugins/reboot/reboot.go
```diff
@@ -37,37 +60,111 @@ var payloadSchema = schematypes.Object{
...
+	pl.runningTasks.Lock()
+
+	// Reboot and/or die trying. The lock is still held, so no other tasks can
+	// start while this process plays out
+	pl.environment.Log.Warning("Initiating reboot")
+	initiateReboot(pl.rebootCommand)
+
+	// By this time, we should have stopped accepting new tasks, so it is safe
+	// to unlock and continue any task completion
+	pl.runningTasks.Unlock()
```
jonasfj: Prefer to use `defer p1.runningTasks.Unlock()` right after the `.Lock()` call..

plugins/reboot/reboot.go
```diff
@@ -37,37 +60,111 @@ var payloadSchema = schematypes.Object{
...
+		runningTasks:  sync.RWMutex{},
+	}
+
+	if c.RebootAfter > 0 {
+		// After rebootAfter hours, call the reboot method. If the lock is held by one or more
+		// running tasks, the reboot will not occur until they finish (but no further tasks will
+		// be able to start)
+		go func() {
+			select {
+			case <-time.After(time.Duration(c.RebootAfter) * time.Second):
```
jonasfj: Just use: https://golang.org/pkg/time/#AfterFunc
```go
time.AfterFunc(time.Duration(c.RebootAfter) * time.Second, func() {
  plugin.environment.Monitor.Warning("....")
  plugin.reboot()
})
```

PR #182, opened 2017-02-27
Monitor wrapper

jonasfj:
This is stuff isn't easy to write tests for... statsum has tests in the statsum client... 
but tests for the integration bits is hard, can't really be automated because they require scopes and we don't have a good API to test if something was submitted to sentry/statsum..

Maybe we can add some later, but for now this is a good start.

note: The mockMonitor and logging monitor are not particular important, most important is to review the monitor interface...

I'll start a PR removing logrus logging, instead we should pass around this the Monitor object, so we have logging, metrics and error reporting all in one :)

### DIFF
--- a/runtime/monitor.go
+++ b/runtime/monitor.go
@@ -119,6 +119,33 @@ type sentry struct {
 	auth       client.Auth
 }
 
+func (s *sentry) Client() (*raven.Client, error) {
+	s.m.Lock()
+	defer s.m.Unlock()
+
+	// Refresh sentry DSN if necessary
+	if s.expiration.Before(time.Now()) {
+		// Fetch DSN
+		res, err := s.auth.SentryDSN(s.project)
+		if err != nil {
+			return nil, err
+		}
+		// Create or update DSN for the client
+		if s.client == nil {
+			s.client, err = raven.New(res.Dsn.Secret)
+		} else {
+			err = s.client.SetDSN(res.Dsn.Secret)
+		}
+		if err != nil {
+			return nil, err
+		}
+		// Set expiration, so we remember to refresh
+		s.expiration = time.Time(res.Expires)
+	}
+
+	return s.client, nil
+}
+
 type monitor struct {
 	*statsum.Statsum
 	*logrus.Entry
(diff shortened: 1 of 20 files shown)


jonasfj:
Added tests, but they print to stderr, and I can't capture that in my test case... so I resort to manual testing for now... besides we have to manually test when sending to sentry anyways...

runtime/monitor.go
```diff
@@ -0,0 +1,334 @@
...
+	Errorf(string, ...interface{})
+	Panic(...interface{})
+	Panicln(...interface{})
+	Panicf(string, ...interface{})
+
+	// Create child monitor with given tags (tags don't apply to statsum)
+	WithTags(tags map[string]string) Monitor
+	WithTag(key, value string) Monitor
+	// Create child monitor with given prefix (prefix applies to everything)
+	WithPrefix(prefix string) Monitor
```
jonasfj: `WithTags/WithTag` is just convenience... I guess we could have  a `MonitorWrapper` struct that implemented one using the other, but it just seems overkill. IMO, we can always refactor the implementation as long as `Monitor` is an interface, so I'm less worried about it. I doubt we'll need more than implementations that I've already added.

We need all the log methods for compatibility with interfaces defined in other packages.
I'm pretty sure I'm not the only one who  has done stuff like: https://godoc.org/github.com/taskcluster/go-got#Logger

runtime/monitor.go
```diff
@@ -0,0 +1,334 @@
...
+			m.sentry.client, rerr = raven.New(res.Dsn.Secret)
+		} else {
+			rerr = m.sentry.client.SetDSN(res.Dsn.Secret)
+		}
+		if rerr != nil {
+			m.Error("Obtained invalid sentry DSN, error: ", rerr)
+			m.Error("Failed to send error: ", err)
+			m.sentry.m.Unlock()
+			return
+		}
```
jonasfj: No, it can't. It's subtle but `m.sentry.client, rerr = raven.New(res.Dsn.Secret)` also sets `rerr`

runtime/monitor.go
```diff
@@ -0,0 +1,334 @@
...
+		// Set expiration, so we remember to refresh
+		m.sentry.expiration = time.Time(res.Expires)
+	}
+	m.sentry.m.Unlock()
```
jonasfj: Yeah, in this case the locking isn't simple... I think there might even be a small race condition.. I'll fix that..

runtime/monitor.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// Send packet
+	_, done := m.sentry.client.Capture(packet, tags)
+	<-done
+}
+
+func (m *monitor) WithTags(tags map[string]string) Monitor {
+	// Merge tags from monitor and tags
+	allTags := make(map[string]string, len(tags))
```
jonasfj: Yes, and no, you can also overwrite tags... I guess we can safely allocate too much...

runtime/monitor.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+}
+
+type loggingMonitor struct {
+	*logrus.Entry
+	prefix string
+}
+
+// NewLoggingMonitor creates a monitor that just logs everything. This won't
+// attempt to to send anything to sentry or statsum.
```
jonasfj: thanks

runtime/monitor.go
jonasfj: So I think you have a fair point here...
We should have:
 1) `type Monitor interface {...}` as an interface
 2) `type monitorBase struct {monitorImpl}`  implements `Monitor`
 3) `type monitorImpl interface {...}` which specifies logic to be implemented by a monitor provider logic... 

If you really want I can refactor... In fact I'm tempted because it's nice :)
On the other hand it seems a bit overkill.. So long as we have `Monitor` as an interface we can always do things to clean up this stuff later.

jonasfj:
I give up waiting travis... I'll fix tests if broken... Need to merge to reduce number of branches.

jonasfj:
It's impossible to refactor when other tests are broken... So had to land it, inorder to refactor on topof this... on top of #183

PR #185, opened 2017-03-01
Remove env log

jonasfj:
This doesn't do much more than fix some bugs and move from `logrus.Entry` to `Monitor`..

Please give a quick skim and land asap... This is liable to bitrot very quickly.

PR #179, opened 2017-02-23
Run `gometalinter`

petemoore:
Requested by @walac.

Looks to me like a lot of noise - do we want to:

1) reduce the number of tools we run?
2) fix the reported issues before merging this PR?

Anyway, I'm happy to start this off, but don't have a lot of motivation to walk through all the reported issues. If others want to join in fixing / reducing the listed problems, feel free to push to the `gometalinter` branch to get added to this PR. :-)

PR #187, opened 2017-03-01
Fixed govendor, I had github.com/statsum/vendor populated locally, so…

jonasfj:
… I had to delete it, and recreate govendor thing

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -4,332 +4,318 @@
 	"package": [
 		{
 			"checksumSHA1": "2avCvSn0XonfApKKuTMyiQmk7s0=",
-			"comment": "v0.10.0-38-g3ec0642",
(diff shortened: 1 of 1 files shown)


jonasfj:
This works... I tested it locally in an empty docker container.. .`govendor sync` works now.

jonasfj:
Please land this to fix master...

jonasfj:
Travis still seems largely offline... probably due to the S3 issue today.

jonasfj:
So failures now seems related to `go fmt`, which gets fixed in #179.

I think we should just merge #179, #185, #186, #187, then fix conflicts and bug and land it...
@petemoore, if you don't get around to do that tomorrow I will. Sorry for creating a mess.

hopefully travis will be reliable tomorrow... night night...

PR #188, opened 2017-03-01
Fix that one bug caused by linter fixes

jonasfj:
Seems pretty cool that linter fixes only caused one bug :)

This should fix it.

### DIFF
--- a/commands/qemu-guest-tools/guesttools_test.go
+++ b/commands/qemu-guest-tools/guesttools_test.go
@@ -26,6 +26,7 @@ func TestGuestToolsSuccess(t *testing.T) {
 	}
 	environment := &runtime.Environment{
 		TemporaryStorage: storage,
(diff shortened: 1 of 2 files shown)


commands/qemu-guest-tools/guesttools_test.go
```diff
@@ -112,7 +114,7 @@ func TestGuestToolsFailed(t *testing.T) {
 	if !resolved {
 		t.Error("Expected the metadata to have resolved the task")
 	}
-	if !result {
+	if result {
```
jonasfj: See change here:
https://github.com/taskcluster/taskcluster-worker/commit/2d16de913fd82859354c6f72b3e5ef0743bdc847#diff-369ced5ee39c3faf552d7308466229ccL115

jonasfj:
Travis appears to still have issues...

PR #191, opened 2017-03-01
Bug 1322256: hostcredentials config transform

djmitche:


djmitche:
Looks like the metalinter failed in travis..

In TaskCluster,
```
--- FAIL: TestGuestToolsFailed (0.01s)
	guesttools_test.go:116: Expected the metadata to get failed result
```
so I think the CI failures can be ignored.

config/hostcredentials/transform.go
```diff
@@ -0,0 +1,76 @@
...
+			if err != nil {
+				log.Printf("decoding JSON from server: %s; continuing to next server", err)
+				continue
+			}
+
+			break FetchLoop
+		}
+
+		log.Printf("list of servers exhausted; sleeping before starting again")
+		time.Sleep(60 * time.Second)
```
petemoore: I wonder if we should have an [exponential backoff](https://github.com/cenkalti/backoff) here? Not a deal breaker though!

config/hostcredentials/transform.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/taskcluster/taskcluster-worker/config"
+)
+
+type provider struct{}
+
+func init() {
+	config.Register("hostcredentials", provider{})
+}
+
+func (provider) Transform(cfg map[string]interface{}) error {
```
petemoore: `cfg map[string]tcclient.Credentials` ?

petemoore:
> Looks like the metalinter failed in travis..

Jonas fixed this, I'll push commits from master.

config/hostcredentials/transform.go
```diff
@@ -0,0 +1,76 @@
...
+			if err != nil {
+				log.Printf("result: %s; continuing to next server", err)
+				continue
+			}
+
+			err = json.Unmarshal(body, &creds)
+			if err != nil {
+				log.Printf("decoding JSON from server: %s; continuing to next server", err)
+				continue
+			}
```
petemoore: Should we also validate the returned body against a json schema here, to make sure e.g. it isn't an error page etc, but we really get valid credentials in accordance with what we require?

We could use the [schematypes](https://github.com/taskcluster/go-schematypes) library for this, or embed the jsonschema as a string in the source code - but we're using schematypes library elsewhere in taskcluster-worker, so that would be more consistent.

config/hostcredentials/transform.go
djmitche: The service isn't very high-load on the backend, either -- getting hit with 1000 of these over a one-minute period after a downtime will not cause issues.

config/hostcredentials/transform.go
djmitche: They end up being validated later, I believe, as part of the configuration.  At least, I assume there's a schema for the config? Jonas?

petemoore: OK, I guess that is true, although if we capture it here because a server has gone awol, it will continue retrying, right? I think if it fails later, there is no recovery, or maybe we'll just go boot-looping?

Also, I guess we should probably send out monitoring alerts if we get bad content back. Should we use the new Monitor code to make sure we create sentry alerts if we get bad responses?

@jonasfj ?

jonasfj: Let's not valid server responses on the client... Ever...

jonasfj: Sentry integration is not available at this stage.... You can't report with credentials.... Either :)

petemoore: > Let's not valid server responses on the client... Ever...

@jonasfj Which of these do you mean?

* Let's not validate server responses on the client... Ever...
* Let's not allow invalid server responses on the client... Ever...

djmitche: the first one :)

djmitche: I think we're always going to have to have a way of monitoring for instances with no taskcluster-worker running.  Boot-looping isn't a great solution unless we can be sure it has a long period (at least 30 minutes).  Otherwise recovering bad hosts is just too hard.

This code does check the respose code, and only proceeds on 200 OK, so that's probably sufficient.  Responses with a 200 but the wrong data are probably going to be HTML, so they'll fail the JSON parse.

petemoore: I can't imagine @jonasfj means the first one, since this is the reason we publish schemas, and we consistently validate inputs/outputs in all our systems, also between our own systems that talk with each other. I think gating inputs/outputs helps avoid nondeterministic behaviour, as code that accepts inputs will make assumptions about it, so if it hasn't validated these assumptions, it can lead to non-deterministic code or even crashes. I'm sure he meant the second option. :)

petemoore: Consider the other system is swapped out for a different system, or consider that the other system is on a different version of the software that provides differently structured data. Imagine, a bug gets introduced to that other system. I think it is just good hygiene, that if you have requirements on inputs, that you validate those requirements, to avoid you operating on this data according to axioms that don't apply. It also helps when an exception is thrown, you have some data about where the problem got introduced, as you have gated the inputs and outputs - you can isolate it down to a single system.

jonasfj: We don't validate responses on clients because we wouldn't be able to add additional properties on the server side in the future, if we did... non of the tc-clients do schema validation. They could do outgoing schema validation if they wanted to, but we do it server side anyways.

Or maybe I'm misunderstanding what you are talking about.

config/hostcredentials/transform.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	var creds interface{}
+
+FetchLoop:
+	for {
+		for _, url := range urls {
+			log.Printf("Trying host-secrets server %s...", url)
+
+			resp, err := http.Get(url.(string))
```
jonasfj: Do use `go-got` for simple HTTP stuff... disclaimer I wrote go-got :)

```go
g := got.New()
res, err := g.Get(url.(string)).Send()
if err != nil {
  ...
}
if err = json.Unmarshal(res.Body, &creds); err != nil {
   ...
}
```

The default go-got will give you:
- Max 5 retries,
- Exponential back-off: 100, 200, 400, 800 ms with 25% randomization
- Retry any 5xx error,
- ErrResponseTooLarge if response body is larger than 10 MiB,
- Timeout at 30 seconds,

If you want to do it in a loop, that's fine... The 5 retries is just quicker than starting with sleep 60s.
You can also change the backoff policy if you don't like it.

The important part is:
- ErrResponseTooLarge if response body is larger than 10 MiB,
- Timeout at 30 seconds,

This provides additional robustness, in case for some reason the HTTP connection just decides to sit there doing nothing... or if for some reason response tries to run the computer out of memory... That won't happen...
Doing this things right with builtin http in golang is tedious, but the extra robustness is lovely.

config/hostcredentials/transform.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (provider) Transform(cfg map[string]interface{}) error {
+	urls, ok := cfg["hostCredentialsUrls"].([]interface{})
+	if !ok {
+		return errors.New(
+			"Expected 'hostCredentialsUrls' property to hold a list of URLs for host-secrets services")
+	}
+
+	var creds interface{}
```
jonasfj: Ah, I see what pmoore means with validation...

Yeah I would do:
```go
var creds struct {
  ClientID string `json:"clientId"`
  AccessToken string `json:"accessToken"`
  Certificate string `json:"certificate"`
}
```

even if it means that we later have to do:
```go
let retval := map[string]string{
  "clientId": creds.ClientID,
  "accessToken": creds.AccessToken,
}
if creds.Certificate != "" {
  retval["certificate"] = creds.Certificate
}
return retval
```

This way:
A) We ensure that we don't inject config values we don't want from a buggy server. Which I think is what @petemoore wanted to guard against (it's not JSON schema validation, but does provide some minimum of sanity).
B) We always return a clone of the cache...

config/hostcredentials/transform_test.go
```diff
@@ -0,0 +1,61 @@
...
+      "accessToken": "atat",
+      "certificate": "{cert}"
+    }`))
+	}))
+	defer good.Close()
+
+	configtest.Case{
+		Transform: "hostcredentials",
+		Input: map[string]interface{}{
+			"creds": map[string]interface{}{"$hostcredentials": true},
```
djmitche: Good thinking, thanks :)

config/hostcredentials/transform.go
djmitche: Ah, thanks, I was looking for timeout support in net/http and didn't find any.

config/hostcredentials/transform.go
```diff
@@ -0,0 +1,76 @@
...
+
+			break FetchLoop
+		}
+
+		log.Printf("list of servers exhausted; sleeping before starting again")
+		time.Sleep(60 * time.Second)
+	}
+
+	return config.ReplaceObjects(cfg, "hostcredentials", func(val map[string]interface{}) (interface{}, error) {
+		return creds, nil
```
djmitche: It's a one-line function.. how much lazier can I be?  I don't understand this comment..

jonasfj: lazy load it...
I would do a global cache from url -> credentials...

And only insert stuff in cache when the `$host-credentials` construct is being replaced.

config/hostcredentials/transform.go
```diff
@@ -0,0 +1,80 @@
...
+			for _, url := range urls {
+				log.Printf("Trying host-secrets server %s...", url)
+
+				resp, err := g.Get(url).Send()
+				if err != nil {
+					log.Printf("result: %s; continuing to next server", err)
+					continue
+				}
+
+				if resp.StatusCode != 200 {
```
jonasfj: Got returns an error if it's not a `2xx` response... maybe that's enough:
https://github.com/taskcluster/go-got/blob/master/got.go#L213-L214

config/hostcredentials/transform.go
```diff
@@ -0,0 +1,80 @@
...
+					continue
+				}
+
+				err = json.Unmarshal(resp.Body, &creds)
+				if err != nil {
+					log.Printf("decoding JSON from server: %s; continuing to next server", err)
+					continue
+				}
+
+				break FetchLoop
```
jonasfj: I would just do a `return creds` here... maybe defined `type credentials struct {...} ` at top-level...

But strive to avoid the nested loop breaking magic... In fact I would probably just create a `fetchCredentials(urls []string) (credentials, error)` function...

Btw, are you sure you wish to loop forever? it's okay with me, I just want to be sure... you can return an error which will be printed and the worker will crash whatever that means.

jonasfj: Also a function like `fetchCredentials` could be unit tested, although there's no need for that here...

djmitche: I do want to loop forever.  Crashing the worker means either someone needs to login and start it back up (on 100's of machines) or it automatically reboots and fails again, and someone has to try to login and fix it before it automatically reboots.  in either case, buildduty is only going to do that once or twice before it becomes our responsibility..

jonasfj: that makes sense.

config/hostcredentials/transform.go
```diff
@@ -0,0 +1,73 @@
...
+				err = json.Unmarshal(resp.Body, &creds)
+				if err != nil {
+					log.Printf("decoding JSON from server: %s; continuing to next server", err)
+					continue
+				}
+
+				retval := map[string]interface{}{
+					"clientId":    creds.ClientId,
+					"accessToken": creds.AccessToken,
+				}
```
petemoore: I still think a json schema would have been better, since there is nothing to guarantee that `clientId` and `accessToken` are included, and we will also this will silently drop any additional properties, rather than fail if additional properties were included.

I won't block on it, but I think my original suggestion was better. That said, it is true json schema validation kicks in later when the entire config is validated, so it isn't the end of the world.

djmitche:
That's not an expected failure mode, and anyway as Jonas has described, as a rule we do not apply schema validation on the client to server respones -- it's already applied on the server.

djmitche:
I fixed the lint errors from the TC builds.  AIUI travis failures are expected.

flodolo:
FTI automation just commented in the wrong bug
https://bugzilla.mozilla.org/show_bug.cgi?id=1322256

petemoore:
thanks @flodolo  I've updated the PR title - sorry about that

PR #193, opened 2017-03-02
Expose artifacts via taskContext

petemoore:
@jonasfj @walac This builds on top of #192 and exposes the artifacts included in upload requests via the task context.

### DIFF
--- a/runtime/artifact.go
+++ b/runtime/artifact.go
@@ -40,14 +40,57 @@ type RedirectArtifact struct {
 	Expires  tcclient.Time
 }
 
+// S3Artifacts returns a copy of all s3 artifact upload requests *made so far*
+// for the given task context. Note it does not guarantee that the upload was
+// successful. Typically this should be called during the Finished task phase,
+// since plugins will typically upload artifacts in the Stopped task phase.
+func (context *TaskContext) S3Artifacts() []queue.S3ArtifactRequest {
+	context.artifactMutex.RLock()
+	defer context.artifactMutex.Unlock()
+	s3ArtifactsCopy := make([]queue.S3ArtifactRequest, len(context.s3Artifacts))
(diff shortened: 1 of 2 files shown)


runtime/artifact.go
```diff
@@ -40,14 +40,57 @@ type RedirectArtifact struct {
 	Expires  tcclient.Time
 }
 
+// S3Artifacts returns a copy of all s3 artifact upload requests *made so far*
+// for the given task context. Note it does not guarantee that the upload was
```
walac: nit: "Note:"

runtime/artifact.go
```diff
@@ -40,14 +40,57 @@ type RedirectArtifact struct {
...
+func (context *TaskContext) S3Artifacts() []queue.S3ArtifactRequest {
+	context.artifactMutex.RLock()
+	defer context.artifactMutex.Unlock()
+	s3ArtifactsCopy := make([]queue.S3ArtifactRequest, len(context.s3Artifacts))
+	copy(s3ArtifactsCopy, context.s3Artifacts)
+	return s3ArtifactsCopy
+}
+
+// ErrorArtifacts returns a copy of all error artifact upload requests *made so
+// far* for the given task context. Note it does not guarantee that the upload
```
walac: nit: "Note:"

petemoore:
I think we decided not to implement this, since plugins would race each other. Instead I think we decided that uploaded artifacts would only be exposed to plugins at Finished task phase.

@jonasfj do you agree? If so, we can close this PR without merging.

jonasfj:
Agree, we only expose​ list of uploaded artifacts in finished, when the log
has been closed... In fact we can use the same method closeLog to make it
available..

Den 13. mar. 2017 4.08 AM skrev "Pete Moore" <notifications@github.com>:

PR #194, opened 2017-03-03
Refactor errors

jonasfj:
Moving generic errors into `runtime` fixing some cast bugs for `MalformedPayloadError`...

I imagine the following error cases going forward:
```
Error cases: (how the worker will handle these errors)
 - engines.MalformedPayloadError
     Message is printed to task log.
     Task resolved exception with malformed-payload.
 - unknown error
     1) Message forwarded to sentry + syslog, with context tags from monitor.
     2) IncidentId is printed to task log.
     3) Task resolved exception with internal-error.
     4) Worker resets/retarts/destroys itself.
 - engines.ErrFatalInternalError
     The event is logged as Error is syslog, steps (1-2) is assumed done by
     plugin/engine implementor. This is useful, if you have an internal error
     that you feel comfortable writing to task log.
     Task resolved exception with internal-error.
     Worker resets/retarts/destroys itself.
 - engines.ErrNonFatalInternalError
     The event is logged as Warn is syslog, steps (1-2) is assumed done by
     plugin/engine implementor. This is useful, if you have an internal error
     that you feel comfortable isn't fatal for the worker.
     Task resolved exception with internal-error.
     (if worker sees many of these, it will reset the machine)
 - engines|plugins.ErrXXX, context.Canceled,
     Error indicates cancelation, unexpected state or unsupported feature,
     these cases must be handled by caller.
```

@walac  ^ you might want to read that too...

-----
Note: we can introduce more errors in the future... And have more special cases for how they should be handled... The reboot plugin might return a `ErrRebootRequired` from `TaskPlugin.Dispose()` who knows...

But I think this is a good start that'll cover most of our initial use-cases reasonably. Later we might find we have other use-cases or want to optimize the flow..

Example: for `ErrNonFatalInternalError` is couldn't talk to metadata service inside the QEMU virtual machine. This would normally be because someone killed the process inside the VM, which I can't do anything about. But it's also not something I want to cause a worker shutdown. It's just an unfortunate turn of events for the given task. Or maybe I learn later that it ought to be an `ErrResourceNotFound` instead, since task author can produce it by force killing the metadata service.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,4 +53,5 @@ lint:
 	go get github.com/alecthomas/gometalinter
 	gometalinter --install
 	# not enabled: aligncheck, deadcode, dupl, errcheck, gas, gocyclo, structcheck, unused, varcheck
-	gometalinter --deadline=10m --line-length=180 --vendor --vendored-linters --disable-all --enable=goconst --enable=gofmt --enable=goimports --enable=golint --enable=gosimple --enable=gotype --enable=ineffassign --enable=interfacer --enable=lll --enable=misspell --enable=staticcheck --enable=test --enable=testify --enable=unconvert --enable=vet --enable=vetshadow ./...
+	# Disabled: testify, test (these two show test errors, hence, they run tests)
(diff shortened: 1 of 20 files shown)


jonasfj:
Looks like the `gometalinter` is timing out... I suspect that `test` and `testify` runs all go tests.. which explains it...

PR #195, opened 2017-03-03
Extended artifact test

petemoore:
Added some checks to the artifact test, and specify charset in mimetype now.

### DIFF
--- a/plugins/livelog/livelog.go
+++ b/plugins/livelog/livelog.go
@@ -79,7 +79,7 @@ func (tp *taskPlugin) Prepare(context *runtime.TaskContext) error {
 
 	err := tp.context.CreateRedirectArtifact(runtime.RedirectArtifact{
 		Name:     "public/logs/live.log",
-		Mimetype: "text/plain",
+		Mimetype: "text/plain; charset=utf-8",
 		URL:      tp.url,
 		Expires:  tp.context.TaskInfo.Expires,
 	})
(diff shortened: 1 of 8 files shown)


plugins/livelog/livelog.go
```diff
@@ -79,7 +79,7 @@ func (tp *taskPlugin) Prepare(context *runtime.TaskContext) error {
 
 	err := tp.context.CreateRedirectArtifact(runtime.RedirectArtifact{
 		Name:     "public/logs/live.log",
-		Mimetype: "text/plain",
+		Mimetype: "text/plain; charset=utf-8",
```
jonasfj: Probably best to make this a constant...
`var mimePlainText = "text/plain; charset=utf-8"`

And then reuse that everywhere... Just so that we only have it once this package...

test/helper_non-windows_test.go
```diff
@@ -0,0 +1,10 @@
+// +build !windows
```
jonasfj: file name should be `_unix_test.go` check what pattern we use in `engine/native/system/` package.

petemoore: bsd/darwin is not unix - so I suggest the other naming should be changed, not this. !windows _means_ non-windows - they are equivalent. unix does not mean non-windows - those are not the only two variants.

plugins/livelog/livelog.go
petemoore: yeah maybe - let's do in a different PR

test/helper_non-windows_test.go
jonasfj: > bsd/darwin is not unix

@petemoore, afaik the BSDs including OS X actually derive from the original unix :)
OS X has certainly been unix certified at times... Regardless, they are all consider unix-like, just like linux (which also isn't unix).
We could also call it `_posix`, but I don't enjoy dashes in golang filenames, it just feels wrong...

PR #196, opened 2017-03-03
Remove ErrContractViolation in favor of panic()

jonasfj:
I'm going to refactor the code that calls plugins and engines such that:
 1) Payload/Config is always valid
 2) Panics are caught and:
    1) reported to sentry,
    2) incidentId printed to task-log
    3) error + incident logged to syslog.

So panicing includes more information about the location than anything else... 
And honestly, we're not going to see these happen.. So we just reduce the amount of boilerplate code.


jonasfj:
@walac, yeah the idea is that we'll catch panics higher up...
We should already be validating schemas at a higher level, so that will never happen, because invalid payload won't be passed down into the plugin/engine.

PR #197, opened 2017-03-06
Longer timeout for travis builds

petemoore:
I noticed [timeouts on travis](https://travis-ci.org/taskcluster/taskcluster-worker/builds/207498899) with no output from `make lint` for 10 mins - seeing if bumping the timeout to 20 mins helps...

See [travis advice](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received).


jonasfj:
@petemoore is this still relevant or did we fix it?

petemoore:
@jonasfj I don't remember!

jonasfj:
travis is passing on master so let's drop this...

PR #199, opened 2017-03-07
Added CapturePanic function to runtime.Monitor

jonasfj:
Logic for capturing panics... I did manual testing and verified that the stack traces looks right.


jonasfj:
I retried the tests... it's just our linters that keeps timing out.. :)

petemoore:
Nice work. 👍

PR #201, opened 2017-03-08
New tests

petemoore:
Ported over some generic worker tests...


test/helper_non-windows_test.go
```diff
@@ -2,9 +2,23 @@
 
 package test
```
jonasfj: please rename this file `helper_unix_test.go`

jonasfj (approved):
This is really awesome!! :)

Can we put the tests that currently fail behind a build constraint like `unfixed_tests` or something like that?

We need to not block CI further, as it slows everything down...

jonasfj:
Note:
As possible future refactoring maybe we could do: `test/integration/missingartifact.go`
```js
package integration

type MissingArtifactTestCase struct {
  Config string
  Payload string
}


func (c MissingArtifactTestCase) Run(t *testing.T) {
  t.Run("MissingArtifactTest", func(t *testing.T) {
    ...
  })
}
```

Then we can have a package like: `test/mock/integration.go`
```go
func TestIntegrationWithMockEngine(t *testing.T) {
  integration.MissingArtifactTestCase{
    Config: `{ /* config for worker with mock engine as JSON */ }`,
    Payload: `{ /* task playload as JSON */ }`,
  }.Run(t)

  // more test cases...
}
```

petemoore:
@jonasfj almost all of these tests are failing - can you take a look and see if there is anything obviously wrong? We seem to be consistently getting `internal-error`, so I think they may have uncovered some buggy code.

petemoore:
The current failures I see are:

* TestDirectoryArtifacts
* TestMissingFileArtifact
* TestMissingDirectoryArtifact
* TestFileArtifactIsDirectory
* TestDirectoryArtifactIsFile
* TestMissingArtifactFailsTest
* TestAbortAfterMaxRunTime
* TestResolveResolvedTask

TestResolveResolvedTask fails because TASK_ID is not set in environment, and this should be the task ID of the running task. Most of the other exceptions seem to be the worker hitting an `internal-error` or resolving a task as `exception` when it should be `failure`.

jonasfj:
@petemoore, so I've started on a `FakeQueue` approach to integration testing...
Meaning I want to do a fake queue implementation in golang so we can define test cases and run them against production queue or fake queue... 

I'm investing a few days this week, and if it's not done by then I'll drop the idea. It's a lot of engineering, but if we get it working we'll have awesome tests.

jonasfj:
Okay, I played with these and got them running, but the setup suffers from quite a few issues:
 * Methods are exported which shouldn't, if we wanted a generic test framework, then the framework and the actual tests should be in two different packages.
 * Linter issues
 * Artifact comparison relies on string comparison... and ordering... this is likely to be buggy it's definitely hard to debug.
 * Artifacts doesn't account of task logs
 * Tests doesn't check that reasonable error messages are written to task logs

This does test some things like content-type and other things better than the `workertest` package I've proposed, but I think we're better of expanding the coverage it provides than fixing these tests.

jonasfj:
@petemoore  if curious config fixes is in the `native-engine-fix-new-tests` branch.

For now I'm giving up and adopting the test cases to the new framework instead... and focusing on improving the assertions there... stealing all the good ideas for test cases and assertions that you've made here...

jonasfj:
We've build some decent integration tests that can run with and without taskcluster credentials... I think we should close this and not embed integration test from generic-worker.

petemoore:
These tests found genuine problems, so we should make sure they pass before closing this - otherwise we have no guarantees that the new mock testing covers the same test cases that the working integration tests covered - so let's at least verify the issues are really fixed before we close this.

jonasfj:
> These tests found genuine problems, so we should make sure they pass before closing this

Yes, I think those are fixed.. but we can't be sure without running the same tests... So we would have merge in from master.. If you are up for this feel free... I'm prefer to not have two integration test frameworks, but if you want to run the tests and prove that they work that's fine...

To be fair, I'll warn that config changed, so running the tests again might be non-simple.

petemoore:
@jonasfj This was pretty straightforward to fix, but this depends on #164 and I can't see where you've moved this config setting to - can you confirm how this is configured now?

PR #202, opened 2017-03-09
Error handling in PluginManager

jonasfj:
@petemoore, heavily inspired by the stuff you did...
I when with an integer instead of a `taskPlugin` because it allows me to also use this for initialization...


jonasfj:
I'm hoping the linter tests will pass on master where I fixed those...

petemoore:
I guess not! :-)

PR #203, opened 2017-03-09
[bugreport] Disabled broken test, we should not put a file in testdata/\n\nWe sho…

jonasfj:
…uld create the file from the command in the payload...

Basically, artifact paths isn't global paths... they are always engine specific... for native they are relative to $HOME, and always required to be under $HOME, so you don't go around exporting random..

This makes a lot of sense when native engine creates a new user account per task. But we could consider disabling it, if we don't do that... Still it's not a crazy limitation... This way you don't need to know what the location of the current working folder is... which is pretty nice.

Besides what business do you have exporting arbitrary files from around the system?

### DIFF
--- a/test/artifacts_test.go
+++ b/test/artifacts_test.go
@@ -20,6 +20,9 @@ import (
 )
 
 func TestUpload(t *testing.T) {
+	t.Skip(`Test case is broken because native engine runs with $HOME as cwd,
+		and extracts artifacts relative to cwd. In fact it doesn't even allow
+		export of artifacts outside of cwd == $HOME.`)
 	if runtime.GOOS == "windows" {
 		t.Skip("Currently not supported on Windows")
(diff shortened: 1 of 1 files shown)


jonasfj:
Arguably, we should fix this test... this PR is just a very temporary thing. Consider it a bug report :)

Oh, and the real problem here is also lack of usage docs... for the native engine.. We can start thinking about how make something that can generate that. I'll talk to @imbstack tomorrow, perhaps we can have a command that can generate a taskcluster-lib-docs compatible tarball with docs specific to the config file given...


Ps. I almost fixed the linter stuff with `-j4` :)

jonasfj:
Can't figure out the last linter bug... it makes no sense to me... and does reproduce, so that's messed up.. tsk tsk...

Issue #208, opened 2017-03-17
Fix README.md: links in the top are nit parsed by Markdown


plutov:

petemoore:
@RaviTezu many thanks! Fixed in #207.

PR #209, opened 2017-03-17
Fixing the README.md to show the build status, godoc, coverage and license badges

RaviTezu:
Fixes: https://github.com/taskcluster/taskcluster-worker/issues/208
Fixing the README.md to show the build status, godoc, coverage and license badges


petemoore:
Hi @RaviTezu many thanks for both the issue and the PR. There was another PR for the same issue (#207) which won the race to get landed, and makes the same fix - but thanks very much for taking the time to raise the PR and the issue. :+1:

PR #210, opened 2017-03-17
[WIP] Taskrun

jonasfj:
@petemoore, please help out with a preliminary review...

My next step is to refactor the worker part... so use `queue.ClaimWork`, but also split up the config into smaller pieces making it easier to hack on, and hopefully employ the acyclicloader.. And finally adding a LifeCyclePolicy abstraction so we can have different policies for EC2, localhost and datacenter.

Anyways, please take a look at the `taskrun` package... It'll rename when I remove the old worker package.


jonasfj:
Notable things to check for is:
 * error handling and reporting
 * race conditions

petemoore:
It looks like good test coverage, so I'd say let's just land, and fix if we hit problems.

jonasfj:

PR #213, opened 2017-03-20
Fix all the linter errors

jonasfj:
This is what we have to do when the linter moves faster than us...

IMO, this isn't so bad, I count one bug that was caught here...


engines/qemu/resultset.go
```diff
@@ -48,7 +48,7 @@ func (r *resultSet) ExtractFolder(path string, handler engines.FileHandler) erro
 		}
 		// If guest uses backslashes our input paths should have that, but the ones
 		// we return should be intepreted as names.
-		p = strings.Replace(p[len(path):], "\\", "/", 0)
+		p = strings.Replace(p[len(path):], "\\", "/", -1)
```
jonasfj: Check it out the linters actually found a bug!

PR #214, opened 2017-03-21
Fix travis and makefile

jonasfj:


jonasfj:
Travis failed to due to linter timeouts...

jonasfj:
lol, so gometalinter ran out of memory on linux :)

jonasfj:
@petemoore, I think this will solve a lot of issues for us..

PR #216, opened 2017-03-21
Life-Cycle Policy extension register

jonasfj:
@petemoore, I don't know if we want this as a new top-level folder... It doesn't seem like it belongs a subpackage of runtime, since it depends on runtime... and it's not really worker a subpackage of the worker/ package either...

Btw, placed the `Stoppable` interface in runtime as `runtime.Stoppable` my thinking is that we can add `Stoppable` to `runtime.Environment` that way any engine/plugin can call `StopNow()` or `StopGracefully()` giving us the ability to write the `reboot` plugin the right way :)

@walac, if curious I'm hoping this will be abstraction we hide EC2 shutdown logic behind... ie. we'll implement an `LifeCyclePolicy` for EC2, another one for datacenter, a third one for say packet.net, etc... Similar to how we pick an engine in config we just pick a life-cycle policy too.

### DIFF
--- a/lifecyclepolicy/stoppableonce.go
+++ b/lifecyclepolicy/stoppableonce.go
@@ -1,0 +1,58 @@
+package lifecyclepolicy
+
+import (
+	"sync"
+
+	"github.com/taskcluster/taskcluster-worker/runtime"
+)
+
+// StoppableOnce is a wrapper that ensures we only call StopGracefully and
+// StopNow once and never call StopGracefully after StopNow.
+//
+// There is never any harm in wrapping with this, it merely limits excessive
+// calls to StopNow() and StopGracefully(). Please note that Stoppable.StopNow()
+// may still be invoked after Stoppable.StopGracefully(), it can even be invoked
+// concurrently.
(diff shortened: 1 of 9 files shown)


lifecyclepolicy/stoppableonce.go
```diff
@@ -0,0 +1,58 @@
...
+type StoppableOnce struct {
+	Stoppable          runtime.Stoppable
+	m                  sync.Mutex
+	stoppingNow        chan struct{}
+	stoppingGracefully chan struct{}
+}
+
+// StopGracefully calls StopGracefully() on the s.Stoppable, if neither
+// StopGracefully() or StopNow() have been called.
+func (s *StoppableOnce) StopGracefully() {
```
petemoore: Can we make use of [sync.Once](https://golang.org/pkg/sync/#Once) here instead?

jonasfj: Then how do I get `StopNow` to overwrite `StopGracefully` but not the other way around?

jonasfj:

PR #217, opened 2017-03-22
Worker2

jonasfj:
can I get a 30% review on this..

I think I could replace the `atomics.Barrier` primitive with `context.Context` it just feels a bit abusive..
Note: I'm adding functions concurrency primitivies in the atomics package to try and avoid making the code too complicated elsewhere...

hmm.. the more I play with it in my mind... I'm starting to think I have to create a non-generic concurrency primitive that does what `activeTasks` and `idleTimer` tries to do...

### DIFF
--- a/worker2/worker.go
+++ b/worker2/worker.go
@@ -1,0 +1,206 @@
+package worker
+
+import (
+	"time"
+
+	schematypes "github.com/taskcluster/go-schematypes"
+	tcclient "github.com/taskcluster/taskcluster-client-go"
+	"github.com/taskcluster/taskcluster-client-go/queue"
+	"github.com/taskcluster/taskcluster-worker/engines"
+	"github.com/taskcluster/taskcluster-worker/lifecyclepolicy"
+	"github.com/taskcluster/taskcluster-worker/plugins"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+	"github.com/taskcluster/taskcluster-worker/runtime/atomics"
+	"github.com/taskcluster/taskcluster-worker/runtime/client"
+	"github.com/taskcluster/taskcluster-worker/runtime/webhookserver"
+)
+
+// A Worker processes tasks
+type Worker struct {
+	environment           runtime.Environment
+	server                webhookserver.Server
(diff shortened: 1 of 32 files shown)


jonasfj:
@petemoore, please just look at the files in the `worker/` folder for 30% review...
The rest is also suggested in other PRs... it's sort of a mess, I tried to file in separate PRs, but also needed it all in one to write this... and didn't really want to land without review on the other PRs :) hehe
Chicken and their eggs..

worker2/worker.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+
+		// If we received zero claims or encountered an error, we wait at-least
+		// pollingInterval before polling again. We start the timer here, so it's
+		// counting while we wait for capacity to be available.
+		var delay <-chan time.Time
+		if claims == nil || len(claims.Tasks) == 0 {
+			delay = time.After(w.pollingInterval)
+		} else {
+			delay = time.After(0)
```
petemoore: I believe task claiming process could run in less time than `pollingInterval`, in which case, we could end up polling more frequently than `pollingInterval`. Also if multiple go routines were claiming work we'd be polling too frequently. Also this is code that would be easy to mess up, so any `0` delays are particularly dangerous as a bug could DDoS our services - would be better for `ClaimWork` to globally throttle, as a centralised safety net, and would also simplify code like this code above.

jonasfj: We only sleep zero if we got tasks from the last `ClaimWork` call...

As long as we're getting new tasks, there is no need to sleep.

jonasfj:
> I see why you created a Barrier primitive, I just wonder if we can do these things without introducing new primitives.

We definitely could do without introducing a new primitive... I could probably have used two `context.Context` objects... But I'm not sure it would be less confusing..

I make the new primitives because I hope I can hide logic in these primitives... So that the concurrency aspects are easier to understand when looking that code that does interesting things.
If barrier is hard to understand then I've definitely failed at that... So the question is did introduction of barrier make this simpler or more complicated?
@petemoore  ^ ?

jonasfj:

PR #218, opened 2017-03-22
Proposing a taskCounter instead

jonasfj:
I'm not sure this is better than what was there before...

please help :)

### DIFF
--- a/worker2/taskcounter.go
+++ b/worker2/taskcounter.go
@@ -1,0 +1,82 @@
+package worker
+
+import (
+	"sync"
+	"time"
+
+	"github.com/taskcluster/taskcluster-worker/runtime/atomics"
+)
+
+// taskCounter keeps count of number of active tasks as well as pending time.
+type taskCounter struct {
+	m         sync.Mutex
+	c         sync.Cond
+	value     int
+	idleTimer atomics.StopWatch
+}
+
+func (c *taskCounter) init() {
+	c.m.Lock()
+	defer c.m.Unlock()
+
+	if c.c.L == nil {
+		c.c.L = &c.m
+		c.idleTimer.Start()
+	}
+}
+
+func (c *taskCounter) IdleTime() time.Duration {
+	c.init()
+	return c.idleTimer.Elapsed()
+}
+
+func (c *taskCounter) WaitForIdle() {
+	c.init()
+	c.m.Lock()
+	defer c.m.Unlock()
+
+	for c.value != 0 {
+		c.c.Wait()
+	}
+}
+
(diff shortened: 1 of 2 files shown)


petemoore:
By the way, why is there a worker**2** package?

worker2/taskcounter.go
```diff
@@ -0,0 +1,82 @@
...
+	"sync"
+	"time"
+
+	"github.com/taskcluster/taskcluster-worker/runtime/atomics"
+)
+
+// taskCounter keeps count of number of active tasks as well as pending time.
+type taskCounter struct {
+	m         sync.Mutex
+	c         sync.Cond
```
jonasfj: @petemoore, you are very right indeed... this is NOT idiomatic.

So I'm trying to do here is create a concurrency primitive that deals exclusive the flow that the worker has...
In this case `taskCounter` tracks the number of tasks currently running, and when this number is zero, the idleTimer is increasing.

That way you can always: 
 * Get the amount of idle time elapsed at a given moment
 * Get the number of tasks currently running
 * Wait for zero tasks to be running
 * Wait for less than N tasks to be running

I could build all these into the worker.... but I would effectively just inline the functions from this file... at different places.. That makes it really hard to see if I'm doing the right thing.
My thinking is that this makes it easier to understand... because I'm splitting out some of the concurrency logic, separating it from the logic that does polling, etc...

Am I all wrong? I fear that if I tried to use channels to track this kind of counter and support the 4 operations listed above it would become hugely complicated.

jonasfj:
@petemoore, you are very right indeed... this is NOT idiomatic.

So I'm trying to do here is create a concurrency primitive that deals exclusive the flow that the worker has...
In this case `taskCounter` tracks the number of tasks currently running, and when this number is zero, the idleTimer is increasing.

That way you can always: 
 * Get the amount of idle time elapsed at a given moment
 * Get the number of tasks currently running
 * Wait for zero tasks to be running
 * Wait for less than N tasks to be running

I could build all these into the worker.... but I would effectively just inline the functions from this file... at different places.. That makes it really hard to see if I'm doing the right thing.
My thinking is that this makes it easier to understand... because I'm splitting out some of the concurrency logic, separating it from the logic that does polling, etc...

Am I all wrong? I fear that if I tried to use channels to track this kind of counter and support the 4 operations listed above it would become hugely complicated.

jonasfj:
I'm going to merge all of these small PRs and make things work...
Then we'll have a really big PR that we rubberstamp review, and we then move along...

It won't be ideal, but at some point we're better off going back and refactoring than being stuck in reviews..

PR #221, opened 2017-03-24
Fix intermittent bugs

jonasfj:
Can I get a quick r+

### DIFF
--- a/commands/qemu-guest-tools/guesttools.go
+++ b/commands/qemu-guest-tools/guesttools.go
@@ -326,10 +326,6 @@ func (g *guestTools) doExecShell(ID string, command []string, tty bool) {
 		err = pipeCommand(shell, handler)
 	}
 
-	// If starting the shell didn't fail, then we wait for the shell to terminate
-	if err == nil {
-		err = shell.Wait()
-	}
 	// If we didn't any error starting or waiting for the shell, then it was a
 	// success.
 	handler.Terminated(err == nil)
@@ -353,6 +349,10 @@ func pipeCommand(cmd *exec.Cmd, handler *interactive.ShellHandler) error {
 		}
 		return nil
 	})
+
+	if err == nil {
+		err = cmd.Wait()
+	}
 
 	return err
 }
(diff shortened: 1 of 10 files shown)


commands/qemu-guest-tools/guesttools.go
```diff
@@ -326,10 +326,6 @@ func (g *guestTools) doExecShell(ID string, command []string, tty bool) {
 		err = pipeCommand(shell, handler)
 	}
 
-	// If starting the shell didn't fail, then we wait for the shell to terminate
-	if err == nil {
-		err = shell.Wait()
-	}
```
jonasfj: I moved this into `pipeCommand|pipePty` to ensure that I'm done streaming before I send the exit message and close the websocket...

plugins/interactive/displayhandler.go
```diff
@@ -66,7 +66,7 @@ func (d *DisplayHandler) sendPings() {
 				return
 			}
 
-			d.monitor.Warn("Failed to send ping, error: ", err)
+			d.monitor.Error("Failed to send ping, error: ", err)
```
jonasfj: Changed these back to error, so they'll cause tests to fail.. and then fixed all the intermittent bug!!! :)
(I hope I fixed them all!)

plugins/interactive/shellhandler.go
```diff
@@ -76,11 +76,11 @@ func (s *ShellHandler) Communicate(setSize SetSizeFunc, abort func() error) {
 	go s.sendPings()
 	go s.waitForSuccess()
 
-	s.streamingDone.Add(2)
+	go s.readMessages()
+
+	s.streamingDone.Add(3)
```
jonasfj: Basically, made it wait til stdin ack messages was sent too...

plugins/interactive/shellserver.go
```diff
@@ -151,15 +159,12 @@ func (s *ShellServer) handleShell(ws *websocket.Conn, shell engines.Shell) {
 
 	// Wait for the shell to terminate
 	success, _ := shell.Wait()
+	wg.Wait() // Wait for pipes to be copied before terminating
```
jonasfj: Wait for all the streams to be copied out before I send the exit message..

jonasfj:
I'll happy with a superficial review... It's okay if nobody but me really cares about the code fixed here...

jonasfj:
Minor tweak that should also address the linter issue... so let see if it passes now.

PR #223, opened 2017-03-31
Worker2

jonasfj:
whenever you have time... note this changes the config schema... 

I know this is a lot of code... pete has done a 30% review on some of it already..
I'm happy if you glance at it, and look a the big picture stuff, ie. how things are structured...

If you like the structure we can fix the bugs later... It should be reasonably tested, and lots of plugin panics should be catched and reported to sentry..

Note: amongst other things plugins can now stop the worker, both gracefully and forcefully. So reboot plugin doesn't have to do hacks anymore.


plugins/pluginmanager.go
```diff
@@ -52,17 +83,18 @@ func PluginManagerConfigSchema() schematypes.Object {
 		MetaData: schematypes.MetaData{
 			Title: "Plugin Configuration",
 			Description: `Mapping from plugin name to plugin configuration.
-                    The 'disabled' key is special and lists plugins that are
-                    disabled. Plugins that are disabled do not require
-                    configuration.`,
+										A plugin is enabled if it has an entry in this mapping, and
+										isn't explicitly listed as 'disabled'. Even plugins that
+										don't require configuration must have an entry, in these
+										cases, empty object will suffice.`,
```
petemoore: I hate to leave formatting comments, but this does look like the indentation has gone wonky.

jonasfj: Yeah, we should use tabs for indentation inside multi-line strings...
I'm working on a tiny`util.Markdown` hack that will remove the indentation and replace `'` with ` so that we can easily write docs...

engines/native/system/process_posix.go
```diff
@@ -77,15 +77,14 @@ func killProcesses(root *process.Process) error {
...
-	if err == nil {
-		// For reach child, we recursively kill all their children too.
-		for _, child := range children {
-			err = killProcesses(child)
-			if err != nil {
-				return err
-			}
+	children, _ := root.Children()
+
+	// For reach child, we recursively kill all their children too.
```
walac: s/reach/each/

engines/native/system/process_posix.go
```diff
@@ -77,15 +77,14 @@ func killProcesses(root *process.Process) error {
...
+	var err2 error
+	for _, child := range children {
+		err = killProcesses(child)
+		if err2 == nil {
```
walac: `err == nil`?

engines/native/system/process_posix.go
```diff
@@ -107,7 +106,11 @@ func killProcesses(root *process.Process) error {
 		}()
 	}
 
-	return nil
+	// if there was an error somewhere we return it
+	if err != nil {
+		return err
+	}
```
walac: Really didn't understand the logic here

runtime/atomics/barrier.go
```diff
@@ -0,0 +1,114 @@
...
+func (c *barrierAsContext) Err() error {
+	if c.b.IsFallen() {
+		return context.Canceled
+	}
+	return nil
+}
+
+func (c *barrierAsContext) Value(interface{}) interface{} {
+	return nil
+}
```
walac: Barrier should be simpler, it is just a synchronization primitive with a `Wait` method. See [boost::barrier](http://www.boost.org/doc/libs/1_33_1/doc/html/barrier.html), for example.

runtime/atomics/counter.go
```diff
@@ -0,0 +1,82 @@
...
+func (c *Counter) Changed() <-chan struct{} {
+	c.m.Lock()
+	defer c.m.Unlock()
+	c.init()
+
+	if c.changed == nil {
+		c.changed = make(chan struct{})
+	}
+	return c.changed
+}
```
walac: What the use case of this counter is? It feels over engineering...

runtime/atomics/stopwatch.go
```diff
@@ -0,0 +1,53 @@
...
+	s.m.Lock()
+	defer s.m.Unlock()
+
+	// If zero value then it's not started yet
+	if s.started == (time.Time{}) {
+		return 0
+	}
+
+	return time.Since(s.started)
+}
```
walac: If thread safety is the only reason to implement our own `Duration`, I think we should not, granularity is too high and chances are the caller already provides a mutex.

runtime/monitoring/doc.go
```diff
@@ -0,0 +1,8 @@
+// Package monitoring provides multiple implementations of runtime.Monitor.
+//
+// In addition to supplying runtime.Monitor implementations this package also
+// provides and a ConfigSchema and a generic New(config) method that can be
```
walac: nit: s/and//

.taskcluster.yml
```diff
@@ -168,3 +168,59 @@ tasks:
...
+          # build tc-worker
+          make build
+          # generate docs
+          taskcluster-worker schema config -o docs/config-schema.json
+          # install taskcluster-lib-docs
+          apt-get update -y
+          apt-get install -y nodejs npm
+          npm install -g taskcluster-lib-docs@3.3.1
+          # upload docs
+          export DOCS_PROJECT=taskcluster-worker DOCS_TIER=integrations DOCS_FOLDER=docs DOCS_README=README.md
```
djmitche: You can use `workers` here now - I fixed it last week.

engines/enginetest/display.go
```diff
@@ -112,6 +116,44 @@ func (c *DisplayTestCase) TestDisplays() {
...
+func (c *DisplayTestCase) TestKillDisplay() {
+	debug("## TestKillDisplay")
+	r := c.newRun()
+	defer r.Dispose()
+	r.NewSandboxBuilder(c.Payload)
+	r.StartSandbox()
+
+	// List displays
+	displays, err := r.sandbox.ListDisplays()
+	nilOrPanic(err, "Failed to ListDisplays")
```
jonasfj: yeah, and we should carry the `*testing.T` variable into all these methods... But we haven't done this yet...

I think we should do it in one big swoop for `enginetest`... pete did it some places, but not in enginetests

engines/native/system/process_posix.go
jonasfj: yeah, okay I have mixed feelings here... Basically, I'm only returning the error from killing a child process, if there is no error with the parent process...

I think we need to interpret the errors and ignore errors about processes already being dead...

engines/native/system/process_posix.go
jonasfj: I prefer to return `err` from parent if there is one, otherwise I return `err2` originating from killing a child process... If there is an `err2` at all..

runtime/atomics/counter.go
jonasfj: yeah, I don't think I ended up using it after all anyways...

runtime/atomics/barrier.go
jonasfj: Yeah, but it's really useful to expose it as a `context.Context`, in case you want to abort something when a barrier falls...

The reason I make barrier a bit more complicated is so that you can wait for it in a few different ways...

The most important thing about the `Barrier` construct is that **a barrier can only fall once**.
And calling `Fall()` more than once just does nothing... which is super nice if you have multiple conditions racing.

runtime/atomics/stopwatch.go
jonasfj: It's not a custom `time.Duration` type... it's just a simple primitive to measure time by calling `Start()` and `Reset()`. The main goal is hide the locking so that we don't have to have a  `time.Time` and a `sync.Mutex` when measuring time.

jonasfj: but yeah, we're only using it once place so I really should just remove it...

.taskcluster.yml
jonasfj: Thanks... bumped to 3.3.2 and fixed...

runtime/util/parallel.go
```diff
@@ -0,0 +1,24 @@
+package util
+
+import "sync"
+
+// Parallel takes a list of functions and calls them all in parallel, returning
+// when all the functions are done.
```
walac: Actually it calls them "concurrently".

jonasfj:
I've adopted your comments, I'll land them in a later PR... Integration tests required  a few other changes.
So I'll put these fix in there.. I can't easily change branch right now... too much in-flux...

PR #229, opened 2017-04-04
reboot after hours, not seconds

djmitche:
I think I had set this to time.Seconds while testing, and forgot to set it back.  All of the docs say "hours".


djmitche:
In puppetagain? Yes, I made a review request for that, too.  I'll need to update that review req with the new tc-worker version, too.

jonasfj:
random note: in the branch `worker2` I'm giving plugins the ability to gracefully shutdown... in a nice way..
I'll likely also give a reboot plugin refresh... and use the new `schematypes.Duration` for config...

djmitche:
@jonasfj that all sounds awesome 🎉

PR #230, opened 2017-04-10
upload docs on push

djmitche:
This duplicates what's on the docs site now, but without using cors-proxy to pull in the README.  It'll be easy to extend to include more documentation, including (perhaps somewhat less easily) auto-generated documentation.


jonasfj:
I've already implemented this in #223... it's listed here:
https://docs.taskcluster.net/reference/integrations/taskcluster-worker 
Under **integrations** because tc-lib-docs doesn't support tier workers yet...

djmitche:
Oh, awesome!

PR #231, opened 2017-04-10
Rework of reboot plugin, and lots of tests

jonasfj:
Filed against `worker2` branch so that it's not so big :)

Half the patch is declarative tests, don't worry too much about those...


plugins/reboot/reboot.go
```diff
@@ -1,161 +1,174 @@
...
+		p.Plugin.rebooted.Do(func() {
+			p.Monitor.Infof("Stopping worker after task as instructed by task.payload.reboot = '%s'", p.RebootAction)
+			p.Plugin.Worker.StopGracefully()
+		})
 	}
+	return nil
+}
 
+func (p *taskPlugin) Exception(runtime.ExceptionReason) error {
+	if p.RebootAction == rebootOnFailure || p.RebootAction == rebootOnException {
```
walac: What about `rebootAlways`?

plugins/reboot/reboot_test.go
```diff
@@ -0,0 +1,193 @@
...
+			"allowTaskReboots": true
+		}`,
+		Payload: `{
+			"delay": 10,
+			"function": "false",
+			"argument": "",
+			"reboot": "on-exception"
+		}`,
+	}.Test()
+}
```
walac: Add test case for when payload is `always` and an exception happens.

plugins/reboot/reboot.go
jonasfj: Great catch!!! :) thanks...

plugins/reboot/reboot_test.go
jonasfj: Fixed, thanks for pointing the bug! :)

jonasfj:
ignore test failure... it's just linter timeouts...

PR #233, opened 2017-04-16
Integration test framework

jonasfj:
@petemoore, this should be a bit nicer than previous PR...
I've split it into reasonable commits...
First commit is a bit of bug fixes and refactoring to fix some issues. It all got intermingled and interdependent.. Tests covering some of these cases will appear later, probably in a different PR.

The `FakeQueue` commit contains the majority of the changes... Feel free to skim through it, but it's not super important the tests are design to run again **both** the _fake queue_ and the _real queue_. So if we have bugs of significance we'll probably find out when testing against the real thing.

The most important thing to review is the _integration testing_ framework. I've tried to make it as declarative as possible. Meaning you define:
 * engine and engine config,
 * plugin configuration
 * tasks
 * assertions for the tasks
   * success and exception reason if not `NoException` (default)
   * Artifacts expected, and conditions on these artifacts
   * whether additional artifacts are allowed
 * If worker is expected to StopNow or StopGracefully on it's own
 * Concurrency, with the trick that concurrency == 0 means tasks will have `task.dependencies` defined such that run in the order given... This is cool if we want to test caches, etc...

Finally, it's possible to write custom assertions... All the assertions are basically functions on the form `func(t *testing.T, ....)`. I don't imagine we'll use that a lot. I'm thinking we'll have functions like `GrepArtifact(substring)` that returns an assertion (a function) that checks that the artifact contains the given substring.

All this is aimed to let us write super high-level integration, where all we have to do is define our input and expected output, and the test framework will run the tests against both fake queue and real queue :)

### DIFF
--- a/worker/workertest/workertest.go
+++ b/worker/workertest/workertest.go
@@ -1,0 +1,315 @@
+package workertest
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http/httptest"
+	"os"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/taskcluster/slugid-go/slugid"
+	tcclient "github.com/taskcluster/taskcluster-client-go"
+	"github.com/taskcluster/taskcluster-client-go/queue"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+	"github.com/taskcluster/taskcluster-worker/runtime/atomics"
+	"github.com/taskcluster/taskcluster-worker/runtime/util"
+	"github.com/taskcluster/taskcluster-worker/worker"
+	"github.com/taskcluster/taskcluster-worker/worker/workertest/fakequeue"
+)
+
+const defaultTestCaseTimeout = 10 * time.Minute
+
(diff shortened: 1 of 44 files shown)


worker/workertest/workertest.go
```diff
@@ -0,0 +1,315 @@
...
+
+// A Task to be included in a worker test case
+type Task struct {
+	Title           string                  // Optional title (for debugging)
+	Payload         string                  // Task payload as JSON
+	Success         bool                    // True, if task should be successfully
+	Exception       runtime.ExceptionReason // Reason, if exception is expected
+	Artifacts       ArtifactAssertions      // Mapping from artifact name to assertion
+	AllowAdditional bool                    // True, if additional artifacts is allowed
+	Status          StatusAssertion         // Optional, custom assertion on status and queue
```
jonasfj: I don't expect we'll use this much, if ever... it's here to allow any custom checks you might want to do.

The signature is `func(t *testing.T, q *queue.Queue, status queue.TaskStatusStructure)` so you can fetch artifacts from a run that isn't the latest, if you wish to do that...

jonasfj:
Current test failure is just linter timing out because worker is overloaded...

PR #237, opened 2017-04-20
Artifact rework

jonasfj:
Integration tests covering larger parts of this will be added later...

### DIFF
--- a/plugins/artifacts/artifacts.go
+++ b/plugins/artifacts/artifacts.go
@@ -1,143 +1,315 @@
-// Package artifacts is responsible for uploading artifacts after builds
 package artifacts
 
 import (
+	"context"
+	"fmt"
 	"mime"
 	"path"
 	"path/filepath"
-	"time"
+	"strings"
+	"sync"
 
 	schematypes "github.com/taskcluster/go-schematypes"
 	"github.com/taskcluster/taskcluster-worker/engines"
 	"github.com/taskcluster/taskcluster-worker/plugins"
 	"github.com/taskcluster/taskcluster-worker/runtime"
+	"github.com/taskcluster/taskcluster-worker/runtime/atomics"
 	"github.com/taskcluster/taskcluster-worker/runtime/ioext"
+	"github.com/taskcluster/taskcluster-worker/runtime/util"
 )
+
+const unknownMimetype = "application/octet-stream"
+
+// Maximum concurrent uploads, note that we might do concurrent uploads for
+// folder artifacts too, causing a total of:
+//   maxUploadConcurrency * maxUploadConcurrency
+const maxUploadConcurrency = 5
 
 type pluginProvider struct {
 	plugins.PluginProviderBase
-}
-
-func (pluginProvider) NewPlugin(plugins.PluginOptions) (plugins.Plugin, error) {
-	return plugin{}, nil
 }
 
 type plugin struct {
 	plugins.PluginBase
 }
 
-func (plugin) PayloadSchema() schematypes.Object {
-	return payloadSchema
-}
-
-func (plugin) NewTaskPlugin(options plugins.TaskPluginOptions) (plugins.TaskPlugin, error) {
-	var p payloadType
-	schematypes.MustValidateAndMap(payloadSchema, options.Payload, &p)
-
-	return &taskPlugin{
-		TaskPluginBase: plugins.TaskPluginBase{},
-		artifacts:      p.Artifacts,
-		context:        options.TaskContext,
-	}, nil
-}
-
 type taskPlugin struct {
 	plugins.TaskPluginBase
-	context   *runtime.TaskContext
-	artifacts []artifact
-}
-
-func (tp *taskPlugin) Stopped(result engines.ResultSet) (bool, error) {
-	nonFatalErrs := []runtime.MalformedPayloadError{}
-
-	for _, artifact := range tp.artifacts {
-		// If expires is set to this time it's the default value
-		if artifact.Expires.IsZero() {
-			artifact.Expires = tp.context.TaskInfo.Expires
-		}
-		switch artifact.Type {
-		case "directory":
-			err := result.ExtractFolder(artifact.Path, tp.createUploadHandler(artifact.Name, artifact.Expires))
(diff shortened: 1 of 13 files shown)


jonasfj:
Fix for the tests was added to another PR... it's just timing robustness... unrelated to changes made here.

plugins/artifacts/artifacts.go
```diff
@@ -1,143 +1,315 @@
...
+		return
 	}
 
-	if len(nonFatalErrs) > 0 {
-		// Only report an exception if the command executed successfully.
-		// The logic behind this it is that no artifact upload failure is
-		// expected for a succeeded run, but failed tasks might be missing
-		// some artifacts.
+	// If resource isn't found, task should fail and we print a message to task log
+	if err == engines.ErrResourceNotFound {
```
walac: Wouldn't this be `err == engines.ErrResourceNotFound && result.Success()`?
If the task has failed, it is expected some artifacts are missing, no reason to fail the plugin.

plugins/artifacts/artifacts.go
```diff
@@ -1,143 +1,315 @@
...
+	if err == runtime.ErrNonFatalInternalError {
+		tp.nonFatalErr.Set(true) // Propagate the non-fatal error
+		tp.monitor.Warn("Received ErrNonFatalInternalError from ResultSet.ExtractFile()")
+		return
+	}
+
+	// If fatal internal error, we want to propagate (what else can we do)
+	if err == runtime.ErrFatalInternalError {
+		tp.fatalErr.Set(true) // Propagate the fatal error
+		tp.monitor.Error("Received ErrFatalInternalError from ResultSet.ExtractFile()")
```
walac: If `err` contains useful information, we should log it here

plugins/artifacts/artifacts.go
```diff
@@ -1,143 +1,315 @@
...
+		))
+		tp.mErrors.Unlock()
+		return
+	}
 
-		return false, nil
+	// If non-fatal internal error, we want to propagate (what else can we do)
+	if err == runtime.ErrNonFatalInternalError {
+		tp.nonFatalErr.Set(true) // Propagate the non-fatal error
+		tp.monitor.Warn("Received ErrNonFatalInternalError from ResultSet.ExtractFile()")
```
walac: If `err` contains useful information, we should log it here

plugins/artifacts/artifacts.go
```diff
@@ -1,143 +1,315 @@
...
+		tp.mErrors.Lock()
+		tp.errors = append(tp.errors, e)
+		tp.mErrors.Unlock()
+		return
 	}
 
-	if reason != "" {
-		tp.context.Log("Artifact upload error handled. Continuing...", name, err.Error())
+	// If resource isn't found, task should fail and we print a message to task log
+	if err == engines.ErrResourceNotFound {
```
walac: See my comment in `processFile`

plugins/artifacts/artifacts.go
```diff
@@ -1,143 +1,315 @@
...
+		tp.mErrors.Unlock()
+		return
+	}
 
-func init() {
-	plugins.Register("artifacts", pluginProvider{})
+	// If non-fatal internal error, we want to propagate (what else can we do)
+	if err == runtime.ErrNonFatalInternalError {
+		tp.nonFatalErr.Set(true) // Propagate the non-fatal error
+		tp.monitor.Warn("Received ErrNonFatalInternalError from ResultSet.ExtractFolder()")
```
walac: Same comment about logging `err`

plugins/artifacts/artifacts.go
```diff
@@ -1,143 +1,315 @@
...
+		return
+	}
+
+	// If fatal internal error, we want to propagate (what else can we do)
+	if err == runtime.ErrFatalInternalError {
+		tp.fatalErr.Set(true) // Propagate the fatal error
+		tp.monitor.Error("Received ErrFatalInternalError from ResultSet.ExtractFolder()")
```
walac: Same comment about logging `err`

plugins/artifacts/doc.go
```diff
@@ -0,0 +1,7 @@
+// Package artifacts is provides a taskcluster-worker plugin that uploads
```
walac: s/is //

worker/workertest/artifactassertions.go
```diff
@@ -20,6 +21,21 @@ func GrepArtifact(substring string) func(t *testing.T, a Artifact) {
 	}
 }
 
+// LogArtifact creates an assetions that logs the artifact, to test log.
```
walac: s/assetions/assertion/

plugins/artifacts/artifacts.go
jonasfj: It doesn't... It just is a specific value...

plugins/artifacts/artifacts.go
jonasfj: We should set failed regardless... of `result.Success()` (just for good measure)...
I would argue the same for the error artifact...

jonasfj: regardless, we definitely need to handle the error... so it's not reported as unhandled error.

plugins/artifacts/artifacts.go
jonasfj: it doesn't... `err == runtime.ErrNonFatalInternalError`, ie. it's a specific value.

plugins/artifacts/artifacts.go
jonasfj: Okay, I'm moving the `CreateErrorArtifact` under the `if result.Success() {`...
but I do want to handle the error...

plugins/artifacts/artifacts.go
jonasfj: contains no info

plugins/artifacts/artifacts.go
jonasfj: contains no info

plugins/artifacts/artifacts.go
jonasfj: okay, changed my mind again...

PR #238, opened 2017-04-20
Improved schema validation error log

jonasfj:
Example task log:

> [taskcluster:error]  Required property 'delay' is missing at task.payload.delay
[taskcluster:error]  Required property 'function' is missing at task.payload.function
[taskcluster:error]  Required property 'argument' is missing at task.payload.argument

Hopefully this will make it easier to figure out what is wrong with the `task.payload` property.
I haven't tried it with complex paths, but it should also work things like `task.payload.artifacts[2].name`


jonasfj:
Sorry, for the mess, but tests were failing so I added a PR improving robustness of the `watchdog` tests...
Changes are unrelated, but this is a tiny PR so seems like a cheap fix in exchange for having tests passing :)

PR #239, opened 2017-04-21
Env overwrite + built-in env vars

jonasfj:
* Reordered things a bit in `env` plugin
 * Allow for env vars to overwrite, extra env vars, and allow to overwrite built-in
 * Added `TASK_ID` and `RUN_ID` as built-in env vars...
 * More tests :)


plugins/env/env.go
```diff
@@ -86,9 +124,7 @@ func (p *taskPlugin) BuildSandbox(sandboxBuilder engines.SandboxBuilder) error {
 			return runtime.NewMalformedPayloadError("Environment variable ", k, " has already been set.")
 		case engines.ErrFeatureNotSupported:
 			return runtime.NewMalformedPayloadError(
-				"Cannot set environment variable ",
-				k,
-				". Engine does not support this operation")
+				"Custom environment variables is not supported in the current configuration of the engine")
```
walac: s/is/are/

PR #247, opened 2017-04-21
Experiment to add armv7l builds/tests to CI

petemoore:
Just thought I'd try this out for fun. Maybe useful if we one day run fennec builds/tests on arm (e.g. in packet.net).

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -8,7 +8,6 @@ metadata:
   source: "{{ event.head.repo.url }}"  # the repo where the pr came from will be inserted here
 tasks:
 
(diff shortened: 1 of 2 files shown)


jonasfj (approved):
This is super awesome!!

Note: I recommend disabling the test cases that fail on arm... we can just do that with a build constraint... right?

Just to have tests passing, we can leave a TODO saying that the build constraint should be removed and tests fixed on arm...

jonasfj:
IMO, it's okay to disable the qemu guest tools tests on arm... I think that's only one breaking...

PR #249, opened 2017-04-24
Early logging

djmitche:
This makes monitors available early in the startup process.  It also copies monitor output to syslog, using some code I found in `commands/daemon`.

I expect the objection to this is that the `schema` command shouldn't need a monitor.  But it exists as a usre-visible command because it's a context-sensitive operation, and part of that context is errors and transient failures -- so I think logging makes sense even there.


jonasfj:
Hmm, not sure if it wouldn't be more appropriate to add a `monitoring.NewSysloggingMonitor()`...
Or maybe let `NewLoggingMonitor` take a `logrus.Logger` as first argument.

Note: given that `reportError` and statsum stuff won't work before config anyways, maybe it's better to just use a logrus thing before then... or even better a custom interface:
```go
package runtime

type Logger interface{
  Debugf(format string, a ...interface{})
  Infof(format string, a ...interface{})
  Warnf(format string, a ...interface{})
  Errorf(format string, a ...interface{})
}
```

Which is implicitly implemented by `Monitor` and `logrus` then we can pass in whatever is available...

I'll think more about this tomorrow... but yeah, we should have logging during config... thanks

djmitche:
I was a little confused why there was `NewMonitor` and `NewLoggingMonitor` where the former is a copy of the latter with statsum/sentry added, but I figured there was good reason for it.

At any rate, `loggingMonitor` does log from `reportError`, which is what I want in this case.

djmitche:
Also, the CI appears to be failing because gometalinter is timing out.

runtime/monitoring/syslog.go
```diff
@@ -1,6 +1,6 @@
 // +build !windows
 
-package daemon
+package monitoring
```
jonasfj: you forgot this in syslog_null...  I've done this in https://github.com/taskcluster/taskcluster-worker/pull/259

PR #251, opened 2017-04-24
Reboot and maxruntime tests for native engine

jonasfj:
More tests will be added later... please review these...

### DIFF
--- a/engines/native/resultset.go
+++ b/engines/native/resultset.go
@@ -101,6 +101,11 @@ func (r *resultSet) ExtractFolder(path string, handler engines.FileHandler) erro
 		if _, ok := err.(*os.PathError); ok && first {
 			return engines.ErrResourceNotFound
 		}
+		// If first path is what we're walking and it's not a directory, then we
+		// didn't find folder at the given path.
+		if first && p == abspath && !info.IsDir() {
+			return engines.ErrResourceNotFound
+		}
 		first = false
 
 		// Ignore folder we can't walk (probably a permission issues)
--- a/engines/native/sandbox.go
+++ b/engines/native/sandbox.go
@@ -32,10 +32,11 @@ type sandbox struct {
 	shells        []*shell
 }
 
-func newSandbox(b *sandboxBuilder) (s *sandbox, err error) {
+func newSandbox(b *sandboxBuilder) (engines.Sandbox, error) {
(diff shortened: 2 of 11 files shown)


jonasfj:
@petemoore, this should cover most of the test cases from `new-tests`...
It's mostly additions and easy to read, please review. This have been open for a few days now, in one of two forms... I just rebased and added more test cases.. But I can't keep the PR pretty going forward as I'll eventually have more refactoring and stuff to fix..

So please review while it's quick to do...

engines/native/sandbox.go
```diff
@@ -32,10 +32,11 @@ type sandbox struct {
 	shells        []*shell
 }
 
-func newSandbox(b *sandboxBuilder) (s *sandbox, err error) {
+func newSandbox(b *sandboxBuilder) (engines.Sandbox, error) {
```
jonasfj: This change was needed because we do `return newSandbox(...)` in `SandboxBuilder` and when `newSandbox` returns `*sandbox(nil)` it gets converted to `engines.Sandbox != nil`, I know that messed up.. But `nil` is a valid value for `*sandbox` and `*sandbox` implements `engines.Sandbox`... So if we're not careful to explicitly return nil, we return a nil implementation... which cases issues if you call methods on it..

engines/native/resultset.go
```diff
@@ -101,6 +101,11 @@ func (r *resultSet) ExtractFolder(path string, handler engines.FileHandler) erro
 		if _, ok := err.(*os.PathError); ok && first {
 			return engines.ErrResourceNotFound
 		}
+		// If first path is what we're walking and it's not a directory, then we
+		// didn't find folder at the given path.
```
walac: I am not sure if I followed the idea here...

jonasfj: Basically, if we did  `filepath.Walk("/home/jonasfj/myfile.txt", func(abspath string, ...` then if `"/home/jonasfj/myfile.txt"` is a file an not a directory, it will call our callback with `abspath == "/home/jonasfj/myfile.txt"`...

So if we try to export a directory artifact, but the path is actually the path to a file instead of a folder, we return that single file.... If `ExtractFolder` is called with the path to a file, it should return `ErrResourceNotFound`... It's a weird go-walk corner case...

PR #253, opened 2017-04-25
Fix #242 - use schematypes.Duration for maxLifeCycle in webhookserver

### LINKED ISSUE #242: webHookServer.maxLifeCycle is not in the same format as plugins.reboot.maxLifeCycle
The former is an integer (in seconds, I guess?) while the latter is a string.  They should be the same format.

jonasfj:


runtime/webhookserver/config.go
```diff
@@ -49,14 +49,12 @@ var statelessDNSConfigSchema = schematypes.Object{
 		"tlsKey":             schematypes.String{},
 		"statelessDNSSecret": schematypes.String{},
 		"statelessDNSDomain": schematypes.String{},
-		"maxLifeCycle": schematypes.Integer{
+		"maxLifeCycle": schematypes.Duration{
 			Title: "Maximum lifetime of the worker in seconds",
```
djmitche: remove "in seconds"

PR #254, opened 2017-04-25
Fix 243 - Make webhookserver config optional with a provider: none, o…

jonasfj:
…ption

Hmm, I'm not sure this is the most elegant solution... Maybe it would be better to allow `AttachWebHook` to return an error, like `ErrFeatureNotSupported` and `ErrNonFatalError`... So we can make a payload error if people try to create interactive when not supported...

Granted I do love the simplicity of not returning errors...

### DIFF
--- a/plugins/interactive/interactive.go
+++ b/plugins/interactive/interactive.go
@@ -180,41 +180,21 @@ func (p *taskPlugin) Started(sandbox engines.Sandbox) error {
 }
 
 func (p *taskPlugin) Stopped(_ engines.ResultSet) (bool, error) {
-	wg := sync.WaitGroup{}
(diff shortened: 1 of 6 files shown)


jonasfj:
writing to task logs makes sense if this was intermittent..

Hmm, you right, this is not the best solution... Maybe I should remove `AttachWebHook` from `TaskContext`, and require that plugins use the `WebHookServer` from `runtime.Environment` directly...
I suspect what I'm doing now is overkill...

Then it's easier to signal it's missing by just leaving `WebHookServer` as `nil`...

jonasfj:
Fixed this with PR https://github.com/taskcluster/taskcluster-worker/pull/255

PR #256, opened 2017-04-26
Use cgo everywhere

djmitche:
Without cgo, we can't use os/user (on any platform, not just Darwin).
And we need os/user to create users, etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,7 @@ GO_MAJ := $(shell echo "$(GO_VERSION)" | cut -f1 -d'.')
 GO_MIN := $(shell echo "$(GO_VERSION)" | cut -f2 -d'.')
 
 uname := $(shell uname)
-is_darwin := $(filter Darwin,$(uname))
-CGO_ENABLE := $(if $(is_darwin),1,0)
+CGO_ENABLED := 1
 
 .PHONY: all prechecks build rebuild check test dev-test tc-worker-env tc-worker
 
@@ -26,7 +25,7 @@ prechecks:
 
 build:
 	go fmt $$(go list ./... | grep -v /vendor/)
-	CGO_ENABLED=$(CGO_ENABLE) go install
+	CGO_ENABLED=$(CGO_ENABLED) go install
 
 rebuild: prechecks build test lint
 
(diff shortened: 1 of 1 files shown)


petemoore:
I remember `go test --race` requiring either CGO enabled or disabled - I don't remember which! I guess we'll find out. Maybe we are lucky. :-)

jonasfj:
So I got race detector working with CGO on linux...

Makefile
```diff
@@ -46,7 +45,7 @@ tc-worker-env:
 	docker build -t taskcluster/tc-worker-env -f tc-worker-env.Dockerfile .
 
 tc-worker:
-	CGO_ENABLED=$(CGO_ENABLE) GOARCH=amd64 go build
+	CGO_ENABLED=$(CGO_ENABLED) GOARCH=amd64 go build
 	docker build -t taskcluster/tc-worker -f tc-worker.Dockerfile .
```
jonasfj: hmm, this is a little unpleasant... if the system has a newer glibc version than what is inside the docker container, then we get issues...

Maybe we fix that issue, if we ever hit it... I'm guessing the solution is to use a different docker image to build the it...

djmitche:
Or an older system for the build.  I think there's some way to version glibc requirements (we do that for Firefox).

jonasfj:
Setting glibc requirements is possible... The only solutions to that I have seen is something like `__asm__(".symver realpath,realpath@GLIBC_2.2.5");`... Doing that from golang feels like a painful thing..
Or setting `-D_GLIBC_2_0_SOURCE`... But I can't find any good documentation for this...

http://ftp.ntu.edu.tw/software/libs/glibc/hjl/compat/
Doesn't feel like the official docs... I can't find any references in glibc manual so I prefer to avoid such things...

djmitche:
@glandium may be able to offer advice.  Just adding `-D<something>` in the Makefile doesn't seem awful.

petemoore:
@djmitche @jonasfj It is worth having a read of https://dave.cheney.net/tag/cgo - maybe we can achieve this a different way?

petemoore:
https://github.com/mitchellh/go-homedir might be an option.

jonasfj:
If we're just doing it to get homedir then I support working around it....

I'm already avoiding CGO for zstd by calling a subprocess instead and
piping through that...
This also has the benefit that if zstd crashes it would propagate to
tc-worker...

Den 27. apr. 2017 10.08 AM skrev "Pete Moore" <notifications@github.com>:

glandium:
> @glandium may be able to offer advice. Just adding -D<something> in the Makefile doesn't seem awful.

It's not clear to me what this is all about, so I can give much advice.

PR #257, opened 2017-04-26
Log task start/done at INFO

djmitche:
This will be useful for time-boxing errors in the logs.


worker/worker.go
```diff
@@ -303,7 +303,7 @@ func (w *Worker) processClaim(claim taskClaim) {
 		"taskId": claim.Status.TaskID,
 		"runId":  strconv.Itoa(claim.RunID),
 	})
-	monitor.Debug("starting to process task")
+	monitor.Info("starting to process task")
 	defer monitor.Debug("done processing task")
```
jonasfj: You migth want this at info level too?

djmitche: Hah, yes, updated.

PR #258, opened 2017-04-26
Enabled CGO in more places, and build in docker container for tc-work…

jonasfj:
…er image


jonasfj:
if tests pass then let's use CGO everywhere :)

PR #266, opened 2017-05-04
Fixed documentation comment

jonasfj:


walac:
Always wanted to know the difference between "doesn't" and "does not" in formal english

jonasfj:
There is no difference... if formally written English `doesn't` is sometimes avoided... I did it here because the function `util.Markdown` changes `'` to ` to support markdown...

PR #267, opened 2017-05-04
CI for qemu-engine running under tc-worker w. qemu-engine on packet...

jonasfj:
On packet I deploy using coreos/docker, and I reuse docker inside the ubuntu VM, so we have:
 `coreos > docker > qemu > ubuntu > docker > qemu > tinycorelinux`

Haha :)

For anyone thinking this is crazy it's:
 1. coreos stable on packet.net
 2. docker running `taskcluster/tc-worker`
 3. taskcluster-worker running a QEMU VM per-task
 4. The per-task VM we run tests in is Ubuntu
 5. Inside the ubuntu VM we run `make tc-worker-env-tests`, which pulls `taskcluster/tc-worker-env`
 6. Inside the docker container we run taskcluster-worker tests w. `go test`
 7. Tests runs QEMU w. tincorelinux as the test image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ GO_MIN := $(shell echo "$(GO_VERSION)" | cut -f2 -d'.')
 uname := $(shell uname)
 CGO_ENABLED := 1
 
-.PHONY: all prechecks build rebuild check test dev-test tc-worker-env tc-worker
+.PHONY: all prechecks build rebuild check test dev-test tc-worker-env tc-worker tc-worker-env-tests
 
 all: rebuild
 
@@ -47,6 +47,15 @@ tc-worker-env:
 tc-worker:
 	CGO_ENABLED=$(CGO_ENABLED) GOARCH=amd64 ./docker-tests.sh go build
 	docker build -t taskcluster/tc-worker -f tc-worker.Dockerfile .
+
+tc-worker-env-tests:
+	docker run \
+		--tty --interactive --rm --privileged \
+		-e DEBUG -e GOARCH -e CGO_ENABLED=$(CGO_ENABLED) \
+		-v $$(pwd):/go/src/github.com/taskcluster/taskcluster-worker/ \
+		taskcluster/tc-worker-env \
+		go test -race -tags 'qemu network system native' -p 1 -v \
+		$$(find  -name '*_test.go' | xargs dirname | sort | uniq)
 
 lint:
 	go get github.com/alecthomas/gometalinter
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ tc-worker:
 	docker build -t taskcluster/tc-worker -f tc-worker.Dockerfile .
 
 tc-worker-env-tests:
-	@if [ `whoami` != 'root' ]; then \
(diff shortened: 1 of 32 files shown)


engines/enginetest/artifacts.go
```diff
@@ -158,12 +158,9 @@ func (c *ArtifactTestCase) TestExtractFolderHandlerInterrupt() {
...
 func (c *ArtifactTestCase) Test() {
-	wg := sync.WaitGroup{}
-	wg.Add(5)
-	go func() { c.TestExtractTextFile(); wg.Done() }()
-	go func() { c.TestExtractFileNotFound(); wg.Done() }()
-	go func() { c.TestExtractFolderNotFound(); wg.Done() }()
-	go func() { c.TestExtractNestedFolderPath(); wg.Done() }()
-	go func() { c.TestExtractFolderHandlerInterrupt(); wg.Done() }()
-	wg.Wait()
+	c.TestExtractTextFile()
```
jonasfj: I made all tests run in sequentially, as it timed out when running in parallel...

I suspect I'll refactor this later to use the new `t.Run` stuff...

Makefile
```diff
@@ -48,6 +48,31 @@ tc-worker:
 	CGO_ENABLED=$(CGO_ENABLED) GOARCH=amd64 ./docker-tests.sh go build
 	docker build -t taskcluster/tc-worker -f tc-worker.Dockerfile .
 
+tc-worker-env-tests:
+	@if [ `whoami` != 'root' ]; then \
```
walac: Isn't there any docker/taskcluster specific environment variable to check?

Makefile
```diff
@@ -48,6 +48,31 @@ tc-worker:
...
+		echo "This command should only be used in the CI environment"; \
+		exit 1 ; \
+	fi
+	@echo '### Pulling tc-worker-env'
+	@docker pull taskcluster/tc-worker-env > /dev/null
+	@echo '### Running govendor sync'
+	@docker run \
+		--tty --rm --privileged \
+		-e DEBUG -e GOARCH -e CGO_ENABLED=$(CGO_ENABLED) \
```
walac: Are you sure you don't want `GOARCH=$(GOARCH)`?

Makefile
```diff
@@ -48,6 +48,31 @@ tc-worker:
...
+		-v $$(pwd):/go/src/github.com/taskcluster/taskcluster-worker/ \
+		taskcluster/tc-worker-env \
+		govendor sync
+	@echo '### Downloading test images'
+	@./engines/qemu/test-image/download.sh
+	@echo '### Running tests'
+	@docker run \
+		--tty --rm --privileged \
+		-e DEBUG -e GOARCH -e CGO_ENABLED=$(CGO_ENABLED) \
```
walac: Same GOARCH comment goes here

Makefile
jonasfj: Don't think it really matters, but let's do it for consistency :)

jonasfj:
QEMU tests will fail because if a bug I made and then rebase this on... don't worry it'll get fix in a later PR...

PR #268, opened 2017-05-04
Fix 262 - Prefix logs with a simple header

jonasfj:
It's possible to inject additional things through configuration....

We can work on adding more built-ins, but since not all workers have a public IPs, etc. I don't know how much makes sense...

Maybe it makes sense to add things like worker: taskcluster-worker, and engine: ...
Perhaps even git revision of taskcluster-worker, but so far we don't have this easily available in the source...

Note: config makes it easy to add dynamic things like publicIP using the transforms... we already use them for things like public IP on packet.net.

As for log footers I suggest these are added by the engines.. and that they aren't exactly footers... but more like a message showing the exit code...

### DIFF
--- a/plugins/logprefix/logprefix.go
+++ b/plugins/logprefix/logprefix.go
@@ -1,0 +1,86 @@
+package logprefix
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+
+	schematypes "github.com/taskcluster/go-schematypes"
+	"github.com/taskcluster/taskcluster-worker/plugins"
+)
+
+type provider struct {
+	plugins.PluginProviderBase
+}
+
+type plugin struct {
+	plugins.PluginBase
+	keys map[string]string
+}
+
+func init() {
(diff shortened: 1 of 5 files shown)


plugins/logprefix/logprefix.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	// Sort list of allKeys (to ensure consistency)
+	sort.Strings(allKeys)
+
+	// Print keys to task log
+	for _, k := range allKeys {
+		v, ok := p.keys[k]
+		if !ok {
+			v = keys[k]
+		}
```
walac: I don't the logic here, why not put everything inside `keys` and print it?

jonasfj: Hmm... that's a really good question... I guess this might be a tiny bit faster... but that's not important...

Note: I need to create `allKeys` to sort the ordering... so it's consistent.. otherwise will randomize... so this doesn't make much difference to me...

PR #275, opened 2017-05-10
Fixed #271 - Set default value of watchdog

### LINKED ISSUE #271: Empty watchdog config fails
```
  plugins:
    watchdog: {}
```
gives
```
<13>May  9 08:41:36 taskcluster-worker: time="2017-05-09T08:41:37-07:00" level=error error="watchdog: worker timeout exceeded" incidentId=a4c10829-ce6f-45ce-8b37-afc8352f2e14 plugin=watchdog prefix=plugin.watchdog 
```

jonasfj:


plugins/watchdog/doc.go
```diff
@@ -10,3 +10,7 @@
 // times a good solution is to report to sentry and kill the worker. Ideally,
 // such bugs can be found and fixed over time.
 package watchdog
+
+import "github.com/taskcluster/taskcluster-worker/runtime/util"
+
+var debug = util.Debug("watchdog")
```
djmitche: Why is this in `doc.go`?

jonasfj: I figured it was good place to put... the `doc.go` contains a comment that relates to the `package watchdog` statement... and this is also a once-per-package kind of thing...

PR #279, opened 2017-05-18
Kms vpn for qemu

jonasfj:
Take a look at my iptables things and tell me if I'm all wrong...
It's not something I can easily write tests for, so I'll probably leave it for later...


jonasfj:
I just forgot to push an updated docker images.. I did a retrigger on tools... let's see if that passes.

engines/qemu/network/iptables.go
```diff
@@ -72,35 +74,66 @@ func ipTableRules(tapDevice string, ipPrefix string, delete bool) [][]string {
...
-		// Allow tctap1 -> tctap1 within allowed subnet
-		{"-o", "tctap1", "-s", subnet, "-j", "ACCEPT"},
-		// Reject all other input for forwarding from tap-device
-		{"-j", "REJECT", "--reject-with", "icmp-net-prohibited"},
-	})
+	forwardInputRules := prefixCommands([]string{"iptables", "-w", xtableLockWait, ruleAction, "fwd_input_" + tapDevice}, append(
+		// Allow tap device -> VPN
+		forwardVPNInputRules,
+		[][]string{
+			// Reject out-going from this tap device to private subnets
```
imbstack: Why do we need to do this?

jonasfj: The thinking is that if we're running tc-worker in packet.net or in data center, then machines might live on some private network... And we don't want to them contact resources inside that network unless explicitly allowed.

For example in scaleway we might have to use the `taskcluster-host-secrets` to grant temporary credentials to anyone on a specific private network.

Also we use the `192.168.0.0/16` subnet for the VMs themselves...

imbstack: Cool. Makes sense. Let's add that as a comment here then. I think this section could use really heavy commenting as to the motivation behind each rule.

jonasfj: The rules are an attempt to enforce that VM use the right source address and allow access to:
 * metadata service at `169.254.169.254` on port 80
 * dns server (dnsmasq)
 * DHCP server (dnsmasq)
 * routes connected through VPN
 * public IPv4 subnet

jonasfj: Good point...

PR #281, opened 2017-05-26
Customizable vm

jonasfj:


commands/qemu-build/buildimage.go
```diff
@@ -133,3 +136,43 @@ func buildImage(
...
 	return nil
 }
+
+// load vm.Machine from file without migration
+func newMachineFromFile(machineFile string) (*vm.Machine, error) {
+	// Read machine.json
+	machineData, err := ioext.BoundedReadFile(machineFile, 1024*1024)
+	if err == ioext.ErrFileTooBig {
+		return nil, runtime.NewMalformedPayloadError(
+			"The file 'machine.json' larger than 1MiB. JSON files must be small.")
```
imbstack: Just curious. Why?

engines/qemu/image/mutableimage.go
```diff
@@ -48,7 +48,7 @@ func NewMutableImage(folder string, size int, machine *vm.Machine) (*MutableImag
 
 	return &MutableImage{
 		folder:  folder,
-		machine: machine.Clone(),
+		machine: machine,
```
imbstack: Do we not clone anymore because machine is now ok to mutate?

commands/qemu-build/buildimage.go
jonasfj: If you construct an image file that contains `machine.json` that is 10GiB large I might happily download and extract the image... but I won't try to load it into memory... As it would crash the worker..

Images are untrusted.

engines/qemu/image/mutableimage.go
jonasfj: No, we don't clone because I removed the method and made it immutable...

PR #284, opened 2017-06-12
Remove VGA options from qemu -- I could not figure out sane defaults,…

jonasfj:
… we can add them back any day

I probably shouldn't add features I can figure out how to use... :)


jonasfj:
thanks for the quick fix review... :)

The test failure seems to be a timeout on travis... our mac test pass so I'm happy... also no reason those test can fail because of this change :)

PR #285, opened 2017-06-14
Private images

jonasfj:
* Tweaks the `fetcher` package to support more sources and be more useful
 * Improves some error cases
 * Moves qemu engine to use fetcher to download images...

Note: This can be reviewed commit by commit, or all at once.. I tried to make each commit sane on it's own...

### DIFF
--- a/runtime/fetcher/errors.go
+++ b/runtime/fetcher/errors.go
@@ -6,17 +6,19 @@ import "fmt"
 // Error message shall be some human-readable explanation of why fetching failed
 // and is expected to fail consistently.
 type BrokenReferenceError struct {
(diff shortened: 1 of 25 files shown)


jonasfj:
Test failures are unrelated, I rebased and fixed the issue I rebased on in another PR...

runtime/fetcher/errors.go
```diff
@@ -0,0 +1,30 @@
+package fetcher
+
+import "fmt"
+
+// BrokenReferenceError is used to communicated references that are broken.
```
walac: nit: s/communicated/communicate/

Issue #280, opened 2017-05-24
prefix logs with task count


jonasfj:
so we can see how many tasks the worker has done since it was booted...

jonasfj:
Fixed in PR #286

PR #286, opened 2017-06-14
Fix issue #280 - Add count of TasksSinceBoot to logprefix

jonasfj:
I'm sure you and gps will find this useful...

Ps. if you have a prettier or more obvious name that `TasksSinceBoot` I'm very interested, I'm not entirely happy with the name...

### DIFF
--- a/plugins/logprefix/logprefix.go
+++ b/plugins/logprefix/logprefix.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"sort"
 	"strconv"
+	"sync/atomic"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


plugins/logprefix/logprefix.go
```diff
@@ -54,10 +56,15 @@ func (provider) NewPlugin(options plugins.PluginOptions) (plugins.Plugin, error)
...
+	taskCount := atomic.AddInt64(&p.taskCount, 1) - 1
+
 	keys := map[string]string{
-		"TaskId":       options.TaskContext.TaskID,
-		"RunId":        strconv.Itoa(options.TaskContext.RunID),
-		"HostBootTime": p.bootTime,
+		"TaskId":         options.TaskContext.TaskID,
+		"RunId":          strconv.Itoa(options.TaskContext.RunID),
+		"HostBootTime":   p.bootTime,
+		"TasksSinceBoot": strconv.FormatInt(taskCount, 10),
```
djmitche: This isn't necessarily since boot - let's call it `TasksSinceStartup`?

jonasfj: I guess we could change to do that...

PR #294, opened 2017-07-09
Refactor worker to use webhooktunnel

ckousik:
The worker can use `webhooktunnelToken` auth endpoint to authenticate and create a tunnel using webhooktunnel proxy.

### DIFF
--- a/runtime/webhookserver/config.go
+++ b/runtime/webhookserver/config.go
@@ -6,7 +6,6 @@ import (
 
 	schematypes "github.com/taskcluster/go-schematypes"
 	"github.com/taskcluster/taskcluster-client-go"
-	"github.com/taskcluster/taskcluster-client-go/auth"
 	"github.com/taskcluster/taskcluster-worker/runtime/util"
 )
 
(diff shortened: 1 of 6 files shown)


worker/worker.go
```diff
@@ -81,7 +81,7 @@ func New(config interface{}) (w *Worker, err error) {
 
 	// Create webhookserver
 	if c.WebHookServer != nil {
-		w.webhookserver, err = webhookserver.NewServer(c.WebHookServer)
+		w.webhookserver, err = webhookserver.NewServer(c.WebHookServer, &c.Credentials)
```
jonasfj: Let's pass in `monitor.WithPrefix("webhookserver")`

runtime/webhookserver/config.go
```diff
@@ -107,6 +119,15 @@ func NewServer(config interface{}) (Server, error) {
 	if schematypes.MustMap(localtunnelConfigSchema, config, &c) == nil {
 		return NewLocalTunnel(c.BaseURL)
 	}
+	if schematypes.MustMap(webhooktunnelConfigSchema, config, &c) == nil {
+		authCl := auth.New(credentials)
+		whresp, err := authCl.WebhooktunnelToken()
+		if err != nil {
+			return nil, err
+		}
+		s, err := NewWebhookTunnel(whresp)
```
ckousik: Do we need to refresh a token? It could be useful when proxy secrets are rotated, but the webhooktunnel is only instantiated once when every worker starts ( https://github.com/taskcluster/taskcluster-worker/blob/master/worker/worker.go#L84 ), so I'm not quite sure how we would reconnect. A solution could be to pass this into the authorizer, but that would require refactoring of whclient.

jonasfj: Yeah, the authorizer should call auth using tc creds... that way it'll refresh if it gets a 401, right?

jonasfj:
@ckousik, I think you have push permission to `taskcluster-worker` repo... if you create a branch on this repo more tests will run :)

PR #295, opened 2017-07-19
Windows image example

jonasfj:


jonasfj:
@petemoore, I figured you might be interested in having a look... Basically `install.ps1` could probably drive whatever powershell install scripts we want to run, like OCC, if was tweaked run outside ec2.

Note: don't worry about tests, I'll fix this...

PR #296, opened 2017-07-25
Whtunnel auth

ckousik:
Final PR for tc-worker webhooktunnel integration. The worker can now use the new whclient API and auth endpoints to connect to the proxy over TLS.


runtime/webhookserver/config.go
```diff
@@ -23,6 +24,13 @@ var localtunnelConfigSchema = schematypes.Object{
 	Required: []string{"provider"},
 }
 
+var webhooktunnelConfigSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"provider": schematypes.StringEnum{Options: []string{"webhooktunnel"}},
```
walac: nit: add `Description` attribute to document what this configuration is about.

runtime/webhookserver/config.go
```diff
@@ -85,8 +94,9 @@ type Server interface {
 // NewServer returns a Server implementing WebHookServer, choosing the
 // implemetation based on the configuration passed in.
 //
-// Config passed must match ConfigSchema.
-func NewServer(config interface{}) (Server, error) {
+// Config passed must match ConfigSchema
```
walac: Any reason to remove end mark here?

runtime/webhookserver/webhooktunnel.go
```diff
@@ -0,0 +1,100 @@
...
+	configurer := func() (whclient.Config, error) {
+		authClient := auth.New(credentials)
+		whresp, err := authClient.WebhooktunnelToken()
+		if err != nil {
+			return whclient.Config{}, errors.New("could not get token from tc-auth")
+		}
+
+		// This hack is needed since proxyURL in auth config is set as a http url
+		proxyURL := whresp.ProxyURL
+		proxyURL = util.MakeWsURL(proxyURL)
```
walac: nit: why not pass `whresp.ProxyURL` directly to `MakeWsURL`?

runtime/webhookserver/webhooktunnel.go
```diff
@@ -0,0 +1,100 @@
...
+		defer wt.m.Unlock()
+		delete(wt.handlers, id)
+	}
+
+	return url, detach
+}
+
+// Stop will close the webhooktunnel client
+func (wt *WebhookTunnel) Stop() {
+	_ = wt.client.Close()
```
walac: I always like to be explicit to the reader that "No, I didn't forget to check the return error value, I just don't care about it" +1

runtime/webhookserver/webhooktunnel.go
```diff
@@ -0,0 +1,100 @@
...
+
+		// This hack is needed since proxyURL in auth config is set as a http url
+		proxyURL := whresp.ProxyURL
+		proxyURL = util.MakeWsURL(proxyURL)
+
+		return whclient.Config{
+			ID:        whresp.TunnelID,
+			ProxyAddr: proxyURL,
+			Token:     whresp.Token,
+			UseDomain: true,
```
jonasfj: Why do we need to pass this? Isn't it evident from the URL we get back from registration... and isn't the registration URL `ProxyAddr`?

jonasfj: This might be a question for `whclient`, more than tc-client...

runtime/webhookserver/webhooktunnel.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// AttachHook adds a new webhook to the server
+func (wt *WebhookTunnel) AttachHook(handler http.Handler) (string, func()) {
+	id := slugid.Nice()
+	wt.m.Lock()
+	wt.handlers[id] = handler
+	wt.m.Unlock()
+
+	url := wt.client.URL() + "/" + id + "/"
```
jonasfj: Shouldn't we ensure that we don't make double `/` in case URL already ends with slash?

runtime/webhookserver/webhooktunnel.go
ckousik: The registration url is `ProxyURL`, but the when I wrote the client I used `ProxyAddr`. `whclient` is not aware of `tc-auth`, so the ID, ProxyAddr, and Token need to be passed in explicitly. `UseDomain` was a simple way to switch between different types of hosting, but I wanted to keep the logic out of `whclient`.

runtime/webhookserver/webhooktunnel.go
ckousik: Client URL is guaranteed to not end with a double slash. Refer: https://github.com/taskcluster/webhooktunnel/blob/master/whproxy/proxy.go#L238-L247

runtime/webhookserver/webhooktunnel.go
jonasfj: I agree... why not put the call in `MakeWsURL` into `whclient.New()` so that it's invisible to us..

runtime/webhookserver/webhooktunnel.go
jonasfj: Can't the type of hosting... be decided server-side? It's the server-side that returns a URL that the proxy is exposed at..
All the client side needs is a URL to call for registration/out-going-web-socket... right?

ckousik: The return value of client.URL() also depends on the hosting method.

PR #301, opened 2017-08-09
Added Webhooktunnel Support

ckousik:
This patch allows the worker to use webhooktunnel to
serve proxy endpoints instead of opening public ports
on the worker.

Webhooktunnel allows http to be served over multiplexed
streams over a websocket. This is exposed as a net.Listener
interface by whclient, and can be used by an http.Server.

Webhooktunnel requires TC credentials for getting an ID and
Token from tc-auth. These can be used to connect to the proxy.

### DIFF
--- a/runtime/webhookserver/webhooktunnel.go
+++ b/runtime/webhookserver/webhooktunnel.go
@@ -1,0 +1,94 @@
+package webhookserver
+
+import (
+	"errors"
+	"net/http"
+	"sync"
+
+	"github.com/taskcluster/slugid-go/slugid"
+	"github.com/taskcluster/taskcluster-client-go"
+	"github.com/taskcluster/taskcluster-client-go/auth"
+	"github.com/taskcluster/webhooktunnel/whclient"
+)
+
+// WebhookTunnel wraps a whclient instance as a WebHookServer
+type WebhookTunnel struct {
+	m        sync.RWMutex
+	handlers map[string]http.Handler
+	// url is assigned when the client connects to the proxy
+	client *whclient.Client
+}
+
+// NewWebhookTunnel returns a pointer to a new WebhookTunnel instance
+func NewWebhookTunnel(credentials *tcclient.Credentials) (*WebhookTunnel, error) {
+	configurer := func() (whclient.Config, error) {
+		authClient := auth.New(credentials)
+		whresp, err := authClient.WebhooktunnelToken()
+		if err != nil {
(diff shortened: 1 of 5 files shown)


runtime/webhookserver/webhooktunnel.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, err
+	}
+
+	wt := &WebhookTunnel{
+		handlers: make(map[string]http.Handler),
+		client:   client,
+	}
+
+	go func() {
+		_ = http.Serve(wt.client, http.HandlerFunc(wt.handle))
```
jonasfj: Can we leave a TODO that an error here should be reported to sentry through monitor, when we finally add that as well...

I'm thinking you can do that in a follow up PR... if you have time...

runtime/webhookserver/webhooktunnel.go
```diff
@@ -0,0 +1,94 @@
...
+	return url, detach
+}
+
+// Stop will close the webhooktunnel client
+func (wt *WebhookTunnel) Stop() {
+	_ = wt.client.Close()
+}
+
+func (wt *WebhookTunnel) handle(w http.ResponseWriter, r *http.Request) {
+	id, path := r.URL.Path[1:23], r.URL.Path[23:]
```
walac: Ok, this seems a little bit scary, could you please add a comment explaining what `URL.Path` is supposed to be? Something like:

```go
// URL.Path format is "/" + slugId (23 characters) + <endopint path>
```

runtime/webhookserver/webhooktunnel.go
```diff
@@ -0,0 +1,94 @@
...
+
+	url := wt.client.URL() + "/" + id + "/"
+	detach := func() {
+		wt.m.Lock()
+		defer wt.m.Unlock()
+		delete(wt.handlers, id)
+	}
+
+	return url, detach
+}
```
walac: Wouldn't be better to return the ID and have a method to detach the hook? This seems more aligned to what the rest of the world would normally expect.

ckousik: This is required for implementing the WebhookServer interface.

jonasfj: Hmm, returning an ID or an object with a URL() method and a Detach() method might have been better... This is definitely my fault in designing the webhookserver interface...

Hmm, perhaps that's something we should refactor in a later PR... Returning an object with two methods would certainly make some of the code where we are using it simpler :)

PR #305, opened 2017-08-22
Qemu guest tools config

jonasfj:


commands/qemu-guest-tools/cmd.go
```diff
@@ -94,3 +129,28 @@ func (cmd) Execute(arguments map[string]interface{}) bool {
 
 	return true
 }
+
+func convertSimpleJSONTypes(val interface{}) interface{} {
```
imbstack: There's not a generic way to do something like this? Maybe a comment to explain what's going on here in general.

jonasfj: it's heartlessly stolen from `config/` package where I do the same thing... we should clean it up in some future for sure...

PR #306, opened 2017-08-25
use webhooktunnel [DO NOT MERGE]

jonasfj:
This is deployed as: "test-dummy-provisioner/dummy-worker-packet"

Example task: https://github.com/taskcluster/taskcluster-worker/blob/0c1b2150688f12c9f8234abbc2f159de1ded45b5/.taskcluster.yml#L176-L197

Adding: `task.payload.interactive = {}` will enable interactive features.

@ckousik, let's figure out why this doesn't work...
Server logs are in papertrail under "tc-worker-ci".

It's probably easiest to test this with `taskcluster-worker shell-server` locally, and just tweak it to use webhooktunnel.. Note that `taskcluster-worker shell wss://...` works with `taskcluster-worker shell-server`

### DIFF
--- a/examples/packet-config.yml
+++ b/examples/packet-config.yml
@@ -59,24 +59,7 @@ config:
     success:        {}
   temporaryFolder:  /mnt/tmp
   webHookServer:
-    provider:           stateless-dns
-    expiration:         '1 day'
-    serverIp:           {$packet: public-ipv4}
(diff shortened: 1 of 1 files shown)


ckousik:
This was a problem with forwarding headers on WS connections on whtunnel. Fixed it there.

PR #309, opened 2017-09-07
 System support groups

jonasfj:
@walac, what are your thoughts here?

Should we allow specification of user groups as this PR suggests?

Or should we just embed the listing of user-groups into the `system.StartProcess` call and always set user groups to whatever groups the provider `owner` has.. I'm tempted to do this instead, as I don't see us wanting to configure user-groups here... If we really care we just give the given user what ever groups it should have in system level configuration.

What is your thoughts?

I need this fixed as right now my `worker` user has group `worker` and no other groups... Hence, it can't user docker... in system config my worker user is member of a lot of groups, but when the task is run under `system.StartProcess` it only gets the `worker` group (which is the primary group).

Given that we don't make it optional specify what group is the primary group now.. maybe we should just always pass whatever groups a user is configured to have in system config... I'm not even sure you can do per-process groups on Windows anyways...

### DIFF
--- a/engines/native/system/process_posix.go
+++ b/engines/native/system/process_posix.go
@@ -150,7 +150,7 @@ func StartProcess(options ProcessOptions) (*Process, error) {
 		// accounts unable to execute process.
 		// If the passed owner matches the current user, we set owner to
 		// nil to allow non-root accounts to succeed.
-		if currentUser.uid == options.Owner.uid {
+		if currentUser.uid == options.Owner.uid && len(options.Groups) == 0 {
 			options.Owner = nil
 		}
 	}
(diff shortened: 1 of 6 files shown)


jonasfj:
I know I've changed my mind as I wrote the description of this PR... if you agree just r- this PR, and I'll fix it tomorrow...

jonasfj:
@petemoore, I would love your insights on this too? (I'm leaning towards saying we keep it simple and just use system groups as that'll definitely work across platforms)

jonasfj:
I changed my mind this is a bad idea... We can make simpler as outlined aboce...

PR #310, opened 2017-09-08
Find and use system groups when starting a process with an explicit o…

jonasfj:
…wner

### DIFF
--- a/engines/native/system/user_darwin.go
+++ b/engines/native/system/user_darwin.go
@@ -7,6 +7,7 @@ import (
 	"path"
 	"strconv"
 
+	"github.com/pkg/errors"
 	"github.com/taskcluster/slugid-go/slugid"
 )
 
@@ -14,8 +15,9 @@ const defaultShell = "/bin/bash"
 
 // User is a representation of a system user account.
 type User struct {
-	uid        uint32 // user id
-	gid        uint32 // primary group id
+	uid        uint32   // user id
+	gid        uint32   // primary group id
+	gids       []uint32 // users group memberships
 	name       string
 	homeFolder string
 	groups     []string // additional user groups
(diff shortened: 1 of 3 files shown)


jonasfj:
@walac, actually, please take a look at this... I wonder if setting `groups: nil` in `syscall.Credentials` haven't caused us issues when running as non-root on OS X?

engines/native/system/user_darwin.go
```diff
@@ -38,9 +40,24 @@ func CurrentUser() (*User, error) {
...
 
+	// Find group ids
+	sgids, err := osUser.GroupIds()
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to list groups for current user")
+	}
+	gids := make([]uint32, len(sgids))
+	for i, sgid := range sgids {
+		igid, err := strconv.Atoi(sgid)
+		gids[i] = uint32(igid)
```
walac: Shouldn't we assign after error checking? I know in practice it doesn't matter in this case, but "check error first, assign the value later" is a good coding practice.

engines/native/system/user_linux.go
```diff
@@ -133,7 +164,30 @@ func CreateUser(homeFolder string, groups []*Group) (*User, error) {
...
+		return nil, errors.Wrap(err, "Failed to list groups for current user")
+	}
+	gids := make([]uint32, len(sgids))
+	for i, sgid := range sgids {
+		igid, err := strconv.Atoi(sgid)
+		gids[i] = uint32(igid)
+		if err != nil {
+			panic(errors.Wrap(err, "Could not convert group id to integer"))
+		}
+	}
```
walac: This part of looking for a user's groups and transforming that to an array of integer gids should go in a internal utility function to avoid code duplication.

jonasfj: good point... I should stop my copy/pasta code...

PR #312, opened 2017-09-20
Cot support

jonasfj:
I'll probably get around to adding some test cases tomorrow...
But if you could give this a quick skim that would be great, thanks :)

### DIFF
--- a/plugins/artifacts/artifacts.go
+++ b/plugins/artifacts/artifacts.go
@@ -43,7 +43,6 @@ type plugin struct {
 	plugins.PluginBase
 	environment *runtime.Environment
 	privateKey  *openpgp.Entity // nil, if COT is disabled
-	config      config
 }
 
 type taskPlugin struct {
@@ -82,7 +81,7 @@ func (pluginProvider) NewPlugin(options plugins.PluginOptions) (plugins.Plugin, 
 			return nil, errors.Wrap(err, "Failed to load private key")
 		}
 		if len(keyring) != 1 {
-			return nil, fmt.Errorf("Expected exactly one private key found: %d", len(keyring))
+			return nil, fmt.Errorf("Expected exactly one private key, found: %d", len(keyring))
 		}
 		key = keyring[0]
 	}
@@ -90,7 +89,6 @@ func (pluginProvider) NewPlugin(options plugins.PluginOptions) (plugins.Plugin, 
 	return &plugin{
 		environment: options.Environment,
 		privateKey:  key,
-		config:      c,
 	}, nil
 }
 
@@ -101,24 +99,20 @@ func (p *plugin) PayloadSchema() schematypes.Object {
 		},
 	}
 	if p.privateKey != nil {
-		if !p.config.AlwaysCreateCOT {
-			schema.Properties["chainOfTrust"] = schematypes.Boolean{
-				Title: "Create chain-of-trust Certificate",
-				Description: util.Markdown(`
-					Generate a 'public/chainOfTrust.json.asc' artifact with signed hashes
-					of the artifacts generated from this task.
-				`),
-			}
+		schema.Properties["chainOfTrust"] = schematypes.Boolean{
+			Title: "Create chain-of-trust Certificate",
+			Description: util.Markdown(`
+				Generate a 'public/chainOfTrust.json.asc' artifact with signed hashes
+				of the artifacts generated from this task.
+			`),
 		}
-		if p.config.CreateCertifiedLog == optionOptional {
-			schema.Properties["certifiedLog"] = schematypes.Boolean{
-				Title: "Create Certified Log",
-				Description: util.Markdown(`
-					Default log artifact is not covered by 'public/chainOfTrust.json.asc',
-					if this is set to 'true' an artifact 'public/logs/certified.log' will
-					be created and covered by chain-of-trust certificate.
-				`),
-			}
+		schema.Properties["certifiedLog"] = schematypes.Boolean{
+			Title: "Create Certified Log",
+			Description: util.Markdown(`
+				Default log artifact is not covered by 'public/chainOfTrust.json.asc',
+				if this is set to 'true' an artifact 'public/logs/certified.log' will
+				be created and covered by chain-of-trust certificate.
+			`),
 		}
 	}
 	return schema
@@ -131,8 +125,8 @@ func (p *plugin) NewTaskPlugin(options plugins.TaskPluginOptions) (plugins.TaskP
 	return &taskPlugin{
 		plugin:       p,
 		artifacts:    P.Artifacts,
-		createCOT:    p.privateKey != nil && (p.config.AlwaysCreateCOT || P.CreateCOT),
-		certifiedLog: p.privateKey != nil && (p.config.CreateCertifiedLog == optionAlways || P.CertifiedLog),
+		createCOT:    p.privateKey != nil && P.CreateCOT,
+		certifiedLog: p.privateKey != nil && P.CertifiedLog,
 		uploaded:     make(map[string][]byte),
 		context:      options.TaskContext,
 		monitor:      options.Monitor,
--- a/plugins/artifacts/artifacts.go
+++ b/plugins/artifacts/artifacts.go
@@ -1,14 +1,24 @@
 package artifacts
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
+	"io"
 	"mime"
 	"path"
 	"path/filepath"
 	"strings"
 	"sync"
 
+	"golang.org/x/crypto/openpgp"
+	"golang.org/x/crypto/openpgp/clearsign"
+
+	"github.com/pkg/errors"
 	schematypes "github.com/taskcluster/go-schematypes"
 	"github.com/taskcluster/taskcluster-worker/engines"
 	"github.com/taskcluster/taskcluster-worker/plugins"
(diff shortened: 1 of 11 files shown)


plugins/artifacts/artifacts.go
```diff
@@ -102,6 +173,134 @@ func (tp *taskPlugin) Stopped(result engines.ResultSet) (bool, error) {
...
+			AdditionalHeaders: map[string]string{
+				"Content-Encoding": "gzip",
+			},
+		})
+		if err != nil {
+			err = errors.Wrap(err, "failed to upload certified.log")
+			tp.monitor.Error(err)
+			return runtime.ErrNonFatalInternalError // We don't expect upload errors to be fatal
+		}
+	}
```
walac: I am confused for what's the advantage of uploading the log file with a different name, but I guess this predates taskcluster-worker.

plugins/artifacts/artifacts.go
```diff
@@ -31,40 +41,101 @@ type pluginProvider struct {
...
+
+func (pluginProvider) NewPlugin(options plugins.PluginOptions) (plugins.Plugin, error) {
+	var c config
+	schematypes.MustValidateAndMap(configSchema, options.Config, &c)
+
+	var key *openpgp.Entity
+	if c.PrivateKey != "" {
+		keyring, err := openpgp.ReadArmoredKeyRing(bytes.NewBufferString(c.PrivateKey))
+		if err != nil {
+			return nil, errors.Wrap(err, "Failed to load private key")
```
walac: nit: add the `err` object to error strings. I can't count how many hours I lost because a log didn't have the original error message.

plugins/artifacts/artifacts.go
```diff
@@ -186,6 +385,11 @@ func (tp *taskPlugin) processFile(result engines.ResultSet, a artifact) {
 		mtype = unknownMimetype
 	}
 
+	// Compute artifact hash for chain-of-trust
+	if err = tp.hashArtifact(a.Name, r); err != nil {
+		goto error
```
walac: I am not an anti-goto person, but if we invert the logic to `err == nil` and move the `UploadS3Artifact` call into the `if` clause, I think we can get rid of it.

plugins/artifacts/artifacts.go
```diff
@@ -226,15 +431,25 @@ func (tp *taskPlugin) processDirectory(result engines.ResultSet, a artifact) {
 			mtype = unknownMimetype
 		}
 
+		// Construct artifact name
+		name := path.Join(a.Name, p)
+
+		var uerr error
+		// Compute artifact hash for chain-of-trust
+		if uerr = tp.hashArtifact(name, r); uerr != nil {
+			goto error
```
walac: Again, refactoring the logic a bit we can avoid the `goto`.

plugins/artifacts/config.go
```diff
@@ -0,0 +1,61 @@
+package artifacts
+
+import (
+	schematypes "github.com/taskcluster/go-schematypes"
+	"github.com/taskcluster/taskcluster-worker/runtime/util"
+)
+
+type config struct {
+	PrivateKey         string `json:"privateKey"`
+	AlwaysCreateCOT    bool   `json:"alwaysCreateCOT"`
```
walac: Is there a real use case scenario for this config? If it is a "nice to have" feature, I am in favor of removing it.

plugins/artifacts/config.go
```diff
@@ -0,0 +1,61 @@
...
+
+				If not given chain-of-trust signing will be disabled.
+			`),
+		},
+		"alwaysCreateCOT": schematypes.Boolean{
+			Title: "Always Create COT Certificate",
+			Description: util.Markdown(`
+				If set to 'true' all tasks will receive a chain-of-trust certificate.
+
+				Otherwise, chain-of-trust certicates is a feature that can enabled on
```
walac: nit: "can be enabled.."

plugins/artifacts/config.go
```diff
@@ -0,0 +1,61 @@
...
+				GPG armoured private key (unencrypted) for signing chain-of-trust
+				certificates.
+
+				If not given chain-of-trust signing will be disabled.
+			`),
+		},
+		"alwaysCreateCOT": schematypes.Boolean{
+			Title: "Always Create COT Certificate",
+			Description: util.Markdown(`
+				If set to 'true' all tasks will receive a chain-of-trust certificate.
```
walac: nit: comma after `'true'`

plugins/artifacts/config.go
```diff
@@ -0,0 +1,61 @@
...
+		configuration, such as private key.
+	`),
+	Properties: schematypes.Properties{
+		"privateKey": schematypes.String{
+			Title: "COT Private Key",
+			Description: util.Markdown(`
+				GPG armoured private key (unencrypted) for signing chain-of-trust
+				certificates.
+
+				If not given chain-of-trust signing will be disabled.
```
walac: nit: comma after `given`

plugins/artifacts/config.go
```diff
@@ -0,0 +1,61 @@
...
+const (
+	optionAlways   = "always"
+	optionOptional = "optional"
+	optionNever    = "never"
+)
+
+var configSchema = schematypes.Object{
+	Title: "Artifact Configuration",
+	Description: util.Markdown(`
+		Configuration for artifact plugin, this is mostly COT (chain-of-trust)
```
walac: nit: I think proper english mandates a period instead of a comma.

plugins/artifacts/artifacts.go
```diff
@@ -31,40 +41,101 @@ type pluginProvider struct {
...
+					of the artifacts generated from this task.
+				`),
+			}
+		}
+		if p.config.CreateCertifiedLog == optionOptional {
+			schema.Properties["certifiedLog"] = schematypes.Boolean{
+				Title: "Create Certified Log",
+				Description: util.Markdown(`
+					Default log artifact is not covered by 'public/chainOfTrust.json.asc',
+					if this is set to 'true' an artifact 'public/logs/certified.log' will
```
walac: nit: I am not 100% sure, but I think the phrase demands a comma after `'true'`

plugins/artifacts/payloadschema.go
```diff
@@ -22,44 +25,40 @@ const (
...
+	Title:       "Artifacts",
+	Description: "Artifacts to be published",
+	Items: schematypes.Object{
+		Properties: schematypes.Properties{
+			"type": schematypes.StringEnum{
+				Title: "Upload type",
+				Description: util.Markdown(`
+					Artifacts can be either an individual 'file' or a 'directory'
+					containing potentially multiple files with recursively included
+					subdirectories
```
walac: nit: period at the end of the phrase.

plugins/artifacts/payloadschema.go
```diff
@@ -22,44 +25,40 @@ const (
...
+				`),
+				Options: []string{typeFile, typeDirectory},
+			},
+			"path": schematypes.String{
+				Title:       "Artifact Path",
+				Description: "File system path of artifact",
```
walac: nit: "of the"

runtime/taskcontext.go
```diff
@@ -49,6 +49,7 @@ type TaskInfo struct {
 	Deadline time.Time
 	Expires  time.Time
 	Scopes   []string
+	Task     interface{} // task definition in it's entirety for COT
```
walac: I can't understand the comment

plugins/artifacts/artifacts.go
```diff
@@ -31,40 +41,101 @@ type pluginProvider struct {
...
-	mErrors     sync.Mutex                      // Guards errors
-	errors      []runtime.MalformedPayloadError // errors to be returned from Stopped()
-	nonFatalErr atomics.Bool                    // if true, Stopped() returns non-fatal error
-	fatalErr    atomics.Bool                    // if true, Stopped() returns fatal error
+	plugin       *plugin
+	context      *runtime.TaskContext
+	artifacts    []artifact
+	createCOT    bool
+	certifiedLog bool
+	uploaded     map[string][]byte // Map from artifact to sha256 hash
```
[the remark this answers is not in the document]
jonasfj: This is auto-formatted by gofmt... it will align if we add a comment on the line below...

plugins/artifacts/artifacts.go
```diff
@@ -31,40 +41,101 @@ type pluginProvider struct {
...
+	mUploaded    sync.Mutex
+	monitor      runtime.Monitor
+	failed       atomics.Bool                    // If true, Stopped() returns false
+	mErrors      sync.Mutex                      // Guards errors
+	errors       []runtime.MalformedPayloadError // errors to be returned from Stopped()
+	nonFatalErr  atomics.Bool                    // if true, Stopped() returns non-fatal error
+	fatalErr     atomics.Bool                    // if true, Stopped() returns fatal error
```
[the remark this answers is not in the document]
jonasfj: hmm, bools are easier to read... and setting `nonFatalErr` to true, after `fatalErr` is true won't overwrite `fatalErr`... So keeps the control flow simple..

Note: this is not new..

plugins/artifacts/artifacts.go
jonasfj: It's in there, the `errors.Wrap(err, ....)` does the right thing :)
In fact it shows some sort of trace, which is really nice...

Checkout: https://godoc.org/github.com/pkg/errors#example-Wrap--Extended

plugins/artifacts/artifacts.go
```diff
@@ -31,40 +41,101 @@ type pluginProvider struct {
...
+	schematypes.MustValidateAndMap(configSchema, options.Config, &c)
+
+	var key *openpgp.Entity
+	if c.PrivateKey != "" {
+		keyring, err := openpgp.ReadArmoredKeyRing(bytes.NewBufferString(c.PrivateKey))
+		if err != nil {
+			return nil, errors.Wrap(err, "Failed to load private key")
+		}
+		if len(keyring) != 1 {
+			return nil, fmt.Errorf("Expected exactly one private key found: %d", len(keyring))
```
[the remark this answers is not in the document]
jonasfj: Could also be less...

plugins/artifacts/artifacts.go
```diff
@@ -31,40 +41,101 @@ type pluginProvider struct {
...
+		if p.config.CreateCertifiedLog == optionOptional {
+			schema.Properties["certifiedLog"] = schematypes.Boolean{
+				Title: "Create Certified Log",
+				Description: util.Markdown(`
+					Default log artifact is not covered by 'public/chainOfTrust.json.asc',
+					if this is set to 'true' an artifact 'public/logs/certified.log' will
+					be created and covered by chain-of-trust certificate.
+				`),
+			}
+		}
```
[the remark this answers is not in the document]
jonasfj: it can certainly be configured that way :)

plugins/artifacts/artifacts.go
```diff
@@ -31,40 +41,101 @@ type pluginProvider struct {
...
+	schematypes.MustValidateAndMap(p.PayloadSchema(), options.Payload, &P)
 
 	return &taskPlugin{
-		artifacts: P.Artifacts,
-		context:   options.TaskContext,
-		monitor:   options.Monitor,
+		plugin:       p,
+		artifacts:    P.Artifacts,
+		createCOT:    p.privateKey != nil && (p.config.AlwaysCreateCOT || P.CreateCOT),
+		certifiedLog: p.privateKey != nil && (p.config.CreateCertifiedLog == optionAlways || P.CertifiedLog),
```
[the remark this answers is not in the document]
jonasfj: If there is no private key, the payload schema does not allow for requesting a COT certificate...

plugins/artifacts/artifacts.go
jonasfj: None... it's just that it's two different plugins, and it's what we do in other workers...

I would love to change artifact names, but it's probably a bit sketchy to do that :)

plugins/artifacts/artifacts.go
jonasfj: Nice, catch I used to have multiple things that had goto error, but now there is only one...

plugins/artifacts/config.go
jonasfj: Not sure, I think we'll just end up setting the config to the default behavior we want then never touch it again...

plugins/artifacts/artifacts.go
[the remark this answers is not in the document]
walac: Notice that in the integer proposal, ERROR_FATAL and NON_FATAL are not mutually exclusive, it is just how we check then, using `==` or `&`

plugins/artifacts/artifacts.go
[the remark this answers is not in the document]
walac: That's fine then

plugins/artifacts/artifacts.go
walac: Oh, my bad, pretend I didn't make this comment :)

plugins/artifacts/artifacts.go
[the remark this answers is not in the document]
walac: I think I was confused because now I see that phrase lacks a comma after `key` word.

plugins/artifacts/config.go
walac: My vote is to remove it, let's not add code candidate to be obsolete :)

plugins/artifacts/artifacts.go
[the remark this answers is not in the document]
walac: Ah, fine then.

plugins/artifacts/artifacts.go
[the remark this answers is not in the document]
jonasfj: ah

plugins/artifacts/config.go
jonasfj: okay, let's remove it...

PR #313, opened 2017-10-05
More error annotation to make bug 1405695 easier to find

jonasfj:


plugins/livelog/livelog.go
```diff
@@ -176,16 +176,16 @@ func (tp *taskPlugin) uploadLog() error {
 
 	zip := gzip.NewWriter(tempFile)
 	if _, err = io.Copy(zip, file); err != nil {
-		return err
+		return errors.Wrap(err, "failed to compressed log")
```
walac: nit: s/compressed/compress/

runtime/artifact.go
```diff
@@ -126,11 +127,11 @@ func (context *TaskContext) createArtifact(name string, req []byte) ([]byte, err
 func putArtifact(urlStr, mime string, stream ioext.ReadSeekCloser, additionalArtifacts map[string]string) error {
 	u, err := url.Parse(urlStr)
 	if err != nil {
-		return err
+		panic(errors.Wrap(err, "failed to parse URL"))
```
walac: That makes more sense in a panic context.

runtime/artifact.go
```diff
@@ -49,17 +51,16 @@ func (context *TaskContext) UploadS3Artifact(artifact S3Artifact) error {
 		StorageType: "s3",
 	})
 	if err != nil {
-		return err
+		panic(errors.Wrap(err, "failed to Marshal json that should have worked"))
```
walac: By "all input is evil" principle, we should not panic if we fail to process data coming from an external entity. We should log the errors message and fail the task nicely, or it may make difficult to diagnose what's going on if things start to explode in production. That said, do these failures go to sentry? If so, I think we are safe to panic.

runtime/artifact.go
```diff
@@ -49,17 +51,16 @@ func (context *TaskContext) UploadS3Artifact(artifact S3Artifact) error {
...
 	parsed, err := context.createArtifact(artifact.Name, req)
 	if err != nil {
 		return err
 	}
 	var resp queue.S3ArtifactResponse
-	err = json.Unmarshal(parsed, &resp)
-	if err != nil {
-		return err
+	if err = json.Unmarshal(parsed, &resp); err != nil {
+		panic(errors.Wrap(err, "failed to parse JSON that have been parsed before"))
```
walac: My "all input is evil" comment applies here too.

runtime/artifact.go
jonasfj: >By "all input is evil" principle, we should not panic if we fail to process data coming from an external entity. 

I mostly agree...

This isn't external...  If the type `queue.S3ArtifactRequest{...}` doesn't work with `json.Marshal(` then that's either always the case or never the case... IMO we could ignore the error because we know it either always happens or never happens... A panic is just a neat assertion here...

runtime/artifact.go
jonasfj: `parsed` is cast from `json.RawMessage` in `context.createArtifact` so we know it is valid JSON.
hence, this is really just an assertion... it should **never** be possible for this to fail..

runtime/artifact.go
jonasfj: What do you mean?

plugins/livelog/livelog.go
jonasfj: nice catch :)

runtime/artifact.go
jonasfj: Basically, I'm trying to say that if an error can never, then we should never propagate it...

I don't see anyway that external input could trigger an error here. So if an error happens we should panic.

staktrace:
@jonasfj Are you going to make a new release with these changes, or should I build from source? (I'd prefer the former)

jonasfj:
@staktrace, oh, yeah I'll tag off a release..
I'm pretty sure latest release wasn't even fully up to date...

PR #316, opened 2017-10-10
Fixed intermittent test case with interactive shell, should also give…

jonasfj:
… fewer error messages in production

### DIFF
--- a/plugins/interactive/shellhandler.go
+++ b/plugins/interactive/shellhandler.go
@@ -129,14 +129,14 @@ func (s *ShellHandler) abort() {
 	})
 }
 
-func (s *ShellHandler) send(message []byte) {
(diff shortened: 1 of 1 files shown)


plugins/interactive/shellhandler.go
```diff
@@ -354,7 +354,7 @@ func (s *ShellHandler) sendAcks() {
 
 		// Send an acknowledgment message (this is for congestion control)
 		binary.BigEndian.PutUint32(ack[2:], uint32(N))
-		s.send(ack)
+		s.send(ack, true)
```
jonasfj: If a program exists before stdin is closed, then we can't acknowledge the rest of stdin... but that's okay...

PR #319, opened 2017-10-31
Fix 409 error when reporting task result, in case task was cancelled

jonasfj:
Fixes: https://sentry.prod.mozaws.net/operations/taskcluster-worker/issues/678186/


jonasfj:
This is the intermittent errror I fixed in another PR just landed...

PR #320, opened 2017-11-03
Script engine tweaks

jonasfj:
Script engine isn't really used for anything... If you validate that this is okay because it only touches script engine that's okay :) hehe

### DIFF
--- a/docs/script-engine.md
+++ b/docs/script-engine.md
@@ -1,0 +1,104 @@
+---
+title: Script Engine
+---
+
+The script-engine allows for declarative tasks. This means that the worker
+is configured with a script and a payload schema. The payload is then passed
+to the script which is responsible for executing the payload. This is mostly
+in cases such as:
+
+ * Security sensitive operations, where tasks shouldn't be allowed to execute
+   arbitrary commands.
+ * Workers without isolation where only very specific hardcoded actions are
+   possible, and payload is used to inject arguments for said actions.
+
+In both cases the main motivation for using script-engine is that it is
+undesirable to have tasks execute arbitrary commands. Whether because there is
+insufficient isolation or executing security sensitive operations, preventing
+tasks from executing arbitrary commands is a good way to avoid tasks causing
+havoc on a worker.
+
+Worker Configuration
+====================
+
+script-engine is configured with a `command` and `schema` as illustrated below:
+
+```yaml
+engine: script
+engines:
+  script:
+    command:
+      - /usr/bin/bash
+      - /home/jonasfj/worker-script.sh
+    schema:
+      # type object is required for the root of the schema
+      type: object
+      properties:
+        binaryToSign: {type: 'string', format: 'uri'}
+        verbose: {type: 'boolean'}
+      required:
+        - binaryToSign
+plugins:
+  disabled:
+    # The following plugins are not useful when not executing arbitrary commands
+    # as the input is JSON and the script decides what artifacts to output.
+    # Similarly, there is no interactive support, when not executing arbitrary
+    # commands. Other plugins may or may not be sensible depending on use-case.
+    - artifacts
+    - env
+    - tcproxy
+    - interactive
+... # other worker configuration keys...
+```
+
+Tasks for this worker must now have a `task.payload` that satisfies the given
+`schema`. Notice that if other plugins are configured they may accept other
+properties on the `task.payload` and the worker will merge such schemas.
+An example task for the configuration above could look like:
+
+```js
+{
+  provisionerId: '...',
+  workerType: '...',
+  created: '...',
+  ... // other top-level keys
+  payload: {
+    binaryToSign: 'https://example.com/file.tar.gz',
+    verbose: true,  // this property is optional
+  },
+}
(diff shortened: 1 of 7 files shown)


engines/script/sandbox.go
```diff
@@ -58,7 +81,11 @@ func (s *sandbox) run() {
 			s.monitor.Errorf("Failed to remove temporary folder, error: %s", err)
 		}
 
-		s.resultSet = &resultSet{success: success}
+		if resultError == nil {
+			s.resultSet = &resultSet{success: success}
+		} else {
+			s.resultError = resultError
+		}
```
walac: I am confused here, shouldn't `result` set be included always set and in case `resultSet.sucess == false` `resultError` be set too?

docs/script-engine.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+Script Interface
+================
+
+In the configuration example above the command/script configured was
+`['/usr/bin/bash', '/home/jonasfj/worker-script.sh']`. This command will
+executed for each task with the following interface:
+
+ * **`stdin`** of the command will be feed the parts of `task.payload` matching
```
walac: s/be//

docs/script-engine.md
```diff
@@ -0,0 +1,104 @@
...
+   configured schema, after which `stdin` is closed.
+ * **Environment variables** `TASK_ID` and `RUN_ID` will be made available to
+   the command.
```
walac: nit: The command has access to the `TASK_ID` and `RUN_ID` variables.

docs/script-engine.md
```diff
@@ -0,0 +1,104 @@
...
+ * **Current working directory** for the command will set to a temporary folder
+   that will be deleted once the task is completed. This folder will contain
+   a `./artifacts/` folder that artifacts should be written to.
```
walac: nit: The engine sets a temporary folder as the current working directory and will delete it after it completes the task. This folder contains a  `./artifacts/` folder where the task artifacts live.

docs/script-engine.md
```diff
@@ -0,0 +1,104 @@
...
+   (notice that public artifacts must be written to `./artifacts/public/...`)
```
walac: nit: (notice that the task will write public artifacts to `./artifacts/public/...`)

docs/script-engine.md
```diff
@@ -0,0 +1,104 @@
...
+ * **`stdout`** of the command will exposed as task log.
```
walac: nit: "the task engine writes stdout to the task log."

docs/script-engine.md
```diff
@@ -0,0 +1,104 @@
...
+ * **`stderr`** of the command will have lines prefixed `[worker:error]` and
```
walac: nit: "the engine prefixes stderr lines with `[worker:error]` and injects that into the task log.

docs/script-engine.md
```diff
@@ -0,0 +1,104 @@
...
+   injected into the task log.
+ * **Signal `SIGKILL`** will be sent if the command should abort and clean-up
+   all sub-processes. Sent if the current task is aborted.
```
walac: nit: the command receives SIGKILL if it should abort and clean up all sub-processes.

docs/script-engine.md
```diff
@@ -0,0 +1,104 @@
...
+ * **Exit code** will be interpreted as follows:
```
walac: nit: The engine interprets exit codes as follows:

docs/script-engine.md
```diff
@@ -0,0 +1,104 @@
...
+   * `0`, task completed (success),
+   * `1`, task failed,
+   * `2`, task exception with reason: `malformed-payload`
+     (a message should have been written to stderr),
```
walac: nit: it should write a message to stderr

docs/script-engine.md
```diff
@@ -0,0 +1,104 @@
...
+   * `3`, task exception with reason: `internal-error`, but worker will continue
+     (a message should have been written to stderr),
+   * `4`, exception with reason: `internal-error`, and worker will terminated gracefully
+     (a message should have been written to stderr).
+   * Other error codes are reserved.
+
+Naturally, a script like `'/home/jonasfj/worker-script.sh'` should start by
+reading `stdin` until `EOF` and then parse the bytes read as UTF-8 encoded JSON.
```
walac: nit: s/parse/parses/

engines/script/sandbox.go
jonasfj: Nope, not if there is an internal error... then there is no `ResultSet`

docs/script-engine.md
jonasfj: I don't know what the task "will" do, but if it wants to write public artifacts they "must be written to"

Nit picking...

docs/script-engine.md
jonasfj: Okay, I recognized that I shoe horned the next a little as I wanted each bullet point to start with the command interface... then explaining how that interface is used to communicate...

PR #322, opened 2017-11-11
Superseding

jonasfj:
I've tried to make an implementation that can be ripped out, should we one day implement this at the queue level.

This makes it a tiny bit more ugly, but keeps the important code clean. Notably, I suspect we'll only use this for hardware workers.

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -318,7 +318,7 @@ func (w *Worker) processClaim(claim taskClaim) {
 	defer w.activeTasks.Decrement()
 
 	// If superseding is enabled, find superseding if one is available
-	// NOTE: This can be removed when some day superseding is implemented in the queue
+	// NOTE: This can be removed when superseding is implemented in the queue
 	if w.options.EnableSuperseding {
 		var done func()
 		claim, done = w.superseding(claim)
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -465,6 +465,9 @@ func (w *Worker) processClaim(claim taskClaim) {
 // superseding returns any superseding task, and a function to be called when
 // processed to resolve other superseded tasks.
 func (w *Worker) superseding(claim taskClaim) (taskClaim, func()) {
+	// Create monitor for any problems we run into
+	m := w.monitor.WithPrefix("superseding")
+
 	var payload struct {
 		SupersederURL string `json:"supersederUrl"`
 	}
@@ -482,7 +485,7 @@ func (w *Worker) superseding(claim taskClaim) (taskClaim, func()) {
 	g := got.New()
 	r, err := g.Get(payload.SupersederURL + "?taskId=" + claim.Status.TaskID).Send()
 	if err != nil {
-		// TODO: warn logging
+		m.Warnf("failed to contact supersederUrl: '%s'", payload.SupersederURL)
 		return claim, func() {}
 	}
 
@@ -490,7 +493,7 @@ func (w *Worker) superseding(claim taskClaim) (taskClaim, func()) {
 		Supersedes []string `json:"supersedes"`
 	}
 	if json.Unmarshal(r.Body, &result) != nil {
-		// TODO: warn logging
+		m.Warnf("failed to JSON parse result from supersederUrl: '%s'", payload.SupersederURL)
 		return claim, func() {}
 	}
 
@@ -499,7 +502,7 @@ func (w *Worker) superseding(claim taskClaim) (taskClaim, func()) {
 		taskID := result.Supersedes[i]
 		s, qerr := w.queue.Status(taskID)
 		if qerr != nil {
-			// TODO: info logging
+			m.WithTag("taskId", taskID).Infof("unable to get task status, error: %v", qerr)
 			return
 		}
 		runID := len(s.Status.Runs) - 1
@@ -511,7 +514,10 @@ func (w *Worker) superseding(claim taskClaim) (taskClaim, func()) {
 			WorkerGroup: w.options.WorkerGroup,
 		})
 		if qerr != nil {
-			// TODO: info logging
+			m.WithTags(map[string]string{
+				"taskId": taskID,
+				"runId":  strconv.Itoa(runID),
+			}).Debug("unable to claimTask from superseder, error: %v", qerr)
 			return
 		}
 		claimAttempts[i] = taskClaim(*c)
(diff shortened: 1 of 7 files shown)


worker/worker.go
```diff
@@ -445,6 +462,177 @@ func (w *Worker) processClaim(claim taskClaim) {
...
+
+	// Remove supersederUrl from payload, so that it doesn't create malformed-payload
+	payload = make(map[string]interface{})
+	if json.Unmarshal(claim.Task.Payload, &payload) == nil {
+		delete(payload, "supersederUrl")
+		claim.Task.Payload, err = json.Marshal(payload)
+		if err != nil {
+			panic(errors.Wrap(err, "failed to serialize data known to be JSON"))
+		}
+	}
```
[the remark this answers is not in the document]
jonasfj: Yeah, I considered it.... but it just seems overkill.. This way I get to keep all the ugly superseded stuff in a single method rather than creating a lot of ugly functions...

jonasfj:
@petemoore, would you mind taking a quick look at this...

Maybe you can review the integration tests to see if what I'm testing is correct...

worker/worker.go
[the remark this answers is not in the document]
walac: Fair enough.

worker/worker.go
```diff
@@ -445,6 +462,179 @@ func (w *Worker) processClaim(claim taskClaim) {
...
+	var payload map[string]interface{}
+	if json.Unmarshal(claim.Task.Payload, &payload) != nil {
+		// Do nothing, if there is an error, it'll show up later and logging will be
+		// more natural.
+		return claim, func() {}
+	}
+
+	// Take supersederUrl out of the payload, as it would break the payload
+	// validation done in TaskRun. We attempt to hide superseding from the rest
+	// of the worker implementation, so that it's only creating a hack here.
```
petemoore: How come it would break the TaskRun validation? This sounds pretty hacky.

worker/worker.go
```diff
@@ -445,6 +462,179 @@ func (w *Worker) processClaim(claim taskClaim) {
...
+		m.Warnf("failed to contact supersederUrl: '%s'", supersederURL)
+		return claim, func() {}
+	}
+
+	var result struct {
+		Supersedes []string `json:"supersedes"`
+	}
+	if json.Unmarshal(r.Body, &result) != nil {
+		m.Warnf("failed to JSON parse result from supersederUrl: '%s'", supersederURL)
+		return claim, func() {}
```
petemoore: We should report this to sentry. Does `Warnf` take care of that?

worker/worker.go
```diff
@@ -445,6 +462,179 @@ func (w *Worker) processClaim(claim taskClaim) {
...
+	}
+
+	claimAttempts := make([]taskClaim, len(result.Supersedes))
+	util.Spawn(len(result.Supersedes), func(i int) {
```
petemoore: What is `util.Spawn`? Can't we just use native go routines here?

worker/worker.go
```diff
@@ -445,6 +462,179 @@ func (w *Worker) processClaim(claim taskClaim) {
...
+			m.WithTags(map[string]string{
+				"taskId": taskID,
+				"runId":  strconv.Itoa(runID),
+			}).Debug("unable to claimTask from superseder, error: %v", qerr)
+			return
+		}
+		claimAttempts[i] = taskClaim(*c)
+	})
+
+	// remove invalid claims
```
petemoore: What causes an invalid claim?

worker/worker.go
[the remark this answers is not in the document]
petemoore: > Yeah, I considered it.... but it just seems overkill.. This way I get to keep all the ugly superseded stuff in a single method rather than creating a lot of ugly functions...

A [function literal](https://golang.org/ref/spec#Function_literals) would solve that.

worker/worker.go
jonasfj: It validates the payload against the payload schema...

And it forbids additional properties.

This is all me ensuring that these superseded properties are restricted to this file and function... So it's easy to rip out and kill this, when someday we have superseded in the queue.

worker/worker.go
jonasfj: Nope, and no... We don't do that in docker worker... we could print it to task log.. but at this point I'm okay with a less than ideal solution.

I want to rip it out later anyways.

worker/worker.go
jonasfj: It's just a short hand... this:
```go
util.Spawn(len(result.Supersedes), func(i int) {
  ... // code goes here
})
```
is the same as writing:
```go
N := len(result.Supersedes)
var wg sync.WaitGroup
wg.Add(N)
for j := 0; j < N; j++ {
  go (func(i int) {
      defer wg.Done()
      ... // code goes here
  })(j)
}
wg.Wait()
```

It's super elegant when you have to do something for each element in an array in parallel.

```go
// Let's say I have
input := []string{"some", "strings", ...}
// And I want output[i] = slow_function(input[i])
// I would just do:
output := make([]string, len(input))
util.Spawn(len(input), func(i int) {
  output[i] = slow_function(input[i])
})
```

Where as the other thing above becomes very long and easy to get wrong or make a mistake with.

worker/worker.go
jonasfj: task was claimed by another worker, so it returned 409, hence, the claim object is empty... We just want those filtered out.

worker/worker.go
```diff
@@ -445,6 +462,177 @@ func (w *Worker) processClaim(claim taskClaim) {
...
+	// Take the last claim we have, as the task we run
+	claim = claims[len(claims)-1]
+	claims = claims[:len(claims)-1]
+
+	// Start a reclaiming loop, and finish off by resolving superseded
+	var stopReclaiming atomics.Once
+	var tasksResolved atomics.WaitGroup
+	tasksResolved.Add(len(claims))
+	ctx, cancel := context.WithCancel(context.Background())
+	go util.Spawn(len(claims), func(i int) {
```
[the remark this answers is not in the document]
jonasfj: We spawn a reclaim loop for each claim... these loops have to run asynchronously... As this method returns the claim that should be processed by the worker.

worker/worker.go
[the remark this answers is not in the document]
jonasfj: > A function literal would solve that.

Yes, again, I don't care... the less is contained in a single function we can easily delete.

PR #329, opened 2018-02-06
Add syslog name to errors

jonasfj:
Improving error messages from syslog errors...

----
I figure no harm can come from adding additional information.


jonasfj:
hmm, clearly CI is broken... or this broke something on OS X...

jonasfj:
I fixed CI in #332

PR #330, opened 2018-02-09
Fix packet config to use webhooktunnel

jonasfj:


jonasfj:
Not really anything to review, just need to land this... using the PR to see if CI works, while I'm at it :)

PR #331, opened 2018-02-23
[WIP] Docker Engine

ckousik:
This pull request aims to provide a minimal implementation of docker-engine for taskcluster-worker using the golang docker client.

### DIFF
--- a/engines/docker/config.go
+++ b/engines/docker/config.go
@@ -1,0 +1,26 @@
+package dockerengine
+
+import (
+	schematypes "github.com/taskcluster/go-schematypes"
+)
+
+type configType struct {
+	DockerSocket   string `json:"dockerSocket"`
+	MaxConcurrency int    `json:"maxConcurrency"`
+}
+
+var configSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"dockerSocket": schematypes.String{
+			Title: "Docker Endpoint",
+			Description: "dockerEndpoint is the endpoint to use for communicating\n" +
+				"with the Docker daemon.",
+		},
+		"maxConcurrency": schematypes.Integer{
+			Title: "Max Concurrency",
+			Description: "maxConcurrency defines the maximum number of tasks \n" +
+				"that may run concurrently on the worker.",
+			Minimum: 1,
+		},
+	},
+}
--- a/engines/docker/config.go
+++ b/engines/docker/config.go
@@ -2,25 +2,21 @@ package dockerengine
 
 import (
 	schematypes "github.com/taskcluster/go-schematypes"
+	"github.com/taskcluster/taskcluster-worker/runtime/util"
 )
 
 type configType struct {
-	DockerSocket   string `json:"dockerSocket"`
-	MaxConcurrency int    `json:"maxConcurrency"`
+	DockerSocket string `json:"dockerSocket"`
 }
 
 var configSchema = schematypes.Object{
 	Properties: schematypes.Properties{
 		"dockerSocket": schematypes.String{
-			Title: "Docker Endpoint",
-			Description: "dockerEndpoint is the endpoint to use for communicating\n" +
-				"with the Docker daemon.",
-		},
-		"maxConcurrency": schematypes.Integer{
-			Title: "Max Concurrency",
-			Description: "maxConcurrency defines the maximum number of tasks \n" +
-				"that may run concurrently on the worker.",
-			Minimum: 1,
+			Title: "Docker Socket",
+			Description: util.Markdown(`Path to the docker socket hosting the remote docker API.
+			
+			If not given the default value 'unix:///var/run/docker.sock' will be used.
+			`),
 		},
 	},
 }
--- a/engines/docker/engine.go
+++ b/engines/docker/engine.go
@@ -1,11 +1,12 @@
 package dockerengine
 
 import (
+	"sync"
+
 	docker "github.com/fsouza/go-dockerclient"
 	schematypes "github.com/taskcluster/go-schematypes"
 	"github.com/taskcluster/taskcluster-worker/engines"
 	"github.com/taskcluster/taskcluster-worker/runtime"
-	"sync"
 )
 
 type engine struct {
@@ -21,6 +22,10 @@ type engine struct {
 
 type engineProvider struct {
 	engines.EngineProviderBase
+}
+
+func init() {
+	engines.Register("docker", engineProvider{})
 }
 
 type configType struct {
@@ -45,7 +50,7 @@ var configSchema = schematypes.Object{
 		},
 	},
 	Required: []string{
-		"socketPath",
+		"dockerEndpoint",
 	},
 }
 
@@ -105,7 +110,7 @@ func (e *engine) NewSandboxBuilder(options engines.SandboxOptions) (engines.Sand
 		return nil, engines.ErrMaxConcurrencyExceeded
 	}
 	e.running += 1
-	return nil, nil
+	return newSandboxBuilder(&p, e, e.Environment.Monitor, options.TaskContext), nil
 }
 
 func (e *engine) Dispose() error {
--- a/engines/docker/engine.go
+++ b/engines/docker/engine.go
@@ -8,7 +8,6 @@ import (
 	"github.com/taskcluster/taskcluster-worker/engines"
 	"github.com/taskcluster/taskcluster-worker/runtime"
 	"github.com/taskcluster/taskcluster-worker/runtime/caching"
-	"github.com/taskcluster/taskcluster-worker/runtime/gc"
 )
 
 type engine struct {
@@ -29,35 +28,7 @@ type engineProvider struct {
 }
 
 func init() {
-	engines.Register("docker", engineProvider{
-		cache: caching.New(imageConstructor, true, &gc.GarbageCollector{}),
-	})
-}
-
-type configType struct {
-	DockerEndpoint string `json:"dockerEndpoint"`
-	MaxConcurrency int    `json:"maxConcurrency"`
-}
-
-var configSchema = schematypes.Object{
-	Properties: schematypes.Properties{
-		"dockerEndpoint": schematypes.String{
-			Title: "Docker Endpoint",
-			Description: "dockerEndpoint is the endpoint to use for communicating\n" +
-				"with the Docker daemon.",
-			//TODO: Add pattern for docker socket
-		},
-		"maxConcurrency": schematypes.Integer{
-			Title: "Max Concurrency",
-			Description: "maxConcurrency defines the maximum number of tasks \n" +
-				"that may run concurrently on the worker.",
-			Minimum: 0,
-			Maximum: 10,
-		},
-	},
-	Required: []string{
-		"dockerEndpoint",
-	},
+	engines.Register("docker", engineProvider{})
 }
 
 func (p engineProvider) ConfigSchema() schematypes.Schema {
@@ -68,8 +39,14 @@ func (p engineProvider) NewEngine(options engines.EngineOptions) (engines.Engine
 	debug("docker engineProvider.NewEngine()")
 	var c configType
 	schematypes.MustValidateAndMap(configSchema, options.Config, &c)
+	if c.DockerSocket == "" {
+		c.DockerSocket = "unix:///var/run/docker.sock"
+	}
+	if c.MaxConcurrency == 0 {
+		c.MaxConcurrency = 1
+	}
 
-	client, err := docker.NewClient(c.DockerEndpoint)
+	client, err := docker.NewClient(c.DockerSocket)
 	if err != nil {
 		return nil, err
 	}
@@ -80,7 +57,7 @@ func (p engineProvider) NewEngine(options engines.EngineOptions) (engines.Engine
 		Environment:    options.Environment,
 		monitor:        options.Monitor,
 		maxConcurrency: c.MaxConcurrency,
-		cache:          p.cache,
+		cache:          caching.New(imageConstructor, true, options.Environment.GarbageCollector),
 		running:        0,
 	}, nil
 }
@@ -114,10 +91,10 @@ func (e *engine) NewSandboxBuilder(options engines.SandboxOptions) (engines.Sand
 	schematypes.MustValidateAndMap(payloadSchema, options.Payload, &p)
 	e.m.Lock()
 	defer e.m.Unlock()
-	if e.maxConcurrency == e.running {
-		return nil, engines.ErrMaxConcurrencyExceeded
-	}
-	e.running += 1
+	// if e.maxConcurrency == e.running {
+	// 	return nil, engines.ErrMaxConcurrencyExceeded
+	// }
+	// e.running += 1
 	return newSandboxBuilder(&p, e, e.Environment.Monitor, options.TaskContext), nil
 }
 
--- a/engines/docker/engine.go
+++ b/engines/docker/engine.go
@@ -1,0 +1,113 @@
+package dockerengine
+
+import (
+	docker "github.com/fsouza/go-dockerclient"
+	schematypes "github.com/taskcluster/go-schematypes"
+	"github.com/taskcluster/taskcluster-worker/engines"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+	"sync"
+)
+
+type engine struct {
+	engines.EngineBase
+	m              sync.Mutex
+	Environment    *runtime.Environment
+	client         *docker.Client
+	monitor        runtime.Monitor
+	maxConcurrency int
+	engineConfig   configType
+	running        int
+}
+
+type engineProvider struct {
+	engines.EngineProviderBase
+}
+
+type configType struct {
+	DockerEndpoint string `json:"dockerEndpoint"`
+	MaxConcurrency int    `json:"maxConcurrency"`
+}
+
+var configSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"dockerEndpoint": schematypes.String{
+			Title: "Docker Endpoint",
+			Description: "dockerEndpoint is the endpoint to use for communicating\n" +
+				"with the Docker daemon.",
+			//TODO: Add pattern for docker socket
+		},
+		"maxConcurrency": schematypes.Integer{
+			Title: "Max Concurrency",
+			Description: "maxConcurrency defines the maximum number of tasks \n" +
+				"that may run concurrently on the worker.",
+			Minimum: 0,
+			Maximum: 10,
+		},
+	},
+	Required: []string{
+		"socketPath",
+	},
+}
+
+func (p engineProvider) ConfigSchema() schematypes.Schema {
+	return configSchema
+}
+
+func (p engineProvider) NewEngine(options engines.EngineOptions) (engines.Engine, error) {
+	var c configType
+	schematypes.MustValidateAndMap(configSchema, options.Config, &c)
+
+	client, err := docker.NewClient(c.DockerEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	return &engine{
+		engineConfig:   c,
+		client:         client,
+		Environment:    options.Environment,
+		monitor:        options.Monitor,
+		maxConcurrency: c.MaxConcurrency,
+		running:        0,
+	}, nil
+}
+
+type payloadType struct {
+	Image   imageType `json:"image"`
+	Command []string  `json:"command"`
+}
+
+var payloadSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"image": imageSchema,
+		"command": schematypes.Array{
+			Title:       "Command",
+			Description: "Command to run inside the container.",
+			Items:       schematypes.String{},
+		},
+	},
+	Required: []string{
+		"image",
+		"command",
+	},
+}
+
+func (e *engine) PayloadSchema() schematypes.Object {
+	return payloadSchema
+}
+
+func (e *engine) NewSandboxBuilder(options engines.SandboxOptions) (engines.SandboxBuilder, error) {
+	var p payloadType
+	schematypes.MustValidateAndMap(payloadSchema, options.Payload, &p)
+	e.m.Lock()
+	defer e.m.Unlock()
+	if e.maxConcurrency == e.running {
+		return nil, engines.ErrMaxConcurrencyExceeded
+	}
+	e.running += 1
+	return nil, nil
+}
+
+func (e *engine) Dispose() error {
+	return nil
+}
--- a/engines/docker/engine.go
+++ b/engines/docker/engine.go
@@ -12,19 +12,17 @@ import (
 
 type engine struct {
 	engines.EngineBase
-	m              sync.Mutex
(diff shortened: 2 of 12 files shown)


engines/docker/image_cache.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (im *imageResource) Dispose() error {
+	im.dispose.Do(func() {
+		client, err := docker.NewClientFromEnv()
+		if err != nil {
+			debug("%v", err)
+			return
+		}
+		err = client.RemoveImage(buildImageName(im.Repository, im.Tag))
+		if err != nil {
```
jonasfj: This error we should report to sentry... and return `runtime.ErrNonFatalInternalError` (I think).

We can report to sentry with monitor.. if like suggested above we keep a pointer to the `engine` structure we can use a monitor on that structure.

engines/docker/image_cache.go
```diff
@@ -0,0 +1,103 @@
...
+			debug("%v", err)
+		}
+	})
+	return nil
+}
+
+func imageConstructor(ctx caching.Context, opts interface{}) (caching.Resource, error) {
+	options := opts.(imageType)
+	client, err := docker.NewClientFromEnv()
```
jonasfj: Ah, I see, you think you can't pass in properties like `*engine` through `imageType`, you can, as long as it has an annotation that makes `json.Marshal` ignore it...

jonasfj: Example:
```go
// imageOptions are passed to imageConstructor through *caching.Cache to provide options for creating an image
type imageOptions struct {
  Image  imageType `json:"image"` // options from task.payload
  Engine *engine   `json:"-"`     // engine reference w. monitor and docker client that's valid for the life-cycle of the image 
}
```

So long as we use `json:"-"` the property will be ignored by `json.Marshal` so when `caching.Cache` computes a hash value for the image (in-order to avoid duplicate images), it won't hash the value of the `*engine` pointer (which would be a true mess).

jonasfj: This trick is useful for passing in references to things like docker client, and monitor that we might want to use later.. But that doesn't determine the identity of the image being downloaded..

engines/docker/image_cache.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	debug("pulling image %s %s", options.Repository, options.Tag)
+	err = client.PullImage(docker.PullImageOptions{
+		Repository:        options.Repository,
+		Tag:               options.Tag,
+		InactivityTimeout: dockerPullImageInactivityTimeout,
+		Context:           ctx,
```
jonasfj: Add a TODO to use `OutputStream` to read progress and report it to `ctx`.
(this is a nice to have, we can develop a helper function that can do this for us later, in another PR).

engines/docker/image_cache.go
```diff
@@ -0,0 +1,103 @@
...
+	}, docker.AuthConfiguration{})
+	if err != nil {
+		debug("error pulling image %v", err)
+		return nil, err
```
jonasfj: I LOVE that you remembered to convert this into a `NewMalformedPayloadError` later...
I'm thinking it might be neat to do it here... that way we just return whatever error we get from pulling the image...

Also if we have an internal error of some sort we can report it and return `ErrNonFatalInternalError` right now, returning any error just becomes a malformed payload error... (which definitely should be the default).

Mostly, I just like having the option of making better errors in the future... In particularly if we ever get around to read the status feed from docker. it might include things like errors with more specific messages, like if image is private, etc. (let's not do that now :) we can have TODO for that)

jonasfj:
@ckousik, I'll try to cleanup tests for taskcluster-worker, then you can merge/rebase, it might be a few days before I figure what is wrong :(

engines/docker/image_cache.go
ckousik: Good point...ctx is a *caching.Context here so this may not be too hard to do in the future

engines/docker/config.go
```diff
@@ -0,0 +1,26 @@
...
+			Title: "Docker Endpoint",
+			Description: "dockerEndpoint is the endpoint to use for communicating\n" +
+				"with the Docker daemon.",
+		},
+		"maxConcurrency": schematypes.Integer{
+			Title: "Max Concurrency",
+			Description: "maxConcurrency defines the maximum number of tasks \n" +
+				"that may run concurrently on the worker.",
+			Minimum: 1,
+		},
```
jonasfj: @ckousik, regarding deployment... let's largely try to leave `// NOTE: ` comments with notes on potential problems.
For qemu-engine it was really neat to deploy tc-worker with docker, because using qemu-engine requires all these external dependencies like qemu, iptables, qemu-img, etc. for network configuration and setup.

For docker-engine we so far don't need any external binaries other than docker. So maybe we can say that deploying docker-engine is:
 * install in a `config.yml` file
 * install `taskcluster-worker`
 * add script to run `taskcluster-worker` as a service (or whatever)

If we only have to install `taskcluster-worker` as a single binary, and we don't have a long list of external dependencies... then maybe it's easier to not use docker...

If we need a lot of external things, we needed a few in docker-worker: https://github.com/taskcluster/docker-worker/tree/master/bin-utils
Then maybe we reconsider... it's also possible that we can embed such resources as binary blobs into the `taskcluster-worker` binary, or some other hack :) Maybe we package them as a tarball we ship along with taskcluster-worker and in config for the engine we specify where to find it (who knows).

It's also possible that we support engine config for specifying where the system root folder is mounted in the docker container that runs taskcluster-worker. That could totally be an engine config option.

-----
But the more I think about it, we might be better off not running taskcluster-worker from inside docker. As we minimize the risk that we remove the container we're running inside, or the image, or something stupid :) hehe

Finally, for volumes and proxies having access to the root system might be a lot easier than some complicated mapping of ports and file system folders.

engines/docker/engine.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+type engine struct {
+	engines.EngineBase
+	m              sync.Mutex
+	Environment    *runtime.Environment
+	client         *docker.Client
+	monitor        runtime.Monitor
+	maxConcurrency int
+	engineConfig   configType
```
jonasfj: You don't need both `configType` and `maxConcurrency`...
I would just add a single property:
```go
   config configType
```
since you're already inside the `engine` struct, there is no need to  call it `engineConfig`

engines/docker/config.go
```diff
@@ -0,0 +1,26 @@
...
+type configType struct {
+	DockerSocket   string `json:"dockerSocket"`
+	MaxConcurrency int    `json:"maxConcurrency"`
+}
+
+var configSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"dockerSocket": schematypes.String{
+			Title: "Docker Endpoint",
+			Description: "dockerEndpoint is the endpoint to use for communicating\n" +
```
jonasfj: Try this:
```go
  Description: util.Markdown(`
    Path to the docker socket hosting the docker remote API.

    If not given default value 'unix:///var/run/docker.sock' will be used.
  `),
```

The `util.Markdown` function removes the indentation... so the strings looks nice. And it changes `'` into back-ticks that will be formatted with `<code>` in markdown. I know it wasn't smart to use single quotes for back-ticks, but I'll find a better solution in the future... We write markdown in JSON schema descriptions, in a not so distant future we'll render this on the docs site :)

This should encourage useful descriptions, so people can do more on their own...

engines/docker/image_cache.go
```diff
@@ -0,0 +1,108 @@
...
+	"time"
+
+	docker "github.com/fsouza/go-dockerclient"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+	"github.com/taskcluster/taskcluster-worker/runtime/atomics"
+	"github.com/taskcluster/taskcluster-worker/runtime/caching"
+	"github.com/taskcluster/taskcluster-worker/runtime/gc"
+)
+
+const dockerPullImageInactivityTimeout = 30 * time.Second
```
jonasfj: please give this a doc comment explaining what it does... I assume it has to do with docker client assuming docker is failing if the response locks up...

engines/docker/image_cache.go
```diff
@@ -0,0 +1,108 @@
...
+func (ctx *cacheContext) Progress(description string, percent float64) {
+	debug("progress: %s %f", description, percent)
+}
+
+func (im *imageResource) MemorySize() (uint64, error) {
+	return im.Size, nil
+}
+
+func (im *imageResource) DiskSize() (uint64, error) {
+	return 0, gc.ErrDisposableSizeNotSupported
```
jonasfj: This is where you want to return `im.Size` :)

engines/docker/image_cache.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (im *imageResource) Dispose() error {
+	var err error
+	im.dispose.Do(func() {
+		err = im.client.RemoveImage(buildImageName(im.Repository, im.Tag))
```
jonasfj: We should probably do some testing on this...

In this case `Tag` is a hash, right?

We could have a scenario where:
`hashA <- hashB <- hashC`

So `hashA` is the base image, `hashB` is built from A and `hashC` is build from B.

cases to consider:
 * If I fetch `hashB` I'll get both `hashA` and `hashB`... but if I delete `hashB`, do I get both removed?
 * If I fetch `hashC` I'll get both `hashA`, `hashB` and `hashC`, if I then fetch `hashB` and remove `hashC` will I remove `hashB` ?

We should probably double check this... and check if docker actually removes all the intermediary image layers...
I think docker tracks which image layers where explicitly pulled and has to be explicitly removed, before the deletes the underlying layers... So I think it works... but we should probably double check :)

Maybe write a comment here on what behavior we have from docker, and why this works... even though we might have two `imageResource`s that uses the same base image layers... and that those base image layers won't disappear until they are both remoted... (ie. that docker does some reference counting for us)

engines/docker/resultset.go
```diff
@@ -0,0 +1,154 @@
...
+		tempfile, err := r.tempStorage.NewFile()
+		if err != nil {
+			return engines.ErrResourceNotFound
+		}
+		if _, err = io.Copy(tempfile, reader); err != nil {
+			return engines.ErrResourceNotFound
+		}
+
+		defer func() {
+			_ = tempfile.Close()
```
jonasfj: None of these errors can be ignored... this one we can plausibly consider a warning...
Since it was the responsibility of the `handler` to upload it or read it and detect errors from that.

But we should report this to sentry as a warning, tag the event with filename.. and use the monitor from the `NewSandboxBuilder` (I think).
```go
monitor := ...monitor.withPrefix('extract-folder')

....

monitor.withTag('filename', hdr.Name).reportWarning(err, "failed to close temporary file, if we failed to delete the file we might run out of diskspace")
```
This way we know reasonably well where we got the error... and we can monitor the issue..

---
Actually, if you read:
 https://github.com/taskcluster/taskcluster-worker/blob/master/engines/resultset.go#L61-L64

It says we can call concurrently, and that `handler` must call `close`

Maybe just leave the call concurrent part as a `// TODO` for now :)

engines/docker/resultset.go
```diff
@@ -0,0 +1,154 @@
...
+		}()
+
+		fname := strings.TrimPrefix(hdr.Name, strip)
+		tempfile.Seek(0, 0)
```
jonasfj: We can't ignore the error... We probably want to do:
`monitor.withTag('filename', hdr.Name).reportError(err, "failed to seek on temporary file")`

And return `runtime.ErrNonFatalInternalError`

engines/docker/resultset.go
```diff
@@ -0,0 +1,154 @@
...
+	if r.tempStorage == nil {
+		return nil, engines.ErrResourceNotFound
+	}
+	tempfile, err := r.tempStorage.NewFile()
+	if err != nil {
+		return nil, runtime.ErrNonFatalInternalError
+	}
+
+	opts := docker.DownloadFromContainerOptions{
+		OutputStream:      tempfile,
```
jonasfj: We don't need to do this... we can use `io.Pipe()`... and that way read the tarfile as it arrives from the write... this saves us one trip to the disk :)

I'm also okay with leaving that as a `TODO`, after all this is just optimizations.

engines/docker/image_cache.go
ckousik: This was confusing, because it looked like this required the size of the disk and not the size of the resource on disk. Should we consider calling this SizeOnDisk() ?

engines/docker/image_cache.go
ckousik: `Remove an image, along with any untagged parent images that were referenced by that image.

Images can't be removed if they have descendant images, are being used by a running container or are being used by a build.
`
So, if `hashA`<-`hashB`<-`hashC`, `hashB` cannot be removed since it has a descendant image `hashC`. In the second case, removing `hashC` should remove `hashB` if it has no other descendant images, but if we have fetched `hashB` and are using it in a container, then it won't be removed. Please note that I am not using the Force flag when removing images.

jonasfj: Yeah, looking [examples](https://docs.docker.com/engine/reference/commandline/rmi/#examples), it seems docker does the right thing.

engines/docker/config.go
```diff
@@ -0,0 +1,22 @@
...
+
+type configType struct {
+	DockerSocket string `json:"dockerSocket"`
+}
+
+var configSchema = schematypes.Object{
+	Properties: schematypes.Properties{
+		"dockerSocket": schematypes.String{
+			Title: "Docker Socket",
+			Description: util.Markdown(`Path to the docker socket hosting the remote docker API.
```
jonasfj: nit: insert line break before `Path...` the `util.Markdown` command will strip empty lines and remove consistent indentation... The first line here is not indented like the second line... nit.. btw, if you have better ideas for a macro we could use to write markdown I'm all ears (but maybe we tweak that in different PR)

engines/docker/engine.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/taskcluster/taskcluster-worker/engines"
+	"github.com/taskcluster/taskcluster-worker/runtime"
+	"github.com/taskcluster/taskcluster-worker/runtime/caching"
+)
+
+type engine struct {
+	engines.EngineBase
+	m           sync.Mutex
+	Environment *runtime.Environment
+	client      *docker.Client
```
jonasfj: nit, but could we rename this to `docker  *docker.Client` ? it would look much better... `engine.client` could be anything.

engines/docker/engine.go
```diff
@@ -0,0 +1,97 @@
...
+	debug("docker engineProvider.NewEngine()")
+	var c configType
+	schematypes.MustValidateAndMap(configSchema, options.Config, &c)
+	if c.DockerSocket == "" {
+		c.DockerSocket = "unix:///var/run/docker.sock"
+	}
+
+	client, err := docker.NewClient(c.DockerSocket)
+	if err != nil {
+		return nil, err
```
jonasfj: `return nil, errors.Wrapf(err, "failed to connect to dockerSocket: '%s'", c.DockerSocket)`

The [`"github.com/pkg/errors"`](https://godoc.org/github.com/pkg/errors) package can be used as drop-in replacement for the built-in `"errors"` package. And it has `errors.Wrap(err, message)` and `errors.Wrapf(err, fmt, arg...)`.

If ever passing up an unhandled error like this, wrapping it helps a lot... it makes it's very clear what when wrong.
Sometimes libraries just return a generic OS error.. like "file not found", which is hard to debug... If we wrap it with `failed to connect to dockerSocket: 'unix:///var/run/docker.sock'` it becomes very obvious, that we failed to connect to docker-socket because the file was missing :)

Note: we could also use `monitor.ReportError` and then return `runtime.ErrFatalInternalError`, but returning a wrapped error is also perfectly fine in this case... As the caller will take care to note that the error happened in `NewEngine` before logging it and reporting it. Also note that if you just return `runtime.ErrFatalInternalError` then caller will assume you've already reported it to sentry and printed it to error log with `monitor.Error(...)`

engines/docker/sandboxbuilder.go
```diff
@@ -0,0 +1,120 @@
...
+		image:     payload.Image,
+		monitor:   monitor,
+		e:         e,
+		taskCtx:   ctx,
+		env:       &docker.Env{},
+		imageDone: make(chan struct{}, 1),
+	}
+	// set image
+	sb.image.engine = sb.e
+	pctx, cancelPull := context.WithCancel(context.Background())
```
jonasfj: We don't need this... Use the `ctx` parameter... it'll be cancelled before `Discard()` is called.

engines/docker/sandbox.go
```diff
@@ -0,0 +1,156 @@
...
+	monitor     runtime.Monitor
+	containerID string
+	resultSet   engines.ResultSet
+	resultErr   error
+	abortErr    error
+	tempStorage runtime.TemporaryStorage
+	resolve     atomics.Once
+	client      *docker.Client
+	taskCtx     *runtime.TaskContext
+	handle      *caching.Handle
```
jonasfj: let's call it `imageHandle`

engines/docker/sandbox.go
```diff
@@ -0,0 +1,156 @@
...
+type sandbox struct {
+	engines.SandboxBase
+	monitor     runtime.Monitor
+	containerID string
+	resultSet   engines.ResultSet
+	resultErr   error
+	abortErr    error
+	tempStorage runtime.TemporaryStorage
+	resolve     atomics.Once
+	client      *docker.Client
```
jonasfj: Instead of holding a reference to `client` let's just add `engine` as a property... both here and in `sandboxBuilder`.

That way you'll also have access to `environment` if we ever need it.. And similarly, if we add new properties in the future.

engines/docker/sandbox.go
```diff
@@ -0,0 +1,156 @@
...
+	return s.resultSet, s.resultErr
+}
+
+func (s *sandbox) wait() {
+	exitCode, err := s.client.WaitContainer(s.containerID)
+	s.resolve.Do(func() {
+		s.resultSet = newResultSet(exitCode == 0, s.containerID, s.client,
+			s.tempStorage, s.handle, s.monitor.WithPrefix("result-set"))
+		if err != nil {
+			s.resultSet = newResultSet(false, s.containerID, s.client, s.tempStorage,
```
jonasfj: Maybe don't create `resultSet` twice if `err != nil`

I suggest that you first determine if it's successful or not and then call `newResultSet`, like:
```go
success := exitCode == 0 && err == nil
s.resultSet = newResultSet(
  success,
  ...
)
```

use multiple lines if the call has a lot of parameters... Also give it `engine` instead of `s.tempStorage`, as we can access `environment.TemporaryStorage` from `engine`...

engines/docker/sandbox.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		s.abortErr = engines.ErrSandboxTerminated
+	})
+}
+
+func (s *sandbox) Kill() error {
+	s.resolve.Do(func() {
+		debug("Sandbox.Kill()")
+		// maybe will use to timeout kill and send SIGKILL
+		killCtx, cancelFunc := context.WithCancel(context.Background())
```
jonasfj: See: https://golang.org/pkg/context/#WithCancel

>  Canceling this context releases resources associated with it, so code should call cancel as soon as the operations running in this Context complete. 

Probably use [`WithTimeout`](https://golang.org/pkg/context/#WithTimeout), like:
```go
ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
defer cancel() 
```

----
Even better something like:

```go
ctx, cancel := context.WithTimeout(context.Background(), dockerEngineKillTimeout)
defer cancel()
_ = s.client.KillContainer(docker.KillContainerOptions{
  ID:      s.containerID,
  Signal:  docker.SIGTERM,
  Context: ctx,
})

err := s.client.KillContainer(docker.KillContainerOptions{
  ID:      s.containerID,
  Signal:  docker.SIGKILL,
})

if (err is NOT because container as already stopped) {
  monitor.reportError(err, 'KillContainer with SIGKILL failed')
}
```

Instead of checking whether the first call to `KillContainer` with `SIGTERM` was successful, we always call a second time with `SIGKILL`. If killing a stopped container causes an error we ignore that error, but otherwise we report any error that would happen in `SIGKILL`.

It's better to handle errors than to try avoid them happening... as you could have a race condition. Note: I don't know if killing a stopped container causes an error.

jonasfj:
@walac, if we land the two PRs I filed into this branch... then I suggest we land this and fix bugs later, rather than aim for a perfect review.

engines/docker/engine.go
```diff
@@ -0,0 +1,97 @@
...
+	},
+}
+
+func (e *engine) PayloadSchema() schematypes.Object {
+	return payloadSchema
+}
+
+func (e *engine) NewSandboxBuilder(options engines.SandboxOptions) (engines.SandboxBuilder, error) {
+	var p payloadType
+	schematypes.MustValidateAndMap(payloadSchema, options.Payload, &p)
```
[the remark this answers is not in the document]
jonasfj: MustXXX will panic... the error case violates an invariant.

engines/docker/engine.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func (p engineProvider) ConfigSchema() schematypes.Schema {
+	return configSchema
+}
+
+func (p engineProvider) NewEngine(options engines.EngineOptions) (engines.Engine, error) {
+	debug("docker engineProvider.NewEngine()")
+	var c configType
+	schematypes.MustValidateAndMap(configSchema, options.Config, &c)
```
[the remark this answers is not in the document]
jonasfj: Panic... tc-worker requires input validation is done before passing it in... this just an assertion..

engines/docker/resultset.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		if hdr.Typeflag == tar.TypeDir {
+			continue
+		}
+		fname := strings.TrimPrefix(hdr.Name, strip)
+		m := monitor.WithTag("filename", fname)
+
+		tempfile, err := r.tempStorage.NewFile()
+		if err != nil {
+			m.ReportError(err, "could not create temporary file")
```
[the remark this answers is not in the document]
jonasfj: agreed

engines/docker/image_cache.go
```diff
@@ -0,0 +1,108 @@
...
+	monitor := options.engine.monitor.WithPrefix("image-cache").WithTag("image", buildImageName(options.Repository, options.Tag))
+
+	// TODO: Use outputstream to write progress
+	debug("pulling image %s %s", options.Repository, options.Tag)
+	err := client.PullImage(docker.PullImageOptions{
+		Repository:        options.Repository,
+		Tag:               options.Tag,
+		InactivityTimeout: dockerPullImageInactivityTimeout,
+		Context:           ctx,
+	}, docker.AuthConfiguration{})
```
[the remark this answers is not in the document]
jonasfj: Won't docker do this?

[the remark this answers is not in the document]
walac: I don't know

jonasfj:
Moved to #343

PR #332, opened 2018-03-04
fix CI + update of stream package

jonasfj:
This should fix two intermittent bugs :) and get CI passing... + update a dependency (I suspect stream caused some deadlocks), so that's nice

### DIFF
--- a/worker/workertest/superseded_test.go
+++ b/worker/workertest/superseded_test.go
@@ -4,19 +4,51 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"sync"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/taskcluster/slugid-go/slugid"
+	"github.com/taskcluster/taskcluster-client-go/queue"
 	"github.com/taskcluster/taskcluster-worker/runtime"
 )
 
 func TestSuperseded(t *testing.T) {
+	// This test aims to check if the superseded functionality works. As we can't
+	// know which order the tasks will be claimed in... This a little complicated
+	// to test. We do it by using the task index to recall which one of them set
+	// to be superseded. When the HTTP server gets called as supersederUrl and
+	// we haven't returned a superseded task yet, we take the next available taskId
+	// as superseded the current taskId. We store the index of the superseded task
+	// as superseded.
+	superseded := -1 // index of task that was superseded
 	var c Case
+	var m sync.Mutex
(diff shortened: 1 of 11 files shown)


jonasfj:
The OSX CI job not running is because pmoore is offline :)

jonasfj:
I can't reproduce the current error on macstadium... but we've certainly fixed a few bugs with this, so I suggest we land it asap.

worker/workertest/superseded_test.go
```diff
@@ -4,19 +4,51 @@ import (
...
 	var c Case
+	var m sync.Mutex
 	s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		m.Lock() // lock access to superseded
+		defer m.Unlock()
+
+		// Get taskId from which this is being called
 		taskID := r.URL.Query().Get("taskId")
+
+		// Create array of superseding taskIds, always start with it's own taskId
```
walac: nit: "its own"

worker/workertest/superseded_test.go
```diff
@@ -4,19 +4,51 @@ import (
...
 		supersedes := []string{taskID}
-		if c.Tasks[0].TaskID == taskID {
-			supersedes = append(supersedes, c.Tasks[1].TaskID)
+
+		// If we haven't made a superseded task yet we do that now
+		if superseded == -1 {
+			// Find index of this taskId
+			for index, task := range c.Tasks {
+				if task.TaskID == taskID {
+					superseded = index
```
walac: shouldn't we add a `break` statement here?

worker/workertest/superseded_test.go
```diff
@@ -28,6 +60,24 @@ func TestSuperseded(t *testing.T) {
...
+	verifyTask := func(i int) StatusAssertion {
+		return func(t *testing.T, q *queue.Queue, status queue.TaskStatusStructure) {
+			// Sanity check for status structure
+			assert.Equal(t, c.Tasks[i].TaskID, status.TaskID, "Expected taskId to match")
+			if i == superseded {
+				// If this is the task that was superseded, we assume it was resolved as such
+				runID := len(status.Runs) - 1
+				assert.Equal(t, status.Runs[runID].ReasonResolved, runtime.ReasonSuperseded.String(), "expected superseded")
+			} else {
+				// If this task was not superseded is should be successful
```
walac: nit: "it should"

worker/workertest/superseded_test.go
jonasfj: No harm either way.

PR #335, opened 2018-03-07
Fix docker kill

jonasfj:


engines/docker/engine_test.go
```diff
@@ -4,8 +4,15 @@ import (
...
-	"github.com/taskcluster/taskcluster-worker/engines/enginetest"
 	"time"
+
+	"github.com/taskcluster/taskcluster-worker/engines/enginetest"
+)
+
+// Image and tag used in test cases below
+const (
+	dockerImageRepository = "alpine"
+	dockerImageTag        = "3.6"
```
jonasfj: The image hash disappeared from the repository without warning... So I guess maybe we have to use tags...
Or well, at-least we might have to support both.

engines/docker/sandbox.go
```diff
@@ -98,59 +98,107 @@ func (s *sandbox) wait() {
...
+
+	// Send SIGTERM and give the container 30s to exit
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel() // always free the context
+	err := s.client.KillContainer(docker.KillContainerOptions{
+		ID:      s.containerID,
+		Signal:  docker.SIGTERM,
+		Context: ctx,
+	})
+	// Report errors if ContainerNotRunning or ctx was timed out
```
walac: This comment conflicts with the function description.

engines/docker/sandbox.go
```diff
@@ -98,59 +98,107 @@ func (s *sandbox) wait() {
...
+	if _, ok := err.(*docker.ContainerNotRunning); err != nil && !ok && ctx.Err() == nil {
+		s.monitor.ReportError(err, "KillContainer with SIGTERM failed")
+		// signal up the stack that something went wrong, this is not a successful kill
+		hasErr = true
```
walac: I think it is better if we add a name to the return value and set it here.

engines/docker/sandbox.go
```diff
@@ -98,59 +98,107 @@ func (s *sandbox) wait() {
...
+	// Free image handle
+	s.handle.Release()
+
+	// Remove temporary storage
+	if err = s.tempStorage.Remove(); err != nil {
+		s.monitor.ReportError(err, "failed to remove temporary storage in disposal of sandbox")
+	}
+
+	// If ErrNonFatalInternalError if there was an error of any kind
+	if hasErr {
```
walac: `hasErr` is never set to `true`

PR #336, opened 2018-03-08
Added per-container network to docker-engine

jonasfj:
This is written on top of `fix-docker-kill`, but otherwise it should be merged into the `docker-engine` branch..

There is a few renames and cleanups, etc...

Notable fix is that containers can't talk to each other anymore. And we're prepared for adding proxy support, which I'll play with later...

### DIFF
--- a/engines/docker/resultset.go
+++ b/engines/docker/resultset.go
@@ -18,22 +18,11 @@ type resultSet struct {
 	engines.ResultSetBase
 	success     bool
 	containerID string
+	networkID   string
 	client      *docker.Client
 	monitor     runtime.Monitor
-	tempStorage runtime.TemporaryFolder
-	handle      *caching.Handle
-}
-
-func newResultSet(success bool, containerID string, client *docker.Client,
-	ts runtime.TemporaryFolder, handle *caching.Handle, monitor runtime.Monitor) *resultSet {
-	return &resultSet{
-		success:     success,
-		containerID: containerID,
-		client:      client,
-		tempStorage: ts,
-		handle:      handle,
-		monitor:     monitor,
-	}
+	storage     runtime.TemporaryFolder
+	imageHandle *caching.Handle
 }
 
 func (r *resultSet) Success() bool {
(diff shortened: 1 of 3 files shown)


engines/docker/resultset.go
```diff
@@ -155,37 +144,54 @@ func (r *resultSet) ExtractFolder(path string, handler engines.FileHandler) erro
...
-		err := r.tempStorage.Remove()
-		if err != nil {
-			r.monitor.ReportWarning(err, "could not remove temporary storage")
-		}
-	}
-	if r.handle != nil {
-		r.handle.Release()
-	}
-	return r.client.RemoveContainer(docker.RemoveContainerOptions{
+	hasErr := false
```
walac: Better to add a name to the return value and set it where it is convenient.

engines/docker/sandbox.go
```diff
@@ -18,48 +21,94 @@ type sandbox struct {
...
+	}
+	networkID := network.ID
+
+	// Inspect network to get the IP assigned
+	network, err = sb.e.client.NetworkInfo(networkID)
+	if err != nil {
+		// Any error here is a fatal error
+		return nil, errors.Wrap(err, "docker.CreateNetwork failed")
 	}
+	// The gateway is the IP of the host machine, we to insert that has hostname "taskcluster"
```
walac: nit: *we to insert* ???

engines/docker/sandbox.go
```diff
@@ -18,48 +21,94 @@ type sandbox struct {
...
+		Internal: false,
+		Labels: map[string]string{
+			"taskId": sb.taskCtx.TaskID,
+		},
+	})
+	if err != nil {
+		// Any error here is a fatal error
+		return nil, errors.Wrap(err, "docker.CreateNetwork failed")
+	}
+	networkID := network.ID
```
walac: this code seems redundant

walac:
Err, I don't get why GitHub marks my review as pending, as I already reviewed it...

jonasfj:
> Err, I don't get why GitHub marks my review as pending, as I already reviewed it...

No idea, sometimes I forget to click "submit review" :)

jonasfj:
Merged into #343

PR #337, opened 2018-03-14
Fix client ID validation pattern

walac:
Client IDs with LDAP portions in its composition may include the '|'
character.


jonasfj:
Nice catch... thanks!

PR #339, opened 2018-03-14
Update dependencies.

walac:
Update taskcluster-client-go and taskcluster-worker/worker packages to
have correct client ID validation.

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -366,10 +366,10 @@
 			"revisionTime": "2017-02-22T01:37:00Z"
 		},
 		{
-			"checksumSHA1": "dK3vEsrwRHwQbmlEmIIGzlZm29I=",
+			"checksumSHA1": "lbsOyFAjyMWGUZ3ojfwDX9+vkU0=",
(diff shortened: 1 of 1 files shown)


vendor/vendor.json
```diff
@@ -401,6 +401,12 @@
 			"revision": "0adb988569ffd40f68ed3422835325cfb03581c5",
 			"revisionTime": "2017-07-06T18:25:07Z"
 		},
+		{
+			"checksumSHA1": "Fas2k5UYHwTrZWoERdA4vCo6lms=",
+			"path": "github.com/taskcluster/taskcluster-worker/worker",
```
jonasfj: I don't think this should sneak into dependencies...

@walac, probably you'll want to work out of a folder checked out as: `$GOPATH/src/github.com/taskcluster/taskcluster-worker/` maybe even just use this repo as the upstream repo..

I generally, don't fork repos to my own user if I have commit access to them.. for golang it gets really dirty if you check things out under a different repo name...

walac:
@jonasfj Fixed

PR #343, opened 2018-03-29
Docker engine dind

jonasfj:
@ckousik, @walac, this is all the PRs to do with docker-engine merged with `master`.

I fixed most of things we found in reviews.. I'm sure I didn't get them all, and there sure is more things to fix.

I'm open to landing it... and filing issues for the remaining things... We have other bugs in tc-worker that needs fixing and those have cross cutting concerns, so long live branches are bound to create painful merge conflicts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,8 @@ why these tests are disabled by default.
    (run tests with `./docker-tests.sh`).
  * `monitor`, tests sentry reporting, statsum submission and logging, requires
    credentials to run successfully.
-
+ * `docker`, tests docker-engine, disabled because tests requires docker
+   installed and must run as root (run tests with `./docker-tests.sh`).
 
 Releasing
 ---------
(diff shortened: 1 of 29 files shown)


README.md
```diff
@@ -65,7 +65,8 @@ why these tests are disabled by default.
    (run tests with `./docker-tests.sh`).
  * `monitor`, tests sentry reporting, statsum submission and logging, requires
    credentials to run successfully.
-
+ * `docker`, tests docker-engine, disabled because tests requires docker
+   installed and must run as root (run tests with `./docker-tests.sh`).
```
walac: Isn't the current user being in the `docker` user group enough?

jonasfj: nope, we need root because we listen on port 80 of the gateway IPs that docker creates for the isolated networks, we need this to be able to support attaching proxies from plugins.

Issue #334, opened 2018-03-04
caching.New(...) should take a Monitor and count cache hits


jonasfj:
`caching.New(...)` should additionally accept a `runtime.Monitor` instance...

So that setup looks somewhat like:
```go
cache := caching.New(constructor, true || false, garbageCollector, monitor.prefix('<prefix>'))
```

Then the `cache` should track cache hit/miss ratio, as well as constructor duration... as metrics: `<prefix>.cache-hit` and `<prefix>.cache-miss` and `<prefix>.creation-time`

That way if we pass `monitor.prefix('image-manager')` to `caching.New(...)` we get metrics: 
`image-manager.cache-hit` and `image-manager.cache-miss` and `image-manager.creation-time` which would be pretty neat :)

Biboswan:
is this issue is available to be taken ?

jonasfj:
sure, go for it... note that we're going to merge PR #343 soon which adds new usages of `caching` :)

jonasfj:
Fixed in #344

PR #344, opened 2018-04-03
Monitor cache

Biboswan:
#334 
If I got it right I shall modify cache_test as well

### DIFF
--- a/plugins/cache/cache.go
+++ b/plugins/cache/cache.go
@@ -79,8 +79,8 @@ func (p *provider) NewPlugin(options plugins.PluginOptions) (plugins.Plugin, err
 		engine:         options.Engine,
 		environment:    options.Environment,
 		monitor:        options.Monitor,
-		sharedCache:    caching.New(constructor, false, options.Environment.GarbageCollector),
-		exclusiveCache: caching.New(constructor, false, options.Environment.GarbageCollector),
+		sharedCache:    caching.New(constructor, false, options.Environment.GarbageCollector, options.Monitor),
+		exclusiveCache: caching.New(constructor, false, options.Environment.GarbageCollector, options.Monitor),
 		lastPurged:     time.Now(),
 		config:         c,
 	}, nil
--- a/runtime/caching/cache.go
+++ b/runtime/caching/cache.go
@@ -20,12 +20,14 @@ type Constructor func(ctx Context, options interface{}) (Resource, error)
 // the option of being an exclusive or shared-cache, meaning if resources may be
 // re-used before they are released.
 type Cache struct {
-	m           sync.Mutex
-	shared      bool
-	entries     []*cacheEntry
-	constructor Constructor
-	tracker     gc.ResourceTracker
-	cachehit    runtime.Monitor
+	m               sync.Mutex
+	shared          bool
+	entries         []*cacheEntry
+	constructor     Constructor
+	tracker         gc.ResourceTracker
+	cachehit        runtime.Monitor
+	constructortime runtime.Monitor
+	countCachehits  float64
 }
 
 // New returns a Cache wrapping constructor such that resources
@@ -33,10 +35,12 @@ type Cache struct {
 // options, if shared is set to true. Otherwise, resources are exclusive.
 func New(constructor Constructor, shared bool, tracker gc.ResourceTracker, monitor runtime.Monitor) *Cache {
 	return &Cache{
-		constructor: constructor,
-		tracker:     tracker,
-		shared:      shared,
-		cachehit:    monitor.WithPrefix("cache-hit"),
+		constructor:     constructor,
+		tracker:         tracker,
+		shared:          shared,
+		cachehit:        monitor.WithPrefix("cache-hit"),
+		constructortime: monitor.WithPrefix("creation-time"),
+		countCachehits:  0,
 	}
 }
 
@@ -93,7 +97,8 @@ func (c *Cache) Require(ctx Context, options interface{}) (*Handle, error) {
 
 		// Take the entry
 		debug("cache entry '%s' found in cache, refCount: %d", optionsHash, e.refCount+1)
-		c.monitor.prefix["cache-hit"].Count("counter-1", 1)
+		c.countCachehits++
+		c.cachehit.Count("count cache-hit", c.countCachehits)
 		entry = e
 		e.refCount++
 		e.m.Unlock()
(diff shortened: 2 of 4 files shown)


walac:
From Travis CI
```
plugins/cache/cache.go:82:30: not enough arguments in call to caching.New
	have (func(caching.Context, interface {}) (caching.Resource, error), bool, gc.ResourceTracker)
	want (caching.Constructor, bool, gc.ResourceTracker, runtime.Monitor)
plugins/cache/cache.go:83:30: not enough arguments in call to caching.New
	have (func(caching.Context, interface {}) (caching.Resource, error), bool, gc.ResourceTracker)
	want (caching.Constructor, bool, gc.ResourceTracker, runtime.Monitor)
make: *** [install] Error 2
```

runtime/caching/cache.go
```diff
@@ -19,21 +20,27 @@ type Constructor func(ctx Context, options interface{}) (Resource, error)
...
-	shared      bool
-	entries     []*cacheEntry
-	constructor Constructor
-	tracker     gc.ResourceTracker
+	m               sync.Mutex
+	shared          bool
+	entries         []*cacheEntry
+	constructor     Constructor
+	tracker         gc.ResourceTracker
+	cachehit        runtime.Monitor
```
jonasfj: a single `monitor` property should do...

runtime/caching/cache.go
```diff
@@ -19,21 +20,27 @@ type Constructor func(ctx Context, options interface{}) (Resource, error)
...
-func New(constructor Constructor, shared bool, tracker gc.ResourceTracker) *Cache {
+func New(constructor Constructor, shared bool, tracker gc.ResourceTracker, monitor runtime.Monitor) *Cache {
 	return &Cache{
-		constructor: constructor,
-		tracker:     tracker,
-		shared:      shared,
+		constructor:     constructor,
+		tracker:         tracker,
+		shared:          shared,
+		cachehit:        monitor.WithPrefix("cache-hit"),
```
jonasfj: Store the monitor and do `WithPrefix` when reporting... in fact you won't need it as `.Count(` takes a name

runtime/caching/cache.go
```diff
@@ -90,6 +97,8 @@ func (c *Cache) Require(ctx Context, options interface{}) (*Handle, error) {
 
 		// Take the entry
 		debug("cache entry '%s' found in cache, refCount: %d", optionsHash, e.refCount+1)
+		c.countCachehits++
+		c.cachehit.Count("count cache-hit", c.countCachehits)
```
jonasfj: Use the `monitor` given in `New` and then call `monitor.Count('cache-hit', 1)`

If there is a cache miss, the call `monitor.Count('cache-miss', 1)`

the monitor will aggregate metrics that have the same name. Take a look at statsum if you're curious.

runtime/caching/cache.go
```diff
@@ -111,9 +120,14 @@ func (c *Cache) Require(ctx Context, options interface{}) (*Handle, error) {
 		c.entries = append(c.entries, entry)
 
 		go entry.created.Do(func() {
-			defer entry.ctx.dispose() // ensure resources are cleanup when constructor is done
+			defer entry.ctx.dispose()
+			// ensure resources are cleanup when constructor is done
+
+			c.constructortime.Time("constructor-creation-time", func() {
```
jonasfj: Use same `monitor` as given in `New`, then call `monitor.Time('constructor-duration', func() {...`

jonasfj:
No, I don't think they are related...

Biboswan:
then any changes to be done?

plugins/cache/cache.go
```diff
@@ -79,8 +79,8 @@ func (p *provider) NewPlugin(options plugins.PluginOptions) (plugins.Plugin, err
 		engine:         options.Engine,
 		environment:    options.Environment,
 		monitor:        options.Monitor,
-		sharedCache:    caching.New(constructor, false, options.Environment.GarbageCollector),
-		exclusiveCache: caching.New(constructor, false, options.Environment.GarbageCollector),
+		sharedCache:    caching.New(constructor, false, options.Environment.GarbageCollector, options.Monitor),
```
jonasfj: `.WithPrefix('shared-cache')` and simlar for exclusive...

What else are the metric names going to be?

runtime/caching/cache.go
```diff
@@ -111,9 +116,14 @@ func (c *Cache) Require(ctx Context, options interface{}) (*Handle, error) {
 		c.entries = append(c.entries, entry)
 
 		go entry.created.Do(func() {
-			defer entry.ctx.dispose() // ensure resources are cleanup when constructor is done
+			defer entry.ctx.dispose()
+			// ensure resources are cleanup when constructor is done
```
jonasfj: This comment pertains to the line of the code that it follows? that seems unwise.

Also some grammar (which was probably there before), but `are cleaned up`

runtime/caching/cache.go
```diff
@@ -111,9 +116,14 @@ func (c *Cache) Require(ctx Context, options interface{}) (*Handle, error) {
...
+
+			c.monitor.Time("constructor-duration", func() {
+				entry.resource, entry.err = c.constructor(entry.ctx, options)
+			})
+			//metric to give creation time of constructor
```
jonasfj: formatting... look how other comments have a space before them...

And maybe rephrase it...

plugins/cache/cache.go
jonasfj: so I guess the questions is if we want them aggregated together or not...

We could get names:
```
shared-cache.cache-hit
shared-cache.cache-miss
shared-cache.constructor-duration
exclusive-cache.cache-hit
exclusive-cache.cache-miss
exclusive-cache.constructor-duration
```
or we could get:
```
cache-hit
cache-miss
constructor-duration
```
if we keep it... my question is if other place there are metrics recorded to `c.monitor` without any prefixes... I don't actually mind aggregating them, that seems decent (I think).

Biboswan: Can there be instances where you would like to see what should be shared and what should be exclusive for better performance. Otherwise better to keep them aggregated to reduce noise..., you will definitely have better idea.

jonasfj: I think reducing noise is useful... depending on how we use caches these stats could easily be somewhat irrelevant anyways.

Yes, it's cool to have them... for other caches they'll make a lot more sense..

Biboswan:
fine its done then..

Biboswan:
done

jonasfj:
Thanks, this is awesome...

PR #346, opened 2018-04-06
Fix tc client

jonasfj:


worker/worker.go
```diff
@@ -402,7 +402,8 @@ func (w *Worker) processClaim(claim taskClaim) {
 			debug("queue.reclaimTask(%s, %d)", claim.Status.TaskID, claim.RunID)
 			result, err := q.ReclaimTask(claim.Status.TaskID, runID)
 			if err != nil {
-				if e, ok := err.(httpbackoff.BadHttpResponseCode); ok && e.HttpResponseCode == 409 {
+				if e, ok := err.(*tcclient.APICallException); ok && e.CallSummary.HTTPResponse.StatusCode == 409 {
```
jonasfj: @petemoore, the problem with breaking tc-client-go is that fixing bugs like this is super hard :)
Luckily I implemented a fake queue implementation that wraps a mock queue, so that I got an error when it got unexpected `ReportCompleted` because it failed to interpret the `409`.

I suspect that we could refactor tc-client-go to return something simple like what js does:
```go
type Error struct {
  StatusCode int `json:"-"`
  Code string `json:"code"`
  Message string `json:"message"`
  Details map[string]interface{} `json:"details"`
}
```

And then we would state that tc-client-go returns these kinds of errors:
 * `Error{...}` for API errors where we get a non-2xx response,
 * `context.Err()` if context was canceled...
 * `errors.Wrap(err, "...explanation of what we tried to do, number of attempts, etc... as human read-able message")` where `err` is passed through from the OS for all underlying OS errors... like network unavailable, etc... (these errors can be cast to `errors.Cause` and investigated if you really want to).

For debugging it's better to have optional logging we can enable...

PR #347, opened 2018-04-06
Fix documentation comment

jonasfj:
Just a quick fix of some ugly docs I noticed..
Using spaces instead of tabs doesn't work with `util.Markdown`


jonasfj:
Feel bad asking for a few, but now you get hit r+, feel free to hit merge too :)

PR #349, opened 2018-04-06
Add some more docs for script-engine

jonasfj:
I suspect none of this is a surprise to @bclary, but still it's nice to document things.
This should add error codes to the covered docs.

### DIFF
--- a/engines/script/config.go
+++ b/engines/script/config.go
@@ -32,7 +32,7 @@ var configSchema = schematypes.Object{
 				_working directory_, this folder can be used for temporary storage and
 				will be cleared between tasks.
 
-				Files and folder stored in './artifacts/' relative to the
+				Files and folders stored in './artifacts/' relative to the
 				_working directory_ will be uploaded as artifacts from
 				the script. Hence, to make a public tar-ball artifact you create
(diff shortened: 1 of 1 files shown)


engines/script/config.go
```diff
@@ -27,13 +27,24 @@ var configSchema = schematypes.Object{
 				Output from the script over 'stdout' will be uploaded as task log.
 				Output from the script over 'stderr' will be prefixed "[worker:error]"
 				and merged with task log.
+
 				The script will be executed with a temporary folder as
 				_working directory_, this folder can be used for temporary storage and
-				will be cleared between tasks. Files and folder stored in './artifacts/'
```
bclary: s/Files and folder/Files and folders/

engines/script/config.go
```diff
@@ -27,13 +27,24 @@ var configSchema = schematypes.Object{
...
-				relative to the _working directory_ will be uploaded as artifacts from
+				will be cleared between tasks.
+
+				Files and folder stored in './artifacts/' relative to the
```
bclary: s/Files and folder/Files and folders/

Issue #351, opened 2018-04-07
Add staticlog plugin 


jonasfj:
It could be pretty neat to have a plugin that only does static logs. Creating an artifact `task.log`.

Mostly, I think it's useful to enable this, if you disable livelogs... Perhaps you don't want outgoing network connections to webhooktunnel. Perhaps in some scenarios, this would be hardening.

Certainly, it would help debug cases where we suspect livelog is guilty of crashes/hangs...

jonasfj:
#fixed in 351

PR #352, opened 2018-04-08
Add tasklog plugin fixing #351

jonasfj:
This adds a simple `tasklog` plugin, that just does static logs.
Maybe I should have called it `staticlog`, but `tasklog` also works.
There is no reason to run this next to `livelog`, it's a strict subset
of the features offered by the `livelog`.

**Motivation**, ability to better lock down security critical workers,
that you don't want to handle requests, even if they are proxied through
`webhooktunnel`. Other motivation is that I suspect some of the bugs
@bclary is hitting could be related to `livelog` or `webhooktunnel`,
if using this plugin instead of `livelog` fixes his problems, then
we know for sure we have a bug in `livelog`. Right now, bugs in `livelog`
is pure speculation.

### DIFF
--- a/plugins/tasklog/tasklog.go
+++ b/plugins/tasklog/tasklog.go
@@ -94,7 +94,7 @@ func (tp *taskPlugin) uploadLog() error {
 
 	// Upload gzipped task.log
 	debug("uploading 'public/logs/task.log'")
-	tp.context.UploadS3Artifact(runtime.S3Artifact{
+	err = tp.context.UploadS3Artifact(runtime.S3Artifact{
 		Name:     "public/logs/task.log",
 		Mimetype: "text/plain; charset=utf-8",
 		Expires:  tp.context.TaskInfo.Expires,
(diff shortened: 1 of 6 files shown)


jonasfj:
This aims to fix #351

plugins/tasklog/tasklog.go
```diff
@@ -0,0 +1,113 @@
...
+	tp.context.UploadS3Artifact(runtime.S3Artifact{
+		Name:     "public/logs/task.log",
+		Mimetype: "text/plain; charset=utf-8",
+		Expires:  tp.context.TaskInfo.Expires,
+		Stream:   tempFile,
+		AdditionalHeaders: map[string]string{
+			"Content-Encoding": "gzip",
+		},
+	})
+	if err != nil {
```
walac: Uh, forgot to assign `err` in the call to `UploadS3Artifact`?

jonasfj: Nice catch!

PR #355, opened 2018-04-11
Fix vendoring in CI

jonasfj:
Basically, a revision disappeared from tc-client-go, I think a branch was
deleted when we landed stuff... Or maybe there is some other bad reason.
Anyways, I hope this fixes the situation...


jonasfj:
I'm working in `hack-ci-to-work` to find a fix for the windows script engine integration test issue...

But this should probably just land asap... If test pass I might just land it tomorrow, as vendoring is boring to review anyways... and this fixes CI.

jonasfj:
This strictly improved things... just fixing the vendor issues is bring to review anyways...

PR #360, opened 2018-04-16
WIP: Refactoring artifact extraction from docker-engine

jonasfj:
Just so @walac can keep an eye on what I'm hacking on :)


engines/docker/resultset.go
```diff
@@ -31,115 +41,189 @@ func (r *resultSet) Success() bool {
...
+
+			// Seek start of temp file
+			if _, rerr = tmpfile.Seek(0, io.SeekStart); rerr != nil {
+				tmpfile.Close()
+				rerr = errors.Wrap(rerr, "failed to seek to start of temporary file, after reading from docker")
+				return
+			}
+
+			// Invoke handler concurrently
+			wg.WaitForLessThan(5) // No more than 5 concurrent calls to handler
```
walac: I see this is pointless, the Go scheduler already leverages concurrency pretty well.

engines/docker/resultset.go
```diff
@@ -31,115 +41,189 @@ func (r *resultSet) Success() bool {
...
 
-		if _, err = io.Copy(tempfile, reader); err != nil {
-			m.ReportError(err, "could not copy file")
-			continue
-		}
+	ctx, cancel := context.WithCancel(r.context)
+	notFound := false
+	var derr, rerr error
+	var interrupted atomics.Once
+	util.Parallel(func() {
```
walac: Is this parallel to what?

engines/docker/resultset.go
```diff
@@ -31,115 +41,189 @@ func (r *resultSet) Success() bool {
...
+		defer cancel()
```
walac: Why do we need to unconditionally call `cancel`?

engines/docker/resultset.go
```diff
@@ -31,115 +41,189 @@ func (r *resultSet) Success() bool {
...
+		debug("ExtractFolder(%s) extracting a folder", path)
+		derr = r.docker.DownloadFromContainer(r.containerID, docker.DownloadFromContainerOptions{
+			Context:           ctx,
+			OutputStream:      streamWriter,
+			Path:              path,
+			InactivityTimeout: 5 * time.Second,
+		})
+		if derr != nil && derr == ctx.Err() {
+			derr = nil // Ignore any error
+			streamWriter.CloseWithError(errors.Wrap(ctx.Err(), "download form docker aborted"))
```
walac: nit: "Download from docker..."

engines/resultset.go
```diff
@@ -8,8 +8,8 @@ import (
 //
 // ResultSet.ExtractFolder(path, handler) takes a FileHandler as the handler
 // parameter. This function may be called sequentially or concurrently, but if
-// it returns an the ResultSet should stop calling it and pass the error through
-// as return value from ResultSet.ExtractFolder.
+// it returns an error the ResultSet should stop calling it and return
```
walac: nit: "... an error, the ResultSet ..."

engines/docker/resultset.go
jonasfj: Just call two functions in parallel and wait for both of them to return.

engines/docker/resultset.go
jonasfj: yeah, but we're creating temporary files... So maybe only having 5 of those around at the same time is a good sanity check..

Maybe it should be higher, but some limit to concurrency is probably a good idea...

engines/docker/resultset.go
```diff
@@ -2,26 +2,36 @@ package dockerengine
...
 	"github.com/taskcluster/taskcluster-worker/runtime"
+	"github.com/taskcluster/taskcluster-worker/runtime/atomics"
 	"github.com/taskcluster/taskcluster-worker/runtime/caching"
 	"github.com/taskcluster/taskcluster-worker/runtime/ioext"
+	"github.com/taskcluster/taskcluster-worker/runtime/util"
 )
 
+// maxExtractedFileSizeAllowed is the maximum file size that will be read out
+// of docker with ExtractFile/ExtraFolder, this is limited for sanity.
+const maxExtractedFileSizeAllowed = 256 // GiB
```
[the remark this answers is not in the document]
jonasfj: Are you sure, my concern is that someone can create a sparse file... and if docker API is smart it'll stream out a sparse file in tar.. and we won't write it as a sparse file, thus crash the host.

engines/docker/resultset.go
```diff
@@ -31,115 +41,189 @@ func (r *resultSet) Success() bool {
...
-	defer func() {
-		err = tarfile.Close()
-		if err != nil {
-			monitor.ReportWarning(err, "could not close temporary tar file")
+	var result ioext.ReadSeekCloser
+	var m sync.Mutex
+	err := r.extractResource(path, func(p string, stream ioext.ReadSeekCloser) error {
+		m.Lock()
+		defer m.Unlock()
+		if result != nil {
```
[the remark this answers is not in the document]
jonasfj: This method can be called twice... in theory.. in practice it never should, unless we're pointing at a folder instead of a file.

engines/docker/resultset.go
```diff
@@ -31,115 +41,189 @@ func (r *resultSet) Success() bool {
...
-
-	reader := tar.NewReader(tarfile)
-	_, err = reader.Next()
-	if err != nil {
-		return nil, runtime.ErrNonFatalInternalError
-	}
+		result = stream
+		return nil
+	})
+	m.Lock() // shouldn't be a problem has all handlers have returned, but it makes the race detector happy
```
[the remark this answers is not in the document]
jonasfj: true, I actually had a bug :) which I found...

engines/docker/resultset.go
jonasfj: Yes, contexts should **always** be canceled.

In this case, the code-path that handles reading the tar-stream should cancel the context given when fetching the tar-stream from docker. Because if the reading of the tar-stream fails, we want to abort the `DownloadFromDocker` call.

jonasfj: I mean `DownloadFromContainer` should be aborted. Usually, it'll always be done, before we're doing reading the tar-stream... but just in case there is an error reading the tar-stream :)

engines/docker/resultset.go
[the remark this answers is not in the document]
walac: Where the 256 GB number comes from?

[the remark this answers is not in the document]
walac: Where does the 256 GB number come from?

engines/docker/resultset.go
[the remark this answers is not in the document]
walac: I think it is just more natural use the last one, as this is not clear for the reader.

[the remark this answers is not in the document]
walac: So, let's use the later `result`

engines/docker/resultset.go
walac: We then should ignore it for now, just setting an arbitrary number in concurrency doesn't fix the problem.

engines/docker/resultset.go
walac: > Why do we need to unconditionally call cancel?

Oh, yeah, you are right.

engines/docker/resultset.go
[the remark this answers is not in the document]
jonasfj: It's arbitrary, I just figure that anything above that size I would rather not have to deal with.

It's purely a sanity check.

engines/docker/resultset.go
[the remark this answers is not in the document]
jonasfj: Okay, I actually want to discard both, and return an error... I clarified a bit in comments, and made it more obvious what I'm doing.

engines/docker/resultset.go
jonasfj: Other engines always call the handler sequentially. So I think 5 is good.

But I'll make it a constant so it's easy to identify. We have these sanity limits many places, some we'll never need to tweak others will need tweaks in the future.

PR #365, opened 2018-04-17
Enable tasklog plugin

jonasfj:


jonasfj:
@bclary, I'll tag a release with this tomorrow...

Issue #366, opened 2018-04-17
runtime/fetcher needs auxiliary function for retriable pipeline downloads


jonasfj:
Sometimes we want to use the `Fetcher` interface, but process the downloaded data in a pipeline, instead of a file. Obviously, such a pipeline has to be resetable...

```golang
func processIncoming(ctx context.Context, stream io.Reader) error

var target fetcher.WriteReseter = fetcher.Pipeline(processIncoming)
```

In `fetcher` we would need:
```golang
func Pipeline(makePipeline func(ctx context.Context, stream io.Reader) error) WriteReseter {
  ...
}
```

This could be done as follows: https://gist.github.com/jonasfj/74144fc09d25a4aa2a0da5d58f152eb1

**Use cases**:
 * download a docker image and do the zstd decompression, image rename and docker load on the fly.
 * same thing for qemu-engine, which currently uses an intermediate file..

jonasfj:
Note. We probably still need a way to wait for the pipeline to be completed after fetcher returns.

jonasfj:
This landed in #369

PR #369, opened 2018-04-21
Fixing #366 by adding fetcher.FetchAsStream()

jonasfj:
This will ideally make it easier to use the `fetcher.Fetcher` interface
in cases where you want to process the output with a stream.
So long as the stream can be aborted and started over again, it should be fine.

### DIFF
--- a/runtime/fetcher/fakes_test.go
+++ b/runtime/fetcher/fakes_test.go
@@ -11,36 +11,36 @@ import (
 	"github.com/taskcluster/taskcluster-worker/runtime/client"
 )
 
-type mockContext struct {
+type fakeContext struct {
 	context.Context
 	queue           client.Queue
 	m               sync.Mutex
 	progressReports []float64
 }
 
-func (c *mockContext) Queue() client.Queue {
+func (c *fakeContext) Queue() client.Queue {
 	return c.queue
 }
 
-func (c *mockContext) Progress(description string, percent float64) {
+func (c *fakeContext) Progress(description string, percent float64) {
 	c.m.Lock()
 	defer c.m.Unlock()
 	debug("Progress: %s - %.02f %%", description, percent*100)
 	c.progressReports = append(c.progressReports, percent)
 }
 
-func (c *mockContext) ProgressReports() []float64 {
+func (c *fakeContext) ProgressReports() []float64 {
 	c.m.Lock()
 	defer c.m.Unlock()
 	return c.progressReports
 }
 
-type mockWriteReseter struct {
+type fakeWriteReseter struct {
 	offset int64
 	buffer []byte
 }
 
-func (w *mockWriteReseter) Write(p []byte) (int, error) {
+func (w *fakeWriteReseter) Write(p []byte) (int, error) {
 	offset := w.offset + int64(len(p))
 	if int64(len(w.buffer)) < offset {
 		w.buffer = append(w.buffer, make([]byte, offset-int64(len(w.buffer)))...)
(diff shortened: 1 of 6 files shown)


runtime/fetcher/fakes_test.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (c *fakeContext) ProgressReports() []float64 {
+	c.m.Lock()
+	defer c.m.Unlock()
+	return c.progressReports
+}
+
+type fakeWriteReseter struct {
+	offset int64
```
walac: What's the role of `offset`?

runtime/fetcher/fakes_test.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (w *fakeWriteReseter) Write(p []byte) (int, error) {
+	offset := w.offset + int64(len(p))
+	if int64(len(w.buffer)) < offset {
+		w.buffer = append(w.buffer, make([]byte, offset-int64(len(w.buffer)))...)
+	}
+	copy(w.buffer[w.offset:], p)
+	w.offset = offset
+	return len(p), nil
```
walac: This can be summarized to:
```go
w.offset := w.offset + int64(len(p))
w.buffer = append(w.buffer, p)
return w.offset, nil
```

But I don't see why `offset` is necessary

runtime/fetcher/fetchasstream.go
```diff
@@ -0,0 +1,104 @@
...
+	// Discard current state if any
+	if s.cancel != nil {
+		s.cancel()
+		s.writer.CloseWithError(ErrStreamReset)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	reader, writer := io.Pipe()
+	s.reader = reader
+	s.writer = writer
```
walac: Before re-assigning the pipes, we need to close the old ones

runtime/fetcher/fetchasstream.go
```diff
@@ -0,0 +1,104 @@
...
+// aborted in which case io.Reader will return ErrStreamReset and the Context
+// will be canceled.
+type StreamHandler func(context.Context, io.Reader) error
+
+// FetchAsStream gets a reference as a stream.
+//
+// Notice that target StreamHandler may be invoked multiple times, if the
+// connection breaks while fetching it might be retried. In which case the
+// Context passed to the target StreamHandler will be canceled.
+func FetchAsStream(context Context, reference Reference, target StreamHandler) error {
```
walac: IIUC, we will call the `StreamHandler` only if `Reset` is called, which may never happen if there is no error.

runtime/fetcher/fakes_test.go
jonasfj: This is old code I just renamed it here... That should be clear from commits.

But I'm happy to adopt this :)  it's nicely spotted.

runtime/fetcher/fetchasstream.go
jonasfj: We close the `writer` in the block earlier... The read pipe is passed to the `StreamHandler`, it does not need to be closed.

runtime/fetcher/fetchasstream.go
jonasfj: The `reference.Fetch` may call `streamReseter.Reset`, I'll add a comment to highlight this, because that would be useful :)

runtime/fetcher/fakes_test.go
```diff
@@ -0,0 +1,221 @@
...
+	defer c.m.Unlock()
+	return c.progressReports
+}
+
+type fakeWriteReseter struct {
+	buffer []byte
+}
+
+func (w *fakeWriteReseter) Write(p []byte) (int, error) {
+	w.buffer = append(w.buffer, p...)
```
walac: Do we need to expand `p`?

jonasfj: yes, because we're appending each byte to an array of bytes... I'm fairly sure golang has optimized this code path :)

PR #371, opened 2018-04-22
Use terraformed tc-worker deployment

jonasfj:
The other one should be killed... It's running on my credit card :)
So let's clean up.


jonasfj:
I assigned the scope necessary in:
https://tools.taskcluster.net/auth/roles/repo%3Agithub.com%2Ftaskcluster%2Ftaskcluster-worker%3A*

But let's see if CI passes now :)

PR #378, opened 2018-05-10
Implement relengapi proxy

walac:
This is going to replaced by https://github.com/mozilla/build-relengapi/issues/390 but works as a short-term solution.


plugins/relengapi/relengapi.go
```diff
@@ -0,0 +1,158 @@
...
+	schematypes.MustValidateAndMap(payloadSchema, options.Payload, &P)
+
+	// If disabled, we return nothing
+	if !P.EnableRelengAPIProxy {
+		return plugins.TaskPluginBase{}, nil
+	}
+
+	debug(fmt.Sprintf("Starting relengapi plugin for task %s", options.TaskContext.TaskID))
+	q := options.TaskContext.Queue()
+	task, err := q.Task(options.TaskContext.TaskID)
```
jonasfj: Instead of doing this we could expose `Scopes` from `TaskInfo` instead. But there is a very good reason we don't do this. Scopes should never be read as a list.



There is only one question you can ask: Does `task.scopes` satisfy "X"?
https://github.com/taskcluster/taskcluster-worker/blob/master/runtime/taskcontext.go#L306

A better option is something like:
```go
allPerms := []string{
  "base.tokens.tmp.issue",
  "base.tokens.usr.issue",
  "base.tokens.usr.revoke.my",
  "base.tokens.usr.view.my",
  "clobberer.post.clobber",
  "tooltool.download.internal",
  "tooltool.download.public",
  "tooltool.upload.internal",
  "tooltool.upload.public",
}
perms := []string{}
for _, p := range allPerms {
  if options.TaskContext.HasScopes("worker:relengapi-proxy:" + p) {
    perms = append(perms, p)
  }
}
```

I'm not sure there is an exhaustive list of all permissions. But just adding the ones we think we need is a good start.

runtime/client/queue.go
```diff
@@ -111,6 +112,12 @@ func (m *MockQueue) GetArtifact_SignedURL(taskID, runID, name string, duration t
 	return args.Get(0).(*url.URL), args.Error(1)
 }
 
+// Task is a mock implementation of github.com/taskcluster/taskcluster-client-go/tcqueue.Task
+func (m *MockQueue) Task(taskID string) (*tcqueue.TaskDefinitionResponse, error) {
```
jonasfj: If we can avoid adding this that'll be good, as any plugin that uses it is making a mistake. See comment above...

plugins/relengapi/relengapi.go
jonasfj: Note: extracting information from scopes is always wrong. In this case we ensure that kleene works as expected, because `task.scopes = ['worker:relengapi-proxy:*']` will have all permissions, similarly `task.scopes = ['w*']` will also have all permissions.

walac:
Hrm, `govendor sync` hangs on Mac :/

PR #381, opened 2018-05-15
Fix artifact tests for docker engine

walac:


engines/enginetest/artifacts.go
```diff
@@ -121,10 +121,11 @@ func (c *ArtifactTestCase) TestExtractNestedFolderPath() {
 
 	// Check that NestedFolderFiles was found
 	m.Lock()
+	defer m.Unlock()
 	for _, f := range c.NestedFolderFiles {
 		found := false
 		for _, f2 := range files {
-			if f == f2 {
+			if strings.HasSuffix(f2, f) {
```
jonasfj: I disagreed with this... it must be the complete relative path.

jonasfj: It must be the case that extracting `/folder` returns the file `/folder/hello.txt` with the name reported as `"hello.txt"`.

If you don't think this is correct, then we should change `enginetest`.

We must conform to the interface specification:
https://github.com/taskcluster/taskcluster-worker/blob/058b5b9b962078c6a1c93281ce386040f63868a1/engines/resultset.go#L14-L16

---
If you want to change the interface specification, that's a different story. Perhaps this is a bug in the artifact plugin?

Based on: https://github.com/taskcluster/taskcluster-worker/blob/master/plugins/artifacts/artifacts.go#L426
I would say it expects the `p` to be a relative path, that doesn't include slashes...
But `a.Name` doesn't end with a slash, according to: https://github.com/taskcluster/taskcluster-worker/blob/master/plugins/artifacts/payloadschema.go#L47-L55

engines/docker/resultset.go
```diff
@@ -103,7 +103,6 @@ func (r *resultSet) ExtractFolder(path string, handler engines.FileHandler) erro
 		if !strings.HasPrefix(name, "/") {
 			name = "/" + name // Ensure we always have an absolute path
 		}
-		name = name[len(path+"/"):]
```
jonasfj: Is it possible that this should be `name = name[len(path):]` since we now have `path += "/"`

engines/enginetest/artifacts.go
walac: I am confused then, the path I am getting is absolute

jonasfj:
integration tests added in #384 should fix the bits, and ensure that we have proper tests all the way :)

jonasfj:
@walac, fyi, you were right I had made a mistake with the `len(path + "/")` thing... or so I think.

PR #383, opened 2018-05-21
[WIP] Implement video feature in docker engine

walac:


engines/docker/engine.go
```diff
@@ -100,6 +121,13 @@ func (e *engine) NewSandboxBuilder(options engines.SandboxOptions) (engines.Sand
 	var p payloadType
 	schematypes.MustValidateAndMap(e.PayloadSchema(), options.Payload, &p)
 
+	p.Devices = funk.UniqString(p.Devices)
```
jonasfj: Is this for deduplication? If so, just add `Unique: true` in the JSON schema for `array`...

engines/docker/engine.go
```diff
@@ -72,6 +84,15 @@ func (e *engine) PayloadSchema() schematypes.Object {
 				Description: "Command to run inside the container.",
 				Items:       schematypes.String{},
 			},
+			"devices": schematypes.Array{
+				Title:       "Devices",
+				Description: "List of host devices required.",
+				Items: schematypes.StringEnum{
+					Options: []string{
+						"video",
```
jonasfj: We could just make the option of `video` be dependent on whether devices are enabled or not... anyways, nit... I'm still not sure what's ideal in either case.

engines/docker/engine.go
walac: In a follow up when we have more options.

engines/docker/video.go
```diff
@@ -0,0 +1,51 @@
...
+	devices []device
+}
+
+func newVideoDeviceManager() (*videoDeviceManager, error) {
+	matches, err := filepath.Glob("/dev/video*")
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to call filepath.Glob function")
+	}
+
+	r := regexp.MustCompile("/dev/video[0-9]+")
```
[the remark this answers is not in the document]
walac: Don't know if I followed the idea, but let's keep things simple for now.

engines/docker/video.go
```diff
@@ -0,0 +1,51 @@
...
+		devices[i].path = matches[i]
+	}
+
+	return &videoDeviceManager{
+		devices: devices,
+	}, nil
+}
+
+func (d *videoDeviceManager) claim() *device {
+	for i := range d.devices {
```
jonasfj: locking?

engines/docker/video.go
```diff
@@ -0,0 +1,51 @@
...
+		if !d.devices[i].claimed {
+			return &d.devices[i]
```
jonasfj: returning a reference to an object stored in a list... it's probably better to make a list pointers to devices..

Just so we don't have to think too much about happens when if we mutate the device... or something like that.

jonasfj: At no point do you set, `dev.claimed = true`

engines/docker/sandbox.go
```diff
@@ -277,5 +298,9 @@ func (s *sandbox) dispose() error {
 	if hasErr {
 		return runtime.ErrNonFatalInternalError
 	}
+
+	if s.videoDev != nil {
+		s.videoDev.claimed = false
```
jonasfj: It's probably best to decide what is ecapsulated where... You have a method `videoDeviceManager.release`

engines/docker/sandbox.go
```diff
@@ -87,6 +106,7 @@ func newSandbox(sb *sandboxBuilder) (*sandbox, error) {
 	storage, err := sb.e.Environment.TemporaryStorage.NewFolder()
 	if err != nil {
 		imageHandle.Release()
+		sb.e.video.release(dev)
```
jonasfj: This is probably too soon to release, the container still exists...

engines/docker/video.go
```diff
@@ -0,0 +1,51 @@
...
+type device struct {
+	path    string
+	claimed bool
+}
+
+type videoDeviceManager struct {
+	devices []device
+}
+
+func newVideoDeviceManager() (*videoDeviceManager, error) {
```
jonasfj: I would suggest something like:

```go
package dockerengine

import (
	"errors"
	"sync"

	"github.com/taskcluster/taskcluster-worker/runtime/atomics"
)

type deviceHandle interface {
	Path() string
	Release()
}

type videoDeviceManager struct {
	m       sync.Mutex
	devices []string
}

// newVideoDeviceManager creates a new videoDeviceManager
func newVideoDeviceManager() (*videoDeviceManager, error) {
	// TODO: Find or create devices
	devices := []string{"/dev/video0"}
	return &videoDeviceManager{devices: devices}, nil
}

// GetVideoDevice returns an idle video device
func (vdm *videoDeviceManager) GetVideoDevice() deviceHandle {
	vdm.m.Lock()
	defer vdm.m.Unlock()

	// ensure we have enough devices
	if len(vdm.devices) < 1 {
		panic(errors.New("insufficent number of video devices available"))
	}

	// Create a device handle, ensuring we can't release twice
	vdh := &videoDeviceHandle{owner: vdm}
	vdh.path, vdm.devices = vdm.devices[0], vdm.devices[1:]

	return vdh
}

// Dispose releases all resources assocated with the videoDeviceManager
func (vdm *videoDeviceManager) Dispose() error {
	vdm.m.Lock()
	defer vdm.m.Unlock()

	return nil // TODO: remove devices if any was created
}

type videoDeviceHandle struct {
	owner    *videoDeviceManager
	path     string
	released atomics.Once
}

func (vdh *videoDeviceHandle) Path() string {
	if vdh.released.IsDone() {
		panic(errors.New("videoDeviceHandle.Path() was called after release"))
	}
	return vdh.path
}

func (vdh *videoDeviceHandle) Release() string {
	ok := vdh.released.Do(func() {
		vdh.owner.m.Lock()
		defer vdh.owner.m.Unlock()

		vdh.owner.devices = append(vdh.owner.devices, vdh.path)
	})
	if !ok {
		debug("WARNING: harmless double release() of videoDeviceHandle")
	}
}

```

In the sandbox and resultset code you just release() the `deviceHandle` the same way we release `imageHandle` and pass it around the same way.

While techincally all this is in the same package, I only use methods starting with capital outside this file... the only exception being: `newVideoDeviceManager()`. This makes it easier to track life-cycles, and it is a motivation for moving device management into separate package, though that's a lot less important..

engines/docker/engine.go
jonasfj: I think I would suggest that we do:
```go
devices: [
  {type: 'video'},
],
```

That way, if in some future we decide we want additional options, we can add them to the object...

PR #392, opened 2018-06-07
Fix zip slip vulnerability

walac:
Ref: https://snyk.io/research/zip-slip-vulnerability

### DIFF
--- a/engines/docker/resultset.go
+++ b/engines/docker/resultset.go
@@ -205,7 +205,12 @@ func (r *resultSet) extractResource(resourcePath string, isFolder bool, handler 
 			wg.Add(1)
 			go func(t runtime.TemporaryFile) {
 				defer wg.Done()
-				if handler(path.Join(path.Dir(path.Clean(resourcePath)), hdr.Name), t) != nil {
+				dirpath := path.Dir(path.Clean(resourcePath))
+				fullpath := path.Join(dirpath, hdr.Name)
+				if !strings.HasPrefix(fullpath, dirpath) {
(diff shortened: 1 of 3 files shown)


engines/docker/resultset.go
```diff
@@ -205,7 +205,12 @@ func (r *resultSet) extractResource(resourcePath string, isFolder bool, handler
 			wg.Add(1)
 			go func(t runtime.TemporaryFile) {
 				defer wg.Done()
-				if handler(path.Join(path.Dir(path.Clean(resourcePath)), hdr.Name), t) != nil {
+				dirpath := path.Dir(path.Clean(resourcePath))
+				fullpath := path.Join(dirpath, hdr.Name)
+				if !strings.HasPrefix(fullpath, dirpath) {
+					panic(fmt.Sprintf("%s: illegal path", hdr.Name))
```
jonasfj: Please use `fmt.Errorf`

plugins/cache/extract.go
```diff
@@ -80,6 +81,10 @@ func extractArchive(source io.Reader, target fileSystem) error {
 		} else if info.Mode().IsRegular() {
 			debug("extracting file: '%s'", header.Name)
 
+			if strings.HasPrefix(header.Name, "..") {
+				return fmt.Errorf("%s: illegal file", header.Name)
```
jonasfj: `return runtime.NewMalformedPayloadError(` and this should be `if strings.hasPrefix(path.Clean(header.name)) {`
that way we'll also catch scenarios where we have `myfolder/../../../haha-evil.binary`

PR #394, opened 2018-06-15
Support redeployability

walac:
We create a configuration called rootURL that specificies the root URL
of the taskcluster service.


djmitche:
@walac I think we should close this without merging?

PR #396, opened 2018-06-28
Error out when downloading files if the status is not 200.

tomprince:


tomprince:
We hit this here https://tools.taskcluster.net/groups/K0NdPGOWTLeLDADhSsVaog/tasks/QgVDwlnXShKlHHTDmK4bbA/runs/0/logs/public%2Flogs%2Flive.log which I suspect was actually caused by hg giving an error that was ignored.

It would be good to add some retries, but this should make the actual cause of the error clearer.

tomprince:
@walac

walac (approved):

PR #397, opened 2018-06-29
plugins/relengapi: proxy all services to new url 

garbas:


walac (approved):
Sounds good for me, but could you please run goimports on source file to format source file according to golang conventions?

